PR #7, opened 2025-01-30
(fix): MCP server adjustments of the tools

radektomasek:
## Description

This PR contains several fixes, including:

- addressed the issue with the 'query_table_data' method that wasn't able to query requested data. The issue was caused by the fact the DB_NAME was prefixed with a hardcoded value.
- added a detail description to the DB methods. The change should help the AI tool to understand the context a better way
- fixed linting issues
- updated tests


src/keboola_mcp_server/server.py
davidesner, line 1: @radektomasek this will not work for shared tables (across projects). The database can vary depending on projects. We want to be using RO workspaces. Db name is in most of the projects KEBOOLA_123 where 123 is a project number. In some older projects or on byodb, the prefix can be different. So the only thing that makes sense to parametrise atm is that prefix

davidesner: I wonder if we can force the general query method to use the table detail resource and form the lath properly including the database just by mentioning guilding it in description

radektomasek: @davidesner Thank you for pointing out. I am going to follow the route which tries to extract the info from available resource by following clearly specified function metadata description. 👍

coverage.xml
```diff
@@ -1,12 +1,12 @@
 <?xml version="1.0" ?>
-<coverage version="7.6.8" timestamp="1738187842769" lines-valid="263" lines-covered="147" line-rate="0.5589" branches-covered="0" branches-valid="0" branch-rate="0" complexity="0">
```
radektomasek, line 2: @davidesner @jordanrburger I think we can exclude this file to be version controlled, what do you think?

src/keboola_mcp_server/server.py
radektomasek: @davidesner I've sent you a quick summary on Slack, but to have a visibility here (also for @jordanrburger) -> I had thought about it and the best solution (to make the tooling more robust and future proof), I implemented a concept of db_check using a pattern matching. 

For now, there are two options, one including the `KEBOOLA_` and the other one based on the db name  variable. All DB logic was also separated into dedicated module.

The flow simply assumes the DB name might be determined based on the testing one of two (or more in the future) prefixes to see which ones work with specific tables and uses that one after a test was successful.

I tried to query multiple tables within the project and everything seemed to working fine and I personally like this pattern, especially as this help to make the tooling more robust.

All  the database logic was also moved into a dedicated file/module.

davidesner: @radektomasek I submitted a small update of the query_data tool. It is now more reliant. The query_table_data tool is still very unreliable and fails for me. The raw SQL tool works quite well, I would consider removing the simplified query_table_data completely

davidesner:
I simplified things optimised the instructions, removed the simplified endpoint as it didn't bring any value. Removed some import sort checks and mypy for now as it was only annoying

jordanrburger (approved):

PR #11, opened 2025-03-14
KAB-817: add the concept of "session state factory"

vita-stejskal:
There is quite a bit of refactoring on this branch which can be split in two main parts:

* Extending the FastMCP to allow attaching a custom "state" to the SSE session initiated when an MCP client opens the conversation to an MCP server via the `sse` transport (i.e. the first `GET /sse` request).
* Updating our existing tools to consume the `mcp.server.fastmcp.Context` object and use the custom "state" attached to the `Context`'s `session.state` field. The "state" is created by our own session state factory function which reads the session parameters and creates `KeboolaClient` (Storage API client) and `ConnectionManager` (Snowflake connector) instances. These instances are than available in the tools.

The FastMCP extenstions are in `mcp.py` file and provide the following stuff:

* Definitions of `SessionParams`, `SessionState` and `SessionStateFactory` -- the things that you need in order to implement your own session state factory which is a function that accepts the `SessionParams` and produces `SessionState`.
* `StatefullServerSession` -- the original `mcp.ServerSession` with the extra field for `SessionState` object
* `_KeboolaServer` -- the subclass of `mcp.server.Server` which modifies `run()` function to accept the `SessionState`, creates `StatefullServerSession` and starts the actual SSE server
* `KeboolaMcpServer` -- the subclass of `mcp.server.FastMcp` which modifies `run_stdio_async()` and `run_sse_async()` functions. The `KeboolaMcpServer` accepts `SessionStateFoactory` in its constructor and the `run_*_async()` functions call the factory in order to get the custom "state" that is then passed to `_KeboolaServer` when starting it.
* The `SessionParams` needed for the `SessionStateFactory` are taken either from the process environment (when using `stdio` transport) or from the HTTP query parameters (when using `sse` transport).

Notes:
* I am surprised that the extensions from `mcp.py` are not already part of the original FastMCP implementation in `mcp` python package. I believe that they would be useful for anyone using FastMCP, especially when using the `sse` transport.
* The FastMCP implementation of the "resources" concept from the MCP (protocol) does not seem to support passing the `mcp.server.fastmcp.Context` object to the functions registered as the resources providers. It seems odd and I did not find anything in the MCP (protocol) itself that would justify this. But I could be wrong.
* The testing of the MCP servers implemented using FastMCP seems to be very challenging. There is no support for testing in the `mcp` python package itself. I'd expect something like Starlette's `TestClient` capable of invoking directly the endpoints of a Starlette-base server.
* In general, my feeling is that the main ideas behind FastMCP are great, but its implementation is still immature and hard to use when you need things beyond the examples from its tutorial.

### DIFF
--- a/src/keboola_mcp_server/mcp.py
+++ b/src/keboola_mcp_server/mcp.py
@@ -49,11 +49,11 @@ def _default_session_state_factory(params: SessionParams) -> SessionState:
 
 class StatefullServerSession(ServerSession):
     def __init__(
-            self,
-            read_stream: MemoryObjectReceiveStream[types.JSONRPCMessage | Exception],
-            write_stream: MemoryObjectSendStream[types.JSONRPCMessage],
-            init_options: InitializationOptions,
-            state: SessionState | None = None
+        self,
+        read_stream: MemoryObjectReceiveStream[types.JSONRPCMessage | Exception],
+        write_stream: MemoryObjectSendStream[types.JSONRPCMessage],
+        init_options: InitializationOptions,
+        state: SessionState | None = None,
     ) -> None:
         super().__init__(read_stream, write_stream, init_options)
         self._state = state or {}
@@ -65,11 +65,11 @@ class StatefullServerSession(ServerSession):
 
 class _KeboolaServer(Server):
     def __init__(
-            self,
-            name: str,
-            version: str | None = None,
-            instructions: str | None = None,
-            lifespan: Callable[["Server"], AbstractAsyncContextManager[LifespanResultT]] | None = None,
+        self,
+        name: str,
+        version: str | None = None,
+        instructions: str | None = None,
+        lifespan: Callable[["Server"], AbstractAsyncContextManager[LifespanResultT]] | None = None,
     ) -> None:
         super().__init__(name, version=version, instructions=instructions, lifespan=lifespan)
 
@@ -83,7 +83,7 @@ class _KeboolaServer(Server):
         # but also make tracing exceptions much easier during testing and when using
         # in-process servers.
         raise_exceptions: bool = False,
-        state: SessionState | None = None
+        state: SessionState | None = None,
     ):
         async with AsyncExitStack() as stack:
             lifespan_context = await stack.enter_async_context(self.lifespan(self))
@@ -93,7 +93,7 @@ class _KeboolaServer(Server):
 
             async with anyio.create_task_group() as tg:
                 async for message in session.incoming_messages:
-                    logger.debug(f'Received message: {message}')
+                    logger.debug(f"Received message: {message}")
 
                     tg.start_soon(
                         self._handle_message,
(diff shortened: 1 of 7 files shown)


jordanrburger:
From what I remember, the changes in this PR won't actually be compatible with Claude Desktop, right? But it's still important/useful for using the MCP server with Cursor and other services?

vita-stejskal:
> From what I remember, the changes in this PR won't actually be compatible with Claude Desktop, right? But it's still important/useful for using the MCP server with Cursor and other services?

The changes on this branch are compatible with Claude Desktop. They have no effect on the "command" transport type which works the same as before -- i.e. you pass the credentials needed by your MCP server via the environment variables.

The Keboola MCP server supports multiple parameters (see [Config](https://github.com/keboola/keboola-mcp-server/blob/KAB-817-statefull-sessions/src/keboola_mcp_server/config.py#L12) class) and they can either be specified via environment variables (use uppercase and prefixed with `KBC_`, e.g. `KBC_STORAGE_TOKEN`) or via the URL parameters (use lowercase, no prefix, e.g. `storage_token`). So, the implementation now supports both types of transport (`command` and `sse`) and they are fully equivalent.

src/keboola_mcp_server/mcp.py
```diff
@@ -0,0 +1,169 @@
...
+            lifespan_context = await stack.enter_async_context(self.lifespan(self))
+            session = await stack.enter_async_context(
+                StatefullServerSession(read_stream, write_stream, initialization_options, state)
+            )
+
+            async with anyio.create_task_group() as tg:
+                async for message in session.incoming_messages:
+                    logger.debug(f"Received message: {message}")
+
+                    tg.start_soon(
```
radcliffkey, line 98: it the server was really stateful, we couldn't really process all messages in parallel, right?

radcliffkey (approved):

src/keboola_mcp_server/mcp.py
vita-stejskal: I think you are right. I took this implementation from `FastMCP` and change the bit that creates the `StatefullServerSession`. Our tools are actually stateless and so it should not be a problem if the client sends multiple requests and they are processed asynchronously.

vita-stejskal:
> From what I remember, the changes in this PR won't actually be compatible with Claude Desktop, right? But it's still important/useful for using the MCP server with Cursor and other services?

AFAIK @mariankrotil successfully tested these changes in Claude Desktop using the `command` transport. So, we should be safe to merge this PR.

jordanrburger:
> > From what I remember, the changes in this PR won't actually be compatible with Claude Desktop, right? But it's still important/useful for using the MCP server with Cursor and other services?
> 
> AFAIK @mariankrotil successfully tested these changes in Claude Desktop using the `command` transport. So, we should be safe to merge this PR.

Awesome! I'll go ahead and merge

jordanrburger (approved):

PR #12, opened 2025-03-18
KAB-871-mcp-ro-component-tools2

mariankrotil:
Partial PR refactoring current component tools code, no new feature added yet 
- modularize component tools code into separate files
- utilizing pydantic.BaseModel for structured output of the tool functions

### DIFF
--- a/src/keboola_mcp_server/component_tools.py
+++ b/src/keboola_mcp_server/component_tools.py
@@ -1,0 +1,51 @@
+from typing import Any, Dict, List, Optional, cast
+
+from mcp.server.fastmcp import Context, FastMCP
+from pydantic import BaseModel, Field
+
+from keboola_mcp_server.client import KeboolaClient
+
+
+class Component(BaseModel):
+    id: str = Field(description="The ID of the component")
+    name: str = Field(description="The name of the component")
+
+
+class ComponentConfig(BaseModel):
+    id: str = Field(description="The ID of the component configuration")
+    name: str = Field(description="The name of the component configuration")
+    description: Optional[str] = Field(description="The description of the component configuration")
+    created: str = Field(description="The creation date of the component configuration")
+
+
+def add_component_tools(mcp: FastMCP) -> None:
+    """Add tools to the MCP server."""
+    mcp.add_tool(list_components, "list_components", "List all available components.")
+    mcp.add_tool(list_component_configs, "list_component_configs", "List all configurations for a given component.")
+
+
+async def list_components(ctx: Context) -> List[Component]:
+    """List all available components."""
+    client = ctx.session.state["sapi_client"]
+    assert isinstance(client, KeboolaClient)
+    r_components = client.storage_client.components.list()
+    return [Component(id=r_comp["id"], name=r_comp["name"]) for r_comp in r_components]
+
+
+async def list_component_configs(
+    component_id: str,
+    ctx: Context,
+) -> List[ComponentConfig]:
+    """List all configurations for a given component."""
+    client = ctx.session.state["sapi_client"]
+    assert isinstance(client, KeboolaClient)
+    r_configs = client.storage_client.configurations.list(component_id)
+    return [
+        ComponentConfig(
+            id=r_config["id"],
+            name=r_config["name"],
+            description=r_config["description"],
+            created=r_config["created"],
+        )
+        for r_config in r_configs
+    ]
--- a/src/keboola_mcp_server/component_tools.py
+++ b/src/keboola_mcp_server/component_tools.py
@@ -21,7 +21,11 @@ class ComponentConfig(BaseModel):
 def add_component_tools(mcp: FastMCP) -> None:
     """Add tools to the MCP server."""
     mcp.add_tool(list_components, "list_components", "List all available components.")
-    mcp.add_tool(list_component_configs, "list_component_configs", "List all configurations for a given component.")
+    mcp.add_tool(
+        list_component_configs,
+        "list_component_configs",
+        "List all configurations for a given component.",
+    )
 
 
 async def list_components(ctx: Context) -> List[Component]:
--- a/src/keboola_mcp_server/cli.py
+++ b/src/keboola_mcp_server/cli.py
@@ -49,10 +49,9 @@ def main(args: Optional[List[str]] = None) -> None:
     parsed_args = parse_args(args)
 
     # Create config from the CLI arguments
-    config = Config.from_dict({
-        'storage_api_url': parsed_args.api_url,
-        'log_level': parsed_args.log_level
-    })
+    config = Config.from_dict(
+        {"storage_api_url": parsed_args.api_url, "log_level": parsed_args.log_level}
+    )
 
     try:
         # Create and run server
--- a/src/keboola_mcp_server/cli.py
+++ b/src/keboola_mcp_server/cli.py
@@ -1,7 +1,6 @@
 """Command-line interface for the Keboola MCP server."""
 
 import argparse
-import asyncio
 import logging
 import sys
 from typing import List, Optional
@@ -24,7 +23,7 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
     parser = argparse.ArgumentParser(description="Keboola MCP Server")
     parser.add_argument(
         "--transport",
-        choices=["stdio"],
+        choices=["stdio", "sse"],
         default="stdio",
         help="Transport to use for MCP communication",
     )
@@ -35,7 +34,7 @@ def parse_args(args: Optional[List[str]] = None) -> argparse.Namespace:
         help="Logging level",
     )
     parser.add_argument(
-        "--api-url", help="Keboola Storage API URL (defaults to https://connection.keboola.com)"
+        "--api-url", default="https://connection.keboola.com", help="Keboola Storage API URL"
     )
 
     return parser.parse_args(args)
@@ -49,11 +48,11 @@ def main(args: Optional[List[str]] = None) -> None:
     """
     parsed_args = parse_args(args)
 
-    # Create config from environment, but override with CLI args
-    config = Config.from_env()
-    if parsed_args.api_url:
-        config.storage_api_url = parsed_args.api_url
-    config.log_level = parsed_args.log_level
+    # Create config from the CLI arguments
+    config = Config.from_dict({
+        'storage_api_url': parsed_args.api_url,
+        'log_level': parsed_args.log_level
+    })
 
     try:
         # Create and run server
--- a/src/keboola_mcp_server/config.py
+++ b/src/keboola_mcp_server/config.py
@@ -1,4 +1,5 @@
 """Configuration handling for the Keboola MCP server."""
+
 import dataclasses
 import logging
 from dataclasses import dataclass
@@ -29,12 +30,12 @@ class Config:
         for f in dataclasses.fields(cls):
             if f.name in d:
                 options[f.name] = d.get(f.name)
-            elif (dict_name := f'KBC_{f.name.upper()}') in d:
+            elif (dict_name := f"KBC_{f.name.upper()}") in d:
                 options[f.name] = d.get(dict_name)
         return options
 
     @classmethod
-    def from_dict(cls, d: Mapping[str, str]) -> 'Config':
+    def from_dict(cls, d: Mapping[str, str]) -> "Config":
         """
         Creates new `Config` instance with values read from the input mapping.
         The keys in the input mapping can either be the names of the fields in `Config` class
@@ -42,7 +43,7 @@ class Config:
         """
         return cls(**cls._read_options(d))
 
-    def replace_by(self, d: Mapping[str, str]) -> 'Config':
+    def replace_by(self, d: Mapping[str, str]) -> "Config":
         """
         Creates new `Config` instance from the existing one by replacing the values from the input mapping.
         The keys in the input mapping can either be the names of the fields in `Config` class
--- a/src/keboola_mcp_server/config.py
+++ b/src/keboola_mcp_server/config.py
@@ -71,3 +71,16 @@ class Config:
                 self.snowflake_database,
             ]
         )
+
+    def __repr__(self):
+        params: list[str] = []
+        for f in dataclasses.fields(self):
+            value = getattr(self, f.name)
+            if value:
+                if 'token' in f.name or 'password' in f.name:
+                    params.append(f'{f.name}=\'****\'')
+                else:
+                    params.append(f'{f.name}=\'{value}\'')
+            else:
+                params.append(f'{f.name}=None')
+        return f'Config({", ".join(params)})'
--- a/src/keboola_mcp_server/config.py
+++ b/src/keboola_mcp_server/config.py
@@ -1,18 +1,17 @@
 """Configuration handling for the Keboola MCP server."""
-
+import dataclasses
 import logging
-import os
 from dataclasses import dataclass
-from typing import Optional
+from typing import Mapping, Optional
 
 logger = logging.getLogger(__name__)
 
 
-@dataclass
+@dataclass(frozen=True)
 class Config:
     """Server configuration."""
 
-    storage_token: str
+    storage_token: Optional[str] = None
     storage_api_url: str = "https://connection.keboola.com"
     log_level: str = "INFO"
     # Add Snowflake credentials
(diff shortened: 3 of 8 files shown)


src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,55 @@
...
+class ComponentConfig(BaseModel):
+    id: str = Field(description="The ID of the component configuration")
+    name: str = Field(description="The name of the component configuration")
+    description: Optional[str] = Field(description="The description of the component configuration")
+    created: str = Field(description="The creation date of the component configuration")
+
+
+def add_component_tools(mcp: FastMCP) -> None:
+    """Add tools to the MCP server."""
+    mcp.add_tool(list_components, "list_components", "List all available components.")
```
vita-stejskal, line 23: You shouldn't have to specify the `name` and `description` of the tool explicitly. FastMCP infers this information from the function's signature and dosctring.

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,55 @@
...
+async def list_components(ctx: Context) -> List[Component]:
+    """List all available components."""
+    client = ctx.session.state["sapi_client"]
+    assert isinstance(client, KeboolaClient)
+    r_components = client.storage_client.components.list()
+    return [Component(id=r_comp["id"], name=r_comp["name"]) for r_comp in r_components]
+
+
+async def list_component_configs(
+    component_id: str,
```
vita-stejskal, line 40: Use `component_id: Annotated[str, Field(description='The ID of a component which configurations should be listed.')]` to describe the meaning of the parameter. See examples in the [FastMCP docs](https://github.com/jlowin/fastmcp?tab=readme-ov-file#tools).

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,55 @@
...
+        "List all configurations for a given component.",
+    )
+
+
+async def list_components(ctx: Context) -> List[Component]:
+    """List all available components."""
+    client = ctx.session.state["sapi_client"]
+    assert isinstance(client, KeboolaClient)
+    r_components = client.storage_client.components.list()
+    return [Component(id=r_comp["id"], name=r_comp["name"]) for r_comp in r_components]
```
vita-stejskal, line 36: Ideally, you should use `Component.model_validate(r_comp)` instead of translating the Storage API response data to the `Component` instance manually. The same applies for the `ComponentConfig` class below.

vita-stejskal (changes requested):
Also please:
* Add tests for the tools -- mock the SAPI client in the `Context` and make sure the tools return expected response objects.
* Try to retrieve the list of the tools exposed by the MCP server and confirm that their descriptions really contain the specifications (presumably JSON schema) of the parameter and output values. You can use the MCP client from [mcp](https://pypi.org/project/mcp/) package to [list the available tools](https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#writing-mcp-clients).
* Increment the project version in `pyproject.toml` file.

mariankrotil:
cc @vita-stejskal 
I added a ComponentClient to handle component requests, avoiding the need to use KeboolClient's get method and expose endpoints within the tool function, which I find messy.
is it ok?

src/keboola_mcp_server/component_tools.py
mariankrotil: Done here b153083614075b48fc52dc238b33dd9cf1d23588

src/keboola_mcp_server/component_tools.py
mariankrotil: Done here b153083614075b48fc52dc238b33dd9cf1d23588

src/keboola_mcp_server/component_tools.py
mariankrotil: Done here b153083614075b48fc52dc238b33dd9cf1d23588

mariankrotil:
> cc @vita-stejskal I added a ComponentClient to handle component requests, avoiding the need to use KeboolClient's get method and expose endpoints within the tool function, which I find messy. is it ok?

Oops. I ran into an issue where we would be modifying the same file. So, I went with the first option and exposed the endpoints, but we can refactor it in the future if needed.

mariankrotil:
> Also please:
> 
> * Add tests for the tools -- mock the SAPI client in the `Context` and make sure the tools return expected response objects.
> * Try to retrieve the list of the tools exposed by the MCP server and confirm that their descriptions really contain the specifications (presumably JSON schema) of the parameter and output values. You can use the MCP client from [mcp](https://pypi.org/project/mcp/) package to [list the available tools](https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#writing-mcp-clients).
> * Increment the project version in `pyproject.toml` file.

* Tests are here: 4806aa33c59f32be2bf1cc010e913a81c441e365 febc2d90d6e5418219296086dce60756cb5fdd43
*  List of available tools in mcp server:
image

tests/test_component_tools.py
```diff
@@ -0,0 +1,86 @@
...
+    assert all(c.name == item["name"] for c, item in zip(result, mock_components))
+
+    mock_client.storage_client.components.list.assert_called_once()
+
+
+@pytest.mark.asyncio
+async def test_list_component_configs(mock_context):
+    """Test list_component_configs tool."""
+    context, mock_client = mock_context
+    mock_client.storage_client.configurations = MagicMock()
```
vita-stejskal, line 44: I don't think that you have to set all those mocks explicitly. `MagicMock` will replicate itself into all attributes that you access on the mock.

vita-stejskal (changes requested):

coverage.xml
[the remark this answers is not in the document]
mariankrotil: I removed it from Git tracking in this repo because it was already listed in the .gitignore file, but that was not working. Coverage files update automatically whenever you create or modify tests and run pytest command as they track how much of our code is covered by the tests. But I think they do not have to be shared in the git repo. Or am I wrong?

tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+    mock_client = MagicMock(spec=KeboolaClient)
+    mock_client.storage_client = MagicMock(spec=Client)
+    mock_client.storage_client.components = MagicMock()
+
+    context.session.state["sapi_client"] = mock_client
+    return context, mock_client
+
+
+@pytest.fixture
+def mock_context_with_db() -> Tuple[Context, KeboolaClient, ConnectionManager, DatabasePathManager]:
```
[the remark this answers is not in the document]
mariankrotil: Removed here: 81c8bb7bf589f6dacaf83ba99ff641350b6bc2c4

tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+
+    context.session.state["sapi_client"] = mock_client
+    context.session.state["connection_manager"] = mock_connection_manager
+    context.session.state["db_path_manager"] = mock_db_path_manager
+
+    return context, mock_client, mock_connection_manager, mock_db_path_manager
+
+
+@pytest.fixture
+def mock_snowflake_connection():
```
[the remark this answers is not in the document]
mariankrotil: removed here: 81c8bb7bf589f6dacaf83ba99ff641350b6bc2c4

tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+        - Mocked Context
+        - Mocked KeboolaClient
+    """
+    context = MagicMock(spec=Context)
+    context.session.state = {}
+
+    # Mock KeboolaClient
+    mock_client = MagicMock(spec=KeboolaClient)
+    mock_client.storage_client = MagicMock(spec=Client)
+    mock_client.storage_client.components = MagicMock()
```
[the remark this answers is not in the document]
mariankrotil: Done here: 81c8bb7bf589f6dacaf83ba99ff641350b6bc2c4

tests/test_component_tools.py
mariankrotil: Removing it leads to an error `FAILED tests/test_component_tools.py::test_list_component_configs- AttributeError: Mock object has no attribute 'components'`

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,68 @@
...
+
+    r_configs = client.storage_client.configurations.list(component_id)
+    logger.info(f"Found {len(r_configs)} configurations for component {component_id}.")
+    return [ComponentConfig.model_validate(r_config) for r_config in r_configs]
+
+
+async def get_component_details(
+    component_id: Annotated[str, "The ID of the component for which details should be retrieved."],
+    ctx: Context,
+) -> Component:
```
vita-stejskal, line 61: I think that this tool is not of much use. It returns exactly the same information as the `list_components()` tool.

It would be much more useful from the user's perspective to have a tool that gives details of a configured component -- i.e. the details of a component configuration in Keboola terminology.

It is the stuff that you get when calling `client.storage_client.configurations.detail(<component-id>, <configuration-id>)`. This should contain the actual parameters and their values that the user used for setting up that component in his project.

vita-stejskal (changes requested):

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,68 @@
...
+    """List all available components."""
+    client = ctx.session.state["sapi_client"]
+    assert isinstance(client, KeboolaClient)
+
+    r_components = client.storage_client.components.list()
+    logger.info(f"Found {len(r_components)} components.")
+    return [Component.model_validate(r_comp) for r_comp in r_components]
+
+
+async def list_component_configs(
```
vita-stejskal, line 43: Please rename this to `list_component_configurations`. The name is visible to the LLM and it's better not to use abbreviations.

src/keboola_mcp_server/component_tools.py
vita-stejskal: So, ideally, this tool should be called `get_component_configuration(component_id: Annotated[str, ...], configuration_id: Annotated[str, ...], ctx: Context) -> ComponentConfig`.

The return type could either be `ComponentConfig` with additional fields that would carry on the actual component configuration (the fields would be optional and `None` by default) or it could be a different pydantic object called `ComponentConfigurationDetail` (or something like that) which would carry on similar information as `ComponentConfig` plus the extra fields.

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,68 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Component(BaseModel):
+    id: str = Field(description="The ID of the component")
+    name: str = Field(description="The name of the component")
+
+
+class ComponentConfig(BaseModel):
```
vita-stejskal, line 17: I'd consider renaming this to `ComponentConfiguration`.

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,68 @@
...
+    id: str = Field(description="The ID of the component configuration")
```
vita-stejskal, line 18: I think that this class needs to carry two IDs -- `component_id` and `configuration_id`.

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,68 @@
...
+from mcp.server.fastmcp import Context, FastMCP
+from pydantic import BaseModel, Field
+
+from keboola_mcp_server.client import KeboolaClient
+
+logger = logging.getLogger(__name__)
+
+
+class Component(BaseModel):
+    id: str = Field(description="The ID of the component")
```
vita-stejskal, line 13: This should be renamed to `component_id`.

vita-stejskal (changes requested):

src/keboola_mcp_server/component_tools.py
mariankrotil: I thought that we have agreed that this improvement for user experience will be part of a different task. I expected to address this terminology distinction between component and configuration by unifying those into the `fully qualified id` in this task: [KAB-916](https://keboola.atlassian.net/browse/KAB-916).  But I can solve it within this task, what do you think?

src/keboola_mcp_server/component_tools.py
mariankrotil: Done here: 77347acbd177f6100038056c129a16539da41237

src/keboola_mcp_server/component_tools.py
mariankrotil: HEre 77347acbd177f6100038056c129a16539da41237

src/keboola_mcp_server/component_tools.py
mariankrotil: renamed here: 
77347acbd177f6100038056c129a16539da41237

src/keboola_mcp_server/component_tools.py
mariankrotil: done here 77347acbd177f6100038056c129a16539da41237

src/keboola_mcp_server/component_tools.py
mariankrotil: Following changes are made here to satisfy this configuration details : 77347acbd177f6100038056c129a16539da41237 but I would focus on it more in the new task.

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,106 @@
...
+        validation_alias=AliasChoices("name", "configuration-name", "configuration_name"),
+        serialization_alias="name",
+    )
+    configuration_description: Optional[str] = Field(
+        description="The description of the component configuration",
+        validation_alias=AliasChoices(
+            "description", "configuration-description", "configuration_description"
+        ),
+        serialization_alias="description",
+    )
```
vita-stejskal, line 47: Without fields that would contain the actual configuration parameters this is still not that useful, although it looks much better now.

vita-stejskal (approved):

ncapek (approved):

src/keboola_mcp_server/component_tools.py
```diff
@@ -12,13 +12,13 @@
 class Component(BaseModel):
     component_id: str = Field(
         description="The ID of the component",
-        validation_alias=AliasChoices("id", "componentId", "component-id", "component_id"),
-        serialization_alias="id",
+        validation_alias=AliasChoices("id", "component_id"),
```
vita-stejskal, line 15: Well, so far we've followed the same naming rules that we use for the REST API and they mandate supporting all of the `snake_case`, `camelCase` and `kebab-case`. We don't have to follow these rules in MCP, because it's not really the REST API. But if we change our approach we should be consistent and make the change everywhere.

So, I'd keep the `"componentId", "component-id"` validation aliases for now.

src/keboola_mcp_server/component_tools.py
vita-stejskal: > I thought that we have agreed that this improvement for user experience will be part of a different task. I expected to address this terminology distinction between component and configuration by unifying those into the fully qualified id in this task: [KAB-916](https://keboola.atlassian.net/browse/KAB-916). But I can solve it within this task, what do you think?

Yes, we agreed that we will refactor the component/configuration tools and somehow address that component/configuration dichotomy. But here I talk about something different. I am not suggesting that we should do that work in the task/PR. It's better to have a call about this in person, ok?

vita-stejskal (approved):

src/keboola_mcp_server/component_tools.py
mariankrotil: ok, added here: 480654ee0f5749184985d548bb1d6173569d33d0

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,135 @@
...
+        get_component_configuration_details,
+    ]
+    for tool in component_tools:
+        logger.info(f"Adding tool {tool.__name__} to the MCP server.")
+        mcp.add_tool(tool)
+
+    logger.info("Component tools initialized.")
+
+
+async def list_components(ctx: Context) -> List[Component]:
```
ncapek, line 82: We might want to consider a different name/verb for all "listing" tools. Sometimes I find it difficult to write good docstings and descriptions because the tool also returns a python List type. I realize the client calls it list, that's why we chose this name. A possible option is "retrieve".

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,135 @@
...
+    component_id: Annotated[
+        str, "The ID of the component for which configurations should be listed."
+    ],
+    ctx: Context,
+) -> List[ComponentConfiguration]:
+    """List all configurations for a given component."""
+    client = KeboolaClient.from_state(ctx.session.state)
+
+    component = await get_component_details(component_id, ctx)
+    r_configs = client.storage_client.configurations.list(component_id)
```
ncapek, line 101: I don't like this variable name. What does `r_configs` stand for?

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+async def get_component_configuration_details(
+    component_id: Annotated[str, "The ID of the component for which details should be retrieved."],
+    configuration_id: Annotated[
+        str, "The ID of the configuration for which details should be retrieved."
+    ],
+    ctx: Context,
+) -> ComponentConfiguration:
+    """Detail a given component configuration."""
```
ncapek, line 116: "Detail a given component configuration"? The English seems a bit off here. Perhaps it should be "Get the detail of a given component configuration"?

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,135 @@
...
+    client = KeboolaClient.from_state(ctx.session.state)
+
+    component = await get_component_details(component_id, ctx)
+    r_config = client.storage_client.configurations.detail(component_id, configuration_id)
```
ncapek, line 120: Same issue as the other comment. A better variable name can hopefully be found.

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,135 @@
...
+async def get_component_details(
+    component_id: Annotated[
+        str, Field(str, description="The ID of the Keboola component you want details about")
+    ],
+    ctx: Context,
+) -> Component:
+    """Retrieve detailed information about a original Keboola component object given component ID."""
+    client = KeboolaClient.from_state(ctx.session.state)
+
+    endpoint = "branch/{}/components/{}".format(client.storage_client._branch_id, component_id)
```
ncapek, line 133: Why are you using `.format()`? I think the recommended way for current python versions is to use f-strings.

src/keboola_mcp_server/component_tools.py
```diff
@@ -0,0 +1,135 @@
...
+    r_component = await client.get(endpoint)
```
ncapek, line 134: Same naming issue as previously.

tests/test_component_tools.py
```diff
@@ -0,0 +1,111 @@
...
+        {"id": "keboola.ex-aws-s3", "name": "AWS S3 Extractor"},
+        {"id": "keboola.ex-google-drive", "name": "Google Drive Extractor"},
+    ]
+    keboola_client.storage_client.components.list = MagicMock(return_value=mock_components)
+
+    result = await list_components(mcp_context_client)
+
+    assert len(result) == 2
+    assert all(isinstance(component, Component) for component in result)
+    assert all(c.component_id == item["id"] for c, item in zip(result, mock_components))
```
ncapek, line 32: I'm still not a fan of these single letter variables, unless maybe `i` in a textbook example of a `for` cycle. Also, since you're iterating over something called `result` why should individual items be prefixed `c`? Presumably because the result actually contains components. But I have to return a few lines up to figure this out, which is unfortunate.

src/keboola_mcp_server/component_tools.py
mariankrotil: fixed here 8d6ee0c6a4c11df20cf217d9c2931a304743f71d

src/keboola_mcp_server/component_tools.py
mariankrotil: changed here as u suggest `get tje detail of ...` : 4bb249aa07d6cf7131575334b4981eb890a97ce9

src/keboola_mcp_server/component_tools.py
mariankrotil: renamed here: 4bb249aa07d6cf7131575334b4981eb890a97ce9

src/keboola_mcp_server/component_tools.py
mariankrotil: `raw_configurations` meaning that it is a configuration in json format returned from the API.
renamed here: 4bb249aa07d6cf7131575334b4981eb890a97ce9

src/keboola_mcp_server/component_tools.py
mariankrotil: renamed here: 4bb249aa07d6cf7131575334b4981eb890a97ce9

vita-stejskal (approved):

PR #14, opened 2025-03-20
KAB-889: fix the fully qualified names of the Keboola tables

vita-stejskal:
This change makes the use of the real database names fetched from the Snowflake connection to create the fully qualified name of a Keboola table rather than just guessing it. This gives the LLM a chance to generate SQL queries with the correct table names.

### DIFF
--- a/src/keboola_mcp_server/database.py
+++ b/src/keboola_mcp_server/database.py
@@ -64,6 +64,7 @@ class DatabasePathManager:
                 cursor = conn.cursor(DictCursor)
 
                 if source_table := table.get("sourceTable"):
+                    # a table linked from some other project
                     schema_name, table_name = source_table["id"].rsplit(sep=".", maxsplit=1)
                     source_project_id = source_table["project"]["id"]
                     result = cursor.execute(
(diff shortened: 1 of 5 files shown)


src/keboola_mcp_server/database.py
```diff
@@ -80,8 +81,16 @@ def get_table_fqn(self, table: dict[str, Any]) -> Optional[TableFqn]:
                     result = cursor.execute(
                         f'select CURRENT_DATABASE() as "current_database", CURRENT_SCHEMA() as "current_schema";'
                     ).fetchone()
-                    db_name, schema_name = result["current_database"], result["current_schema"]
-                    table_name = table["name"]
+                    db_name = result["current_database"]
```
davidesner, line 84: @vita-stejskal  I am not sure about this, we would need to make sure that the connection has a current database setup, it is not required for snowflake connection. Just in case I want to point out this may be in clash with https://github.com/keboola/keboola-mcp-server/blob/master/src/keboola_mcp_server/database.py#L186 where the method `find_working_connection` is only looking for `KEBOOLA_` db names. We should review the connection initialisation logic, it's not ideal atm.

vita-stejskal: > I am not sure about this, we would need to make sure that the connection has a current database setup, it is not required for snowflake connection.

Are you saying that `CURRENT_DATABASE()` function could return `null`? The [docs](https://docs.snowflake.com/en/sql-reference/functions/current_database) don't mention situations when the function returns `null` or does not work for some reason.

Also it seems that connections to a workspace created in Keboola UI do set the current/default database. Are there pitfalls? cc @odinuv

vita-stejskal: > Just in case I want to point out this may be in clash with https://github.com/keboola/keboola-mcp-server/blob/master/src/keboola_mcp_server/database.py#L186 where the method find_working_connection is only looking for KEBOOLA_ db names. We should review the connection initialisation logic, it's not ideal atm.

Yes, I am aware of this and tried to improve the situation by adding another `ConnectionPattern` which prefixes the database name with `SAPI_`. But even that does not cover all the possibilities.

Anyway, I'll get rid of all this code for "finding snowflake connection" when I start using the "workspace" endpoints in Storage API. That way the MCP tools will only need the workspace ID (the correct one of the many), but will not have to concern itself with snowflake connections.

PR #16, opened 2025-03-24
Kab 863 refactor tables and buckets tools

ncapek:
This PR will contain the refactor for all table and bucket related MCP tools. Specifically to convert the outputs of these tools into pydantic objects.

### DIFF
--- a/src/keboola_mcp_server/server.py
+++ b/src/keboola_mcp_server/server.py
@@ -3,11 +3,11 @@
 import csv
 import logging
 from io import StringIO
-from typing import Any, cast, Dict, List, Optional, TypedDict
-from pydantic import BaseModel, Field
+from typing import Any, Dict, List, Optional, TypedDict, cast
 
 import snowflake.connector
 from mcp.server.fastmcp import Context, FastMCP
+from pydantic import BaseModel, Field
 
 from keboola_mcp_server.mcp import (
     KeboolaMcpServer,
@@ -15,6 +15,7 @@ from keboola_mcp_server.mcp import (
     SessionState,
     SessionStateFactory,
 )
+
 from .client import KeboolaClient
 from .config import Config
 from .database import ConnectionManager, DatabasePathManager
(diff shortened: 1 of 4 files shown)


src/keboola_mcp_server/server.py
```diff
@@ -3,49 +3,62 @@
...
-    dataSizeBytes: int
+class BucketInfo(BaseModel):
+    id: str = Field(..., description="Unique identifier for the bucket")
+    name: str = Field(..., description="Name of the bucket")
+    description: Optional[str] = Field(None, description="Description of the bucket")
+    stage: Optional[str] = Field(
+        None, description="Stage of the bucket ('in' for input stage, 'out' for output stage)"
+    )
+    created: str = Field(..., description="Creation timestamp of the bucket")
+    tables_count: Optional[int] = Field(None, description="Number of tables in the bucket")
```
vita-stejskal, line 34: I don't have a strong preference here, but I'd say that we should stick to the Keboola's general convention of using `cameCase` field names when serializing to / deserializing from JSON. So, all the fields that contain the underscore char should define:
```
validation_alias=AliasChoices('tableCount', 'table_count', 'table-count'),
serialization_alias='tableCount'
```
Although this is not a REST API, it's better to be consistent.

tests/test_server.py
```diff
@@ -124,3 +147,69 @@ async def query_table_data(
             server.__dict__["query_table"] = original_query_table
         if original_query_table_data:
             server.__dict__["query_table_data"] = original_query_table_data
+
+
+@pytest.mark.asyncio
+async def test_list_all_buckets(test_config: Config, mock_buckets: List[Dict[str, Any]]) -> None:
```
vita-stejskal, line 179: Unless I am badly mistaken these tests in their current form just test nothing. You replace the server's tool with a mock and then directly call that mock, right? IMO a better approach would be to mock the `KeboolaClient` in the `Context` sent to the server's tool and then check what that tool returns. The mocking could be tricky due to the lengthy call chains used in `KeboolaClient`.

vita-stejskal (changes requested):
This looks good in general, thanks! I'd fix those few things mentioned in my review and you should be good to merge.

vita-stejskal:
@ncapek When you merge this PR, please open another one and move the buckets/tables related tools (functions) to a separate file (e.g. `storage_tools.py`). In the `create_server()` function in `server.py` use `mcp.add_tool()` to add the individual tools.

vita-stejskal (dismissed):

PR #19, opened 2025-03-27
Kab 864 sql sapi

vita-stejskal:
This simplifies the MCP server set up by removing most of the snowflake related credentials. The only parameters that are needed now are:

* `--api-url` -- URL for accessing Keboola Storage API
* `KBC_STORAGE_TOKEN` -- Storage API token
* `KBC_WORKSPACE_USER` -- username for accessing the Snowflake workspace

The SQL tools for executing SQL SELECT queries calls the Storage API rather than accessing the Snowflake directly.


jordanrburger (dismissed):

src/keboola_mcp_server/sql_tools.py
```diff
@@ -24,26 +156,22 @@ async def query_table(
...
-            writer.writerow(columns)
-            writer.writerows(result)
+    workspace_manager = WorkspaceManager.from_state(ctx.session.state)
+    result = await workspace_manager.execute_query(sql_query)
+    if result.is_ok:
+        if result.data:
+            data = result.data
+        else:
+            # non-SELECT query, this should not really happen, because this tool is for running SELECT queries
+            data = SqlSelectData(columns=["message"], rows=[{"message": result.message}])
```
mariankrotil, line 166: When it should not happen. Should we raise an exception or is there a reason we need to handle this case?

src/keboola_mcp_server/sql_tools.py
```diff
@@ -1,12 +1,144 @@
...
+        self._workspace_id: str | None = None
+        self._table_fqn_cache: dict[str, TableFqn] = {}
+
+    async def _get_workspace_id(self) -> str:
+        if self._workspace_id:
+            return self._workspace_id
+
+        for workspace in await self._client.get("workspaces"):
+            assert isinstance(workspace, dict)
+            _id = workspace.get("id")
```
mariankrotil, line 77: If a workspace isn't a dictionary, should we consider skipping it or is this a case it should not occur?

src/keboola_mcp_server/sql_tools.py
```diff
@@ -1,12 +1,144 @@
...
+                db_name = row["current_database"]
+                if "." in table_id:
+                    # a table local in a project for which the snowflake connection/workspace is open
+                    schema_name, table_name = table_id.rsplit(sep=".", maxsplit=1)
+                else:
+                    # a table not in the project, but in the writable schema created for the workspace
+                    # TODO: we should never come here, because the tools for listing tables can only see
+                    #  tables that are in the project
+                    schema_name = row["current_schema"]
+                table_name = table["name"]
```
mariankrotil, line 132: The `table_name` is always overwritten after `if "." in table_id`  branch in the end of `if-else` block
If it is intentional I would do:
`schema_name, _ = table_id.rsplit(sep=".", maxsplit=1)`

mariankrotil (changes requested):

mariankrotil:
Now, when the number of snowflake credentials is reduced, we can improve README instructions for setup of mcp-server using by moving the rest environment variables from the `.env` file directly into the configuration script command and removing it. I can test it and change it after your merge.

src/keboola_mcp_server/sql_tools.py
vita-stejskal: The `workspace` should always be a dictionary. The assert confirms that and also indicates this to python's typing systems, so that your IDE knows that too.

src/keboola_mcp_server/sql_tools.py
vita-stejskal: The query tools is called by the LLM and it is the LLM which generates the SQL passed to the tool. The tool explicitly says that it expects the SQL to be a SELECT statement, but the actual SQL statement passed to the tool by the LLM can be anything. IMO it still makes sense to try to run that SQL (even if it is not a SELECT statement) and pass its data back to the LLM.

tests/test_sql_tools.py
```diff
@@ -0,0 +1,112 @@
...
+def context(client: KeboolaClient, mocker) -> Context:
+    client.get.return_value = [{"id": "workspace_1234", "connection": {"user": "user_1234"}}]
+
+    ctx = mocker.MagicMock(Context)
+    ctx.session = (session := mocker.MagicMock(StatefullServerSession))
+    type(session).state = (state := mocker.PropertyMock())
+    state.return_value = {
+        KeboolaClient.STATE_KEY: client,
+        WorkspaceManager.STATE_KEY: WorkspaceManager(client=client, workspace_user="user_1234"),
+    }
```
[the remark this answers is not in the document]
vita-stejskal: I guess that it's just a matter of style really. I find the use of expression assignments simple and compact. I will rewrite it of course, if you insist.

src/keboola_mcp_server/sql_tools.py
vita-stejskal: Fixed in dcb07fb. You have a keen eye, thanks!

vita-stejskal:
> Now, when the number of snowflake credentials is reduced, we can improve README instructions for setup of mcp-server using by moving the remaining environment variables from the `.env` file directly into the configuration script command and removing it. I can test it and change it after your merge.

I removed the obsolete env vars in 8839902.

I like the idea of getting rid of `.env` file, but it's a separate task/issue which should not be solved here. I'll file a separate task for this. Ideally, I'd want to set the env vars via the `env: { ... }` dictionary in the MCP server's configuration the same way as it is done for non-WSL setup.

tests/test_sql_tools.py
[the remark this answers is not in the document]
mariankrotil: No, it is ok

src/keboola_mcp_server/sql_tools.py
mariankrotil: Sure, you are right.

mariankrotil (approved):

mariankrotil:
> > Now, when the number of snowflake credentials is reduced, we can improve README instructions for setup of mcp-server using by moving the remaining environment variables from the `.env` file directly into the configuration script command and removing it. I can test it and change it after your merge.
> 
> I removed the obsolete env vars in [8839902](https://github.com/keboola/keboola-mcp-server/commit/883990279a0b71dfb870ceed13a9272a2b15f982).
> 
> I like the idea of getting rid of `.env` file, but it's a separate task/issue which should not be solved here. I'll file a separate task for this. Ideally, I'd want to set the env vars via the `env: { ... }` dictionary in the MCP server's configuration the same way as it is done for non-WSL setup.

Yes, leave me the task. However, I suppose I wont be able to use env: {} field directly in the configuration file  as the environment variables are not passed to the Linux system, instead I will add two `export` commands to the bash script commands  within the args:[] configuration field.

PR #20, opened 2025-03-27
add tools for description updates

ncapek:
This PR add (separate) MCP tools for modifying the descriptions of Bucket, Table and Component objects.
Additional methods were also added to the client, since these tools require POSTing to endpoints.


src/keboola_mcp_server/storage_tools.py
```diff
@@ -131,6 +133,26 @@ def set_description(cls, values):
         return values
 
 
+class UpdateBucketDescriptionSuccessResponse(BaseModel):
+    success: bool = True
+    data: List[dict]
```
vita-stejskal, line 138: What is in this list?

src/keboola_mcp_server/storage_tools.py
```diff
@@ -131,6 +133,26 @@ def set_description(cls, values):
...
+
+
+class UpdateBucketDescriptionErrorResponse(BaseModel):
```
vita-stejskal, line 141: Do we really need these special error response DTOs? So far, it seemed ok to just let a tool to raise exception. The agent seems to handle the exceptions correctly -- understands that there was an error and tries to correct its actions.

src/keboola_mcp_server/storage_tools.py
```diff
@@ -131,6 +133,39 @@ def set_description(cls, values):
...
+    def extract_metadata(cls, values):
+        metadata = values.get("metadata", [])
+        if isinstance(metadata, list) and metadata:
+            entry = metadata[0]
+            return {
+                "success": True,
+                "description": entry.get("value"),
+                "timestamp": entry.get("timestamp"),
+            }
+
```
mariankrotil, line 167: What if the if-statement is false? Should not we return anything, at least `values` ?

mariankrotil: The same for the `UpdateBucketDescriptionResponse.extract_from_response`

src/keboola_mcp_server/storage_tools.py
```diff
@@ -211,79 +232,45 @@ async def update_bucket_description(
...
-):
+) -> UpdateBucketDescriptionResponse:
     """
     Update the description for a given Keboola bucket.
 
     Args:
         bucket_id: The ID of the bucket to update.
         description: The new description for the bucket.
-
+        ctx: The request context with session state.
```
vita-stejskal, line 242: This docstring is used as the tool description and sent to the LLM. It should not contain the `Args`, because they are described in the function signature. And we definitely should not tell the LLM about `ctx` parameter, which is internal in the MCP server.

src/keboola_mcp_server/storage_tools.py
```diff
@@ -211,79 +232,45 @@ async def update_bucket_description(
...
     Returns:
-        The response from the API.
+        A validated UpdateBucketDescriptionResponse instance.
```
vita-stejskal, line 244: This should also be attached to the return type via `Annotated[UpdateBucketDescriptionResponse, Field(description="...")]`. And since it does not tell much more than the returned DTO type, I would not even bother mentioning it here in the docstring again.

The same comments apply to `update_table_description()` tool.

tests/test_storage_tools.py
```diff
@@ -177,3 +211,46 @@ async def test_get_table_metadata(mcp_context_client, mock_table_data) -> None:
...
+
+
+@pytest.mark.asyncio
+async def test_update_table_description_success(
+    mcp_context_client, mock_update_table_description_response
+) -> None:
+    """Test successful update of table description."""
+
+    # Mock the Keboola client post method
+    keboola_client = mcp_context_client.session.state["sapi_client"]
```
vita-stejskal, line 244:
```suggestion
    keboola_client = KeboolaClient.from_state(mcp_context_client.session.state)
```

tests/test_storage_tools.py
```diff
@@ -177,3 +211,46 @@ async def test_get_table_metadata(mcp_context_client, mock_table_data) -> None:
...
+    result = await update_table_description(
+        table_id="in.c-test.test-table",
+        description="Updated test description",
+        ctx=mcp_context_client,
+    )
+
+    assert isinstance(result, UpdateTableDescriptionResponse)
+    assert result.success is True
+    assert result.description == "Updated test description"
+    assert result.timestamp == "2025-04-07T16:47:18+0200"
```
vita-stejskal, line 256: It would be good to add `keboola_client.post.assert_called_once_with(...)`.

tests/test_storage_tools.py
```diff
@@ -177,3 +211,46 @@ async def test_get_table_metadata(mcp_context_client, mock_table_data) -> None:
...
         assert col_id.db_identifier == exp_col_id["db_identifier"]
+
+
+@pytest.mark.asyncio
+async def test_update_bucket_description_success(
+    mcp_context_client, mock_update_bucket_description_response
+) -> None:
+    """Test successful update of bucket description."""
+
+    keboola_client = mcp_context_client.session.state["sapi_client"]
```
vita-stejskal, line 222:
```suggestion
    keboola_client = KeboolaClient.from_state(mcp_context_client.session.state)
```

tests/test_storage_tools.py
```diff
@@ -177,3 +211,46 @@ async def test_get_table_metadata(mcp_context_client, mock_table_data) -> None:
...
+    result = await update_bucket_description(
+        bucket_id="in.c-test.bucket-id",
+        description="Updated bucket description",
+        ctx=mcp_context_client,
+    )
+
+    assert isinstance(result, UpdateBucketDescriptionResponse)
+    assert result.success is True
+    assert result.description == "Updated bucket description"
+    assert result.timestamp == "2025-04-07T17:47:18+0200"
```
vita-stejskal, line 234: It would be good to add `keboola_client.post.assert_called_once_with(...)`.

vita-stejskal (approved):

PR #21, opened 2025-03-28
Kab 904 mcp using wsl readme

mariankrotil:
This PR is a copy of the previous one adding instructions to readme on how to run mcp server from WSL, but with a correct Jira task ID and minor changes which facilitate the usage.
-> The script wrapper that was supposed to run the mcp server is removed since it can be accomplished with raw commands.
-> The problem I had was that the script commands which are used in the WSLinux has to be encapsulated with ' quotes because the Windows and Linux have slightly different line parsers, and windows parsed the arguments which were expected to be passed to Linux parser - the single quotes solved this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,48 @@ To use this server with Cursor AI, you have two options for configuring the tran
 }
 ```
 
+### Option 3: Using WSL Standard I/O (wsl stdio)
+
+```json
+{
+  "mcpServers": {
+    "keboola_wsl": {
+        "command": "wsl.exe",
+        "args": [
+            "bash",
+            "-c",
+            "'source ~/wsl/path/to/the/script/run_mcp.sh'"
+        ],
+    }
+  }
+}
+```
+
+- where `run_mcp.sh` follows:
+```shell
+#!/bin/bash
+
+PATH_TO_THE_ENV_VARS=~/wsl/path/to/the/file/.env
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -162,6 +162,38 @@ To use this server with Cursor AI, you have two options for configuring the tran
...
+        "/wsl_path/to/keboola-mcp-server/venv/bin/python -m keboola_mcp_server.cli --transport stdio'"
+      ]
+    }
+  }
+}
+```
+- where `/wsl_path/to/keboola-mcp-server/.env` file contains environment variables:
+```shell
+export KBC_STORAGE_TOKEN="your-keboola-storage-token"
+export PYTHONPATH="/wsl_path/to/keboola-mcp-server/src"
```
vita-stejskal, line 187: Is this really needed? I know it is mentioned in the Claude instructions, but I don't think that it is needed. Could you please check that it really is needed and explain why (or remove it, if it's not needed)? Thanks

mariankrotil: Do you mean the PYTHONPATH environment var? 
It works without configuring the python path env var explicitly

mariankrotil: Even Claude desktop works

vita-stejskal: Thanks

vita-stejskal (approved):

PR #22, opened 2025-03-31
Kab 911 refactor storage tools

ncapek:
This PR is concerned with separating the storage related tools into a separate module from the `server.py` module.

### DIFF
--- a/src/keboola_mcp_server/storage_tools.py
+++ b/src/keboola_mcp_server/storage_tools.py
@@ -30,12 +30,6 @@ class BucketInfo(BaseModel):
         None, description="Stage of the bucket ('in' for input stage, 'out' for output stage)"
     )
     created: str = Field(description="Creation timestamp of the bucket")
-    # table_count: Optional[int] = Field(
-    #     None,
-    #     description="Number of tables in the bucket",
-    #     validation_alias=AliasChoices("tableCount", "table_count", "table-count"),
-    #     serialization_alias="tableCount",
-    # )
     data_size_bytes: Optional[int] = Field(
         None,
         description="Total data size of the bucket in bytes",
@@ -43,12 +37,20 @@ class BucketInfo(BaseModel):
         serialization_alias="dataSizeBytes",
     )
 
-    tables_count: int = 0  # Default value, updated dynamically
+    tables_count: Optional[int] = Field(
+        default=None,
+        description="Number of tables in the bucket",
+        validation_alias=AliasChoices("tablesCount", "tables_count", "tables-count"),
+        serialization_alias="tablesCount",
+    )
 
     @model_validator(mode="before")
     @classmethod
     def set_table_count(cls, values):
-        values["tables_count"] = len(values.get("tables", []))  # Count tables if provided
+        if isinstance(values.get("tables"), list):
+            values["tables_count"] = len(values["tables"])
+        else:
+            values["tables_count"] = None
         return values
 
 
@@ -106,8 +108,9 @@ async def get_bucket_metadata(
     """Get detailed information about a specific bucket."""
     client = ctx.session.state["sapi_client"]
     assert isinstance(client, KeboolaClient)
+    print(client.storage_client.buckets.detail(bucket_id))
     raw_bucket = cast(Dict[str, Any], client.storage_client.buckets.detail(bucket_id))
-    print(raw_bucket)
+    print(f"Inside bucket metadata: {raw_bucket}")
 
     return BucketInfo(**raw_bucket)
 
@@ -117,6 +120,7 @@ async def list_bucket_info(ctx: Context) -> List[BucketInfo]:
     client = ctx.session.state["sapi_client"]
     assert isinstance(client, KeboolaClient)
     raw_bucket_data = client.storage_client.buckets.list()
+    print(f"Inside list bucket info: {raw_bucket_data}")
 
     return [BucketInfo(**raw_bucket) for raw_bucket in raw_bucket_data]
 
--- a/src/keboola_mcp_server/storage_tools.py
+++ b/src/keboola_mcp_server/storage_tools.py
@@ -51,6 +51,7 @@ class BucketInfo(BaseModel):
         values["tables_count"] = len(values.get("tables", []))  # Count tables if provided
         return values
 
+
 class TableColumnInfo(BaseModel):
     name: str = Field(description="Name of the column")
     db_identifier: str = Field(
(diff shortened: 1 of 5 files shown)


ncapek:
@vita-stejskal Shall I also move the tools related to description updating into the `storage_tools` within this PR? If so, we can abandon the other PR and just deal with everything here. It might be cleaner in the end.

vita-stejskal:
> @vita-stejskal Shall I also move the tools related to description updating into the `storage_tools` within this PR? If so, we can abandon the other PR and just deal with everything here. It might be cleaner in the end.

I'd do that separately. Let's continue with the PRs that you've already opened.

vita-stejskal (approved):

vita-stejskal:
Please increment the project version.

ncapek:
> I assume that there have been no changes in the code/implementation and that you just moved the stuff to `storage_tools.py` file. The exception to that are the tests. Am I correct?

Additionally the BucketInfo model class now has the fixed logic to deal with `tables_count`. And I'm still working out how to correctly handle the `get_table_metadata` tool since it seems that there were some changes in the DatabasePathManager methods relating to the table FQN.

ncapek:
> Please increment the project version.

As a general rule, I always increment the versions right before merging. Because if someone merges before me, I have to update that file again.

ncapek:
> > I assume that there have been no changes in the code/implementation and that you just moved the stuff to `storage_tools.py` file. The exception to that are the tests. Am I correct?
> 
> Additionally, support for the `tables_count` calculation was added. Also, I want to confirm that the table columns are being constructed correctly. I'm a little confused about the column name vs column dbIdentifier distinction. Sample tool response: `{"id": "in.c-foo.foo", "name": "foo", "primaryKey": [], "created": "2024-04-03T13:57:52+0200", "rowCount": null, "dataSizeBytes": 1536, "columns": ["id", "text"], "columnIdentifiers": [{"name": "id", "dbIdentifier": "\"id\""}, {"name": "text", "dbIdentifier": "\"text\""}], "dbIdentifier": "\"SAPI_9909\".\"in.c-foo\".\"foo\""}`

I looked into and updated the test accordingly. Should be fine.

ncapek:
> I assume that there have been no changes in the code/implementation and that you just moved the stuff to `storage_tools.py` file. The exception to that are the tests. Am I correct?

Finally, I added support for better derivation of the `description` field. Now it takes it from the description field and if not present, tries to take it from the metadata.

However for buckets, this only seems to work the the `get_bucket_tool` because in `list_bucket_info` the client does not return metadata by default and the client does not seem to support additional arguments - `client.storage_client.buckets.list()`. However for tables it seems to be different, there it is returning metadata - `client.storage_client.buckets.list_tables(bucket_id)`

vita-stejskal:
It would be so much easier to review if you moved the tools without any change to the new location and then did the improvements in a separate PR.

ncapek:
> It would be so much easier to review if you moved the tools without any change to the new location and then did the improvements in a separate PR.

Sorry for the complication. I considered doing it in two separate PRs, but I couldn't find a clean way of doing it. Since the changes involve the same code, it would most likely lead to conflicts and potentially a larger mess. But perhaps there is a way to design the branches such that this wouldn't be an issue.

vita-stejskal:
> Sorry for the complication. 

It's not a big deal.

> But perhaps there is a way to design the branches such that this wouldn't be an issue.

I think that if you branched off the "improvements" branch of the "tools moved" branch rather than the "master" branch, then that would do the trick.

src/keboola_mcp_server/storage_tools.py
```diff
@@ -0,0 +1,180 @@
...
+        values["description"] = extract_description(values)
+        return values
+
+
+async def get_bucket_metadata(
+    bucket_id: Annotated[str, Field(description="Unique ID of the bucket.")], ctx: Context
+) -> BucketInfo:
+    """Get detailed information about a specific bucket."""
+    client = ctx.session.state["sapi_client"]
+    assert isinstance(client, KeboolaClient)
```
vita-stejskal, line 136: You should now be able to use:
```
```

src/keboola_mcp_server/storage_tools.py
```diff
@@ -0,0 +1,180 @@
...
+    """Get detailed information about a specific table including its DB identifier and column information."""
+    client = ctx.session.state["sapi_client"]
+    assert isinstance(client, KeboolaClient)
+    raw_table = cast(Dict[str, Any], client.storage_client.tables.detail(table_id))
+
+    columns = raw_table.get("columns", [])
+    column_info = [TableColumnInfo(name=col, db_identifier=f'"{col}"') for col in columns]
+
+    workspace_manager = ctx.session.state["workspace_manager"]
+    assert isinstance(workspace_manager, WorkspaceManager)
```
vita-stejskal, line 163: You should now be able to use:
```
workspace_manager = WorkspaceManager.from_state(ctx.sessino.state)
```

tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+
+@pytest.fixture
+def mcp_context() -> Context:
+    """Create a mock context."""
+    context = MagicMock(spec=Context)
+    return context
+
+
+@pytest.fixture
+def mcp_context_client(keboola_client: KeboolaClient, mcp_context: Context) -> Context:
```
vita-stejskal, line 27: It seems that you could remove this fixture and do its job directly in the `map_context()` fixture. The `mcp_fixture()` by itself does not seem to be used anywhere else then just here.

tests/test_storage_tools.py
```diff
@@ -0,0 +1,181 @@
...
+    assert isinstance(result, TableDetail)
+    assert result.id == mock_table_data["raw_table_data"]["id"]
+    assert result.name == mock_table_data["raw_table_data"]["name"]
+    assert result.primary_key == mock_table_data["raw_table_data"]["primary_key"]
+    assert result.rows_count == mock_table_data["raw_table_data"]["rows_count"]
+    assert result.data_size_bytes == mock_table_data["raw_table_data"]["data_size_bytes"]
+    assert result.columns == mock_table_data["raw_table_data"]["columns"]
+    assert result.db_identifier == mock_table_data["additional_data"]["snowflake_fqn"]
+    print(result.column_identifiers)
+    print(mock_table_data["additional_data"]["column_identifiers"])
```
vita-stejskal, line 176: I assume these are just `print` statements for debugging and so they could be removed.

vita-stejskal (dismissed):

mariankrotil (approved):

PR #23, opened 2025-03-31
Kab 871 mcp ro tools3

mariankrotil:
This PR is checkout from the branch `KAB-871-mcp-ro-component-tools2` . New tools for component and jobs are added.
Component tools:
- list components / component configs
- get components  / component configs / metadata details 
Jobs tools (using queue Job api):
- list most recent jobs
- list most recent jobs for component / component & its config  
   - additional parameters for list jobs functions = filter by status, restrict number via limit param and shift results by offset
 - get jobs detail

Here following commits will 
- add  sort by parameter like startTime, endTime, createdTime - asc and desc order parameter

Future JIRA task will 
- unify component and component config id as a component
- distinguish between components and transformations
- add difference between types extractors & writers for components..... transformations / applications for transformations


mariankrotil:
example of usage
image

PR #30, opened 2025-04-06
Kab 906 big query

vita-stejskal:
This adds dependency on `google-cloud-bigquery` python package and uses its `Client` to run SQL queries on the BigQuery workspace.

* (fixed already) IMPORTANT -- current implementation does not support tables that are linked from other projects to the project with your BigQuery workspace. This will be added later.
* The instructions for setting MCP server have changed. Instead of `KBC_WORKSPACE_USER` you need to specify `KBC_WORKSPACE_SCHEMA`. BigQuery workspaces don't have the "user" setting/credential.
* Besides of `KBC_WORKSPACE_SCHEMA` you will also have to supply `GOOGLE_APPLICATION_CREDENTIALS` environment variable. The instructions are in `README.md`
* There is a new tool called `get_sql_dialect` which provides info on the SQL dialect used by the backend database. The `query_table` tool gives the LLM instructions to check the SQL dialect.
* I've updated the `TableDetail` and `TableColumnInfo` dataclasses that are returned from the storage tools. They now use fields such as `fullyQualifiedName` (tables) and `quotedName` (columns).

It seems to work fine in a basic project. I tested it in [this Keboola project](https://connection.us-east4.gcp.keboola.com/admin/projects/498/dashboard) which contains `country-capitals` table in a `foo` bucket and I asked the Cursor AI agent the following question:
```
What's the distance between the westernmost and easternmost country capitals?
```
It called multiple MCP tools to first check what tables are available, what SQL dialect it should use and then conjured a fairly complex BQ statement which involved `ST_DISTANCE` and `ST_GEOGPOINT` functions. The query finished successfully and agent's response was correct.


jordanrburger (dismissed):
Tested! All is working well. Thanks, Vita 🔥

tests/test_storage_tools.py
```diff
@@ -48,18 +46,20 @@ def mock_table_data() -> Dict[str, Any]:
...
+                TableColumnInfo(name="name", quoted_name="#name#"),
+                TableColumnInfo(name="value", quoted_name="#value#"),
             ],
         },
     }
 
 
 @pytest.fixture
-def mock_buckets() -> List[Dict[str, Any]]:
+def mock_buckets() -> Sequence[Mapping[str, Any]]:
```
ncapek, line 62: Why did you decide on Sequence instead of List?

tests/test_storage_tools.py
```diff
@@ -48,18 +46,20 @@ def mock_table_data() -> Dict[str, Any]:
     return {
         "raw_table_data": raw_table_data,  # What the client returns
         "additional_data": {
-            "snowflake_fqn": '"SAPI_TEST"."in.c-test"."test-table"',  # What workspace_manager should return
-            "column_identifiers": [  # Expected transformed column identifiers
-                {"name": "id", "db_identifier": '"id"'},
-                {"name": "name", "db_identifier": '"name"'},
-                {"name": "value", "db_identifier": '"value"'},
+            # What workspace_manager should return
+            "table_fqn": TableFqn("SAPI_TEST", "in.c-test", "test-table", quote_char="#"),
```
ncapek, line 50: Why did you decide to change this to objects instead of raw data? I thought the point of having raw data was to test that the constructors work correctly.

ncapek (dismissed):

tests/test_sql_tools.py
```diff
@@ -59,15 +49,77 @@ def context(client: KeboolaClient, mocker) -> Context:
...
+async def test_query_table(query: str, result: QueryResult, expected: str, mocker):
+    workspace_manager = mocker.AsyncMock(WorkspaceManager)
+    workspace_manager.execute_query.return_value = result
 
+    ctx = mocker.MagicMock(Context)
+    ctx.session = (session := mocker.MagicMock(StatefullServerSession))
+    type(session).state = (state := mocker.PropertyMock())
+    state.return_value = {
+        WorkspaceManager.STATE_KEY: workspace_manager,
+    }
```
mariankrotil, line 61: We can use fixture out of this code since it is used multiple times in this file.

tests/test_storage_tools.py
vita-stejskal: The fixtures should be immutable. Mutating them accidentally in a test may affect the test itself or other tests too (if the fixture has the scope broader than just "function"). The `Sequence`/`Mapping` type hints won't make the return value immutable of course, but at least they indicate to your IDE that the value should not be mutated and your IDE will hopefully show a warning in the editor if you try to mutate the fixture.

tests/test_storage_tools.py
vita-stejskal: My main motivation was to get rid of this:
```
    mock_fqn = MagicMock()
    mock_fqn.snowflake_fqn = mock_table_data["additional_data"]["snowflake_fqn"]
```
It mocks `TableFqn` dataclass which is pointless (you can simply just instantiate that class) and dangerous (because it hard-wires `snowflake_fqn` function to the test). If `TableFqn` gets changed and has this function removed/renamed the test will still work. The IDE's refactoring system won't help you either, because it does not understand that you are actually mocking `TableFqn` dataclass.

Other reasons is that this fixure is used to test `get_table_metadata()` function and it should do just that. It should not be testing `TableFqn` and its deserialization.

Finally, I still don't like this `mock_table_data()` fixture. It should IMO be turned into the values of `@pytest.mark.parametrize()` instead. It's values are specific to a single test and it combines both the "mock inputs" of the test (`raw_table_data` and `table_fqn`) with the expected results (`columns`). Which makes it hardly reusable.

tests/test_sql_tools.py
vita-stejskal: The problem is in the mocked `WorkspaceManager` instance. It is different for each test. But I agree that there is a bit too much boilerplate code and I should try to reduce it.

tests/test_sql_tools.py
```diff
@@ -94,19 +150,175 @@ class TestWorkspaceManager:
...
+                ),
+            ),
+            # (
+            #         "create table foo (id integer, name varchar);",
+            #         QueryResult(status="ok", message="1 table created"),
+            # ),
+            # (
+            #         "bla bla bla",
+            #         QueryResult(status="error", message="Invalid SQL..."),
+            # ),
```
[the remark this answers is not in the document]
vita-stejskal: Fixed in 35a40b6.

ncapek (approved):

PR #36, opened 2025-04-09
Kab 961 unify tool naming

ncapek:
This PR is concerned with unifying the naming of MCP tools and related classes to be consistent across tool types.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 license = "MIT"
 authors = [
-    { name = "Keboola", email = "support@keboola.com" }
+    { name = "Keboola", email = "devel@keboola.com" }
 ]
 dependencies = [
     "mcp[cli] ~= 1.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "0.4.1"
+version = "0.5.0"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "0.4.0"
+version = "0.4.1"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/src/keboola_mcp_server/storage_tools.py
+++ b/src/keboola_mcp_server/storage_tools.py
@@ -15,9 +15,9 @@ logger = logging.getLogger(__name__)
 def add_storage_tools(mcp: FastMCP) -> None:
     """Adds tools to the MCP server."""
     mcp.add_tool(get_bucket_detail)
-    mcp.add_tool(retrieve_buckets_in_project)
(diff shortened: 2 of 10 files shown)


src/keboola_mcp_server/storage_tools.py
```diff
@@ -128,27 +128,27 @@ def set_description(cls, values):
...
     assert isinstance(client, KeboolaClient)
     raw_bucket = cast(Dict[str, Any], client.storage_client.buckets.detail(bucket_id))
 
-    return BucketInfo(**raw_bucket)
+    return BucketDetail(**raw_bucket)
 
 
-async def list_bucket_info(ctx: Context) -> List[BucketInfo]:
-    """List information about all buckets in the project."""
+async def retrieve_buckets_in_project(ctx: Context) -> List[BucketDetail]:
```
vita-stejskal, line 142: Is the `retrieve_` (instead of `list_` or `get_`) prefix any better for the LLM? I thought that using `list_` prefix made a good sense for tools that actually return lists of things? But maybe it does not. Or maybe it does not matter.

On the same note -- is `_in_project` suffix significant for the LLM? Does adding it make any difference for the LLM? All our tools work in the scope of a single Keboola project. Should we amplify this fact by adding `_in_project` to the names of all our tools? It seems dubious, but I could be wrong.

src/keboola_mcp_server/storage_tools.py
```diff
@@ -128,27 +128,27 @@ def set_description(cls, values):
...
+    """Retrieve information about all buckets in the project."""
```
vita-stejskal, line 143: While unifying things what do you thing about the use of the 3rd person singular (i.e. retrieves, gets, lists) versus the 2nd person singular (imperative) (i.e. retrieve, get, list)? Do we want to unify this? If so, then which form do we want to use?

Typically, docs guides prefer the 3rd person singular. People are very inconsistent in following this guideline and the LLM probably does not care much. But I personally find the 3rd person singular form more correct and logincal, because the docs says what the function does -- i.e. The function "retrieves" stuff. It also feels less ambiguous than the imperative form which is often the same as the infinitive form.

vita-stejskal (approved):

src/keboola_mcp_server/storage_tools.py
ncapek: We had some discussions with @mariankrotil about these things over the last couple of days. The main motivation to not use `list` as a verb in the function names was that it often lead to some incredibly awkward docstrings or annotations. The `_in_project` suffix is only because we expected that some use cases might be broader than a single project. For example updating/creating a Component (the real core component) is probably broader than a single project. But for now, we can get rid of this suffix.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "0.4.0"
+version = "0.4.1"
```
radcliffkey, line 7: Renaming is an incompatible change, so theoretically we should increment the major version, but I'm in favour of `0.5.0`. Definitely this is not just a patch.

ncapek: Good point. I will use `0.5.0`

src/keboola_mcp_server/storage_tools.py
ncapek: I switched to the third person variant of the docstings.

src/keboola_mcp_server/storage_tools.py
ncapek: Removed

vita-stejskal (approved):

PR #37, opened 2025-04-11
KAB-874-create-sql-transformation-tool

mariankrotil:
Modification based on the discussion.

This PR adds tool for creating SQL Transformation based on the utilized backend - Snowflake and BigQuery backends are supported.  For now, the tool only creates sql transformation without output mappings, which have to be set later by the user in UI when required by the query. Input mappings are incorporated within the sql query via fully qualified ids.
 
- create_sql_transformation with following params:
    - name - of the new transformation, if specified otherwise generated by llm
    - description - of the new transformation, if specified otherwise generated by llm
    - sql_statemet - the sql query itself which should follow current backend's sql dialect. - generated by llm


src/keboola_mcp_server/component_tools.py
```diff
@@ -512,4 +608,98 @@ async def get_component_configuration_details(
...
+            component_id=transformation_id,
+            component=component,
+        )
+        logger.info(
+            f"Created new transformation configuration: {new_transformation_configuration.configuration_id} for "
+            f"component: {transformation_id}."
+        )
+        return new_transformation_configuration
+    except Exception as e:
+        logger.error(f"Error when creating new transformation configuration: {e}")
```
vita-stejskal, line 701: It's better to use `logger.exception(...)` which also logs the exception traceback.

src/keboola_mcp_server/component_tools.py
```diff
@@ -512,4 +608,98 @@ async def get_component_configuration_details(
...
+    ComponentConfiguration,
+    Field(
+        description="Newly created SQL Transformation Configuration.",
+    ),
+]:
+    """
+    Creates an SQL transformation using the specified name, SQL query following the current SQL dialect, and a detailed
+    description.
+    CONSIDERATIONS:
+        - The SQL query statement is executable and must follow the current SQL dialect (BigQuery, Snowflake), which
```
vita-stejskal, line 644: I wouldn't mention the dialects explicitly. When the list of the supported dialects changes, this instruction will become obsolete or somebody will have to remember to update it.

tests/test_component_tools.py
```diff
@@ -409,3 +413,121 @@ def test_handle_component_types(
...
+@pytest.mark.asyncio
+async def test_create_transformation_configuration(
+    mcp_context_components_configs: Context,
+    mock_component: dict[str, Any],
+    mock_configuration: dict[str, Any],
+    sql_dialect: str,
+    expected_component_id: str,
+    expected_configuration_id: str,
+    expected_error: bool,
+    test_branch_id: str,
```
vita-stejskal, line 435: Although the `test_branch_id()` fixture was not added in the PR, it's name is unfortunate. The `test_*` prefix should be reserved for the test functions.

vita-stejskal (approved):

tests/test_component_tools.py
mariankrotil: renamed here: e1bdb1d340a5c018652312f8f9ab0bd7fc9609f5

src/keboola_mcp_server/component_tools.py
mariankrotil: Fixed here: e1bdb1d340a5c018652312f8f9ab0bd7fc9609f5

src/keboola_mcp_server/component_tools.py
mariankrotil: Ok, removed here: e1bdb1d340a5c018652312f8f9ab0bd7fc9609f5

tests/test_component_tools.py
mariankrotil: using `mock_branch_id`

PR #39, opened 2025-04-16
Kab 873 start a component job tool

mariankrotil:
This PR adds a tool for starting a new job to the MCP.  PR also removes `metrics` field from the job detail model as it contains metadata that are irrelevant to the user, but adds a new JOB_STATUS enum field `created` because the newly created job is annotated respectively.

start_new_job tool accepts:
-  component id  = for which component / transformation the job should run
- configuraiton id = for which configuration the job should run
and returns:
- newly created job detail with the new job id.

Question: Should we consider also configuration raw ids in the tool's parameters ?

### DIFF
--- a/tests/test_jobs_tools.py
+++ b/tests/test_jobs_tools.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Annotated, Any, Type, Union
+from typing import Any, Type, Union
 from unittest.mock import MagicMock
 
 import pytest
@@ -12,12 +12,13 @@ from keboola_mcp_server.jobs_tools import (
     JobListItem,
     get_job_detail,
     retrieve_jobs,
-    create_job_run,
+    start_job,
 )
 
 
 @pytest.fixture
 def mock_jobs() -> list[dict[str, Any]]:
+    """list of mock jobs - simulating the api response."""
     return [
         {
             "id": "123",
@@ -46,6 +47,7 @@ def mock_jobs() -> list[dict[str, Any]]:
 
 @pytest.fixture
 def mock_job() -> dict[str, Any]:
+    """mock job - simulating the api response."""
     return {
         "id": "123",
         "status": "success",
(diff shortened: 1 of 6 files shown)


ncapek:
> Question: Should we consider also configuration raw ids in the tool's parameters ?

I don't know what that is.

tests/test_jobs_tools.py
```diff
@@ -230,3 +231,84 @@ async def retrieve_jobs_with_component_id_without_config_id(
...
+        ctx=context, component_id=component_id, configuration_id=configuration_id
+    )
+
+    assert isinstance(job_detail, JobDetail)
+    assert job_detail.result == {}
+    assert job_detail.id == mock_job["id"]
+    assert job_detail.status == mock_job["status"]
+    assert job_detail.component_id == component_id
+    assert job_detail.config_id == configuration_id
+    assert job_detail.result == {}
```
ncapek, line 257: All of this stuff can be checked by pydantic's `model_validate_json()`, you don't have to assert every field individually.

tests/test_jobs_tools.py
```diff
@@ -230,3 +231,84 @@ async def retrieve_jobs_with_component_id_without_config_id(
...
+        (None, False, {}),  # None is valid and converted to {}
+        (
+            ["result1", "result2"],
+            True,
+            ValueError,
+        ),  # list is not a valid result type, we raise an error
+    ],
+)
+def test_job_detail_model_validate_for_result_field(
+    result_type: Union[list, dict, None],
```
ncapek, line 301: Can a result really be all of these? Seems like bad design to have so many possible output types.

vita-stejskal (approved):
Please rename the tool and fix the descriptions in its signature.

tests/test_jobs_tools.py
```diff
@@ -230,3 +231,87 @@ async def retrieve_jobs_with_component_id_without_config_id(
...
+    is_exception: bool,
+    expected_result: Union[dict, Type[Exception]],
+    mock_job: dict[str, Any],
+):
+    """Tests JobDetail model validate for result field."""
+    mock_job["result"] = result_type
+    if is_exception:
+        assert isinstance(expected_result, type) and issubclass(expected_result, Exception)
```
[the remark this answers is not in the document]
mariankrotil: here: d6c6f3da6ea010de68fb3bc22787b4e725418b37

mariankrotil:
> I see that `JobDetail` still has Optional field values. We had a discussion in Slack that the current proper way to do this is to have empty base type values, e.g. "" for no string. I didn't personally test this, but this is how other handled it in their code.

I assume it was meant for tool parameters, not for DTOs.

PR #40, opened 2025-04-16
Kab 874 create sql transformation

mariankrotil:
This PR adds the missing parameter for specifying created table names within SQL query in the create SQL Transformation tool and enables to configure output mappings directly in the mcp server (automatically). The output bucket name is is derived from the transformation name e.g. `transformation_name='example transformation name' --> bucket_name='example-transformation-name'` and the destination where a created table is stored is then e.g.  `table_name='table-name' ---> destination='out.c-example-transformation-name.table-name'`

now the `create_sql_transformation` tool accepts:
- transformation name (generated by llm if not specified)
- transformation description (generated by llm if not specified)
- sql statements (generated by llm. LLM should consider following: each statement is a separate item in the list and should follow current SQL dialect. Input tables should be referenced by fully qualified ids, created output tables within query should be only names)
- created table names (generated by llm, This list should be empty if no table creation is applied within query otherwise it contains plain table names created in the query)

### DIFF
--- a/src/keboola_mcp_server/component_tools.py
+++ b/src/keboola_mcp_server/component_tools.py
@@ -707,9 +707,7 @@ async def create_sql_transformation(
     )
 
     client = KeboolaClient.from_state(ctx.session.state)
-    endpoint = (
-        f"branch/{client.storage_client._branch_id}/components/{transformation_id}/configs"
-    )
+    endpoint = f"branch/{client.storage_client._branch_id}/components/{transformation_id}/configs"
 
     logger.info(
         f"Creating new transformation configuration: {name} for component: {transformation_id}."
--- a/src/keboola_mcp_server/component_tools.py
+++ b/src/keboola_mcp_server/component_tools.py
@@ -652,7 +652,7 @@ async def create_sql_transformation(
         Sequence[str],
         Field(
             description=(
-                "The SQL executable query statements written in the current SQL dialect."
+                "The executable SQL query statements written in the current SQL dialect. "
                 "Each statement should be a separate item in the list."
             ),
         ),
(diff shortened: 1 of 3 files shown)


mariankrotil:
I am providing screenshots of example how the tool can be used.
Screenshot 2025-04-16 134847
Screenshot 2025-04-16 134911

src/keboola_mcp_server/component_tools.py
```diff
@@ -424,28 +427,48 @@ class Table(BaseModel):
...
+    statements: Sequence[str], transformation_name: str, output_table_mappings: Sequence[str]
+) -> TransformationConfiguration:
     """
-    Utility function to set the transformation configuration from SQL statement.
+    Utility function to set the transformation configuration from code statements.
     It creates the expected configuration for the transformation, parameters and storage.
-    :param sql_statement: The SQL statements
+    :param statements: The code statements (sql for now)
+    :param transformation_name: The name of the transformation from which the bucket name is derived as in the UI
+    :param output_table_mappings: The output table mappings created tables in the transformation
```
vita-stejskal, line 438:
```suggestion
    :param output_tables: The output tables of the transformation, created by the code statements
```

tests/test_component_tools.py
```diff
@@ -514,26 +524,66 @@ async def test_create_transformation_configuration_fail(
...
+            "out.c-test-name",
+        ),
+    ],
+)
 def test_get_transformation_configuration(
-    sql_statement: str,
+    sql_statements: list[str],
+    created_table_names: list[str],
+    transformation_name: str,
+    expected_destination: str,
```
vita-stejskal, line 557: So, the `expected_destination` is in fact the `expected_bucket_id`, right?

vita-stejskal (approved):

src/keboola_mcp_server/component_tools.py
mariankrotil: Fixed here: 65c22bb56341df5458c30093633220ed0cf875cb

PR #41, opened 2025-04-22
Kab 980 wheel package

vita-stejskal:
This builds the wheels package, but does not publish it to pypi.org yet. I'll add that in another PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
         uses: dorny/test-reporter@v1
         if: success() || failure() # run this step even if a previous step failed
         with:
-          name: Test results
+          name: Test results (${{ matrix.python-version }})
           path: ./test-results.xml
           reporter: 'java-junit'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,8 @@ on:
     branches: [ main ]
 
 jobs:
-  test:
+  build:
+    name: Build, test and package
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


mariankrotil (approved):

.github/workflows/ci.yml
```diff
@@ -6,51 +6,67 @@ on:
...
+          python -m build --wheel
+
+      - name: Get package version
+        id: get_version
+        run: |
+          VERSION=`python3 -c 'import importlib.metadata; print(importlib.metadata.version("keboola_mcp_server"))'`
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+
+      - name: Upload wheel package
+        if: matrix.python-version == '3.10'
```
odinuv, line 67: přemýšlím jak tuhle podmínku napsat jinak - až dropneš support pro 3.10, vyhodíš to z matrix a přestane se to releasovat :)

odinuv (approved):

radcliffkey (approved):
Can you also remove `uv.lock` file from the repo?
- we're not using uv
- it contains old dep versions

.github/workflows/ci.yml
vita-stejskal: Well, the condition will have to be updated when the matrix is changed due to dropping the support of Python 3.10. It's not nice, but there will surely be other traps too.

vita-stejskal:
> Can you also remove `uv.lock` file from the repo?

Yes, will do that, but in another PR.

.github/workflows/ci.yml
odinuv: yup, my point was that this is sort of invisible - drop support, and then few weeks later you realize that no releaseses are produced :)

PR #43, opened 2025-04-23
Kab 980 push to pypi org

vita-stejskal:
This adds the job to push `keboola_mcp_server` wheels package to `pypi.org`. I've tested it all up to the final step that actually does the push. That I plan to test in `master` branch after smoke testing the `development` branch and merging into `master` (including the changes in `ci.yaml` from this PR).

The `deploy_to_pypi` step is only run when the CI workflow runs from a GitHub "release" tag -- i.e. a tag such as `vX.Y.Z` and its `X.Y.Z` part is identical to the version in `pyproject.toml`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,10 @@ jobs:
     needs:
       - build
     runs-on: ubuntu-latest
-    if: startsWith(github.ref, 'refs/tags/') && needs.build.outputs.is_semantic_tag == 'true'
+    if: |
+      startsWith(github.ref, 'refs/tags/') &&
+      needs.build.outputs.is_semantic_tag == 'true' &&
+      needs.build.outputs.tag == format('v{0}', needs.build.outputs.version)
     steps:
       - name: Download wheel package
         uses: actions/download-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,19 +96,8 @@ jobs:
           name: keboola_mcp_server-${{ needs.build.outputs.version }}-py3-none-any.whl
           path: dist/
 
-      - name: List dist files
-        run: |
-          ls -la dist/
-
-      - name: Print build outputs
-        run: |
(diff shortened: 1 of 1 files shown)


radcliffkey (approved):
approved but I don't really know much about github action configs

.github/workflows/ci.yml
```diff
@@ -58,15 +59,45 @@ jobs:
...
           retention-days: 7
+
+  deploy_to_pypi:
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    if: |
+      startsWith(github.ref, 'refs/tags/') &&
+      needs.build.outputs.is_semantic_tag == 'true' &&
+      needs.build.outputs.tag == format('v{0}', needs.build.outputs.version)
```
odinuv, line 91: k čemu je tahle podmínka?

.github/workflows/ci.yml
```diff
@@ -58,15 +59,45 @@ jobs:
...
+    steps:
+      - name: Download wheel package
+        uses: actions/download-artifact@v4
+        with:
+          name: keboola_mcp_server-${{ needs.build.outputs.version }}-py3-none-any.whl
+          path: dist/
+
+      - name: Publish package
+        uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
odinuv, line 100: tady je schválně hash?

odinuv:
@vita-stejskal you should be able to to test end-to-end by tagging from the branch, there is nothing preventing it to run in such a way (which is OK)

.github/workflows/ci.yml
vita-stejskal: I took this from [sapi-python-client](https://github.com/keboola/sapi-python-client/blob/master/.github/workflows/push.yml#L132). The hash is a bit unusual.

vita-stejskal:
> @vita-stejskal you should be able to to test end-to-end by tagging from the branch, there is nothing preventing it to run in such a way (which is OK)

That is true, but I am afraid of successfully pushing a version of Keboola MCP Server to pypi.org which was not tested.

PR #44, opened 2025-04-23
Kab 964 component documentation tool

ncapek:
This PR makes use of the new Kaibot endpoint which fetches component documentation.
This endpoint is now used instead of the one that fetches data from the storage API.
Tests were modified to reflect these changes.
An AIServiceclient was added.


src/keboola_mcp_server/client.py
```diff
@@ -53,9 +51,12 @@ def __init__(
         queue_api_url = (
             f"{self._PREFIX_QUEUE_API_URL}{storage_api_url.split(self._PREFIX_STORAGE_API_URL)[1]}"
         )
+        ai_service_api_url = f"{self._PREFIX_AISERVICE_API_URL}{storage_api_url.split(self._PREFIX_STORAGE_API_URL)[-1]}"
 
         self.base_storage_api_url = storage_api_url
         self.base_queue_api_url = queue_api_url
+        # self.base_ai_service_api_url = ai_service_api_url
+        self.base_ai_service_api_url = "http://localhost:5000/v1"
```
vita-stejskal, line 59: This line needs to be removed and the line above uncommented. I'd say.

vita-stejskal: It would be even better to remove `self.base_ai_service_api_url` and `self.base_queue_api_url` altogether, because they are not use anywhere else, but in this function. So, it's enough to have the local variables `queue_api_url` and `ai_service_api_url`. And they should be moved closer to the places where they are used.

src/keboola_mcp_server/client.py
```diff
@@ -297,3 +299,33 @@ def _search(self, params: Dict[str, Any], **kwargs) -> Dict[str, Any]:
...
+    Attributes:
+        base_url (str): The base URL for this endpoint.
+        token (str): A key for the Storage API.
+    """
+
+    def __init__(self, root_url: str, token: str):
+        """
+        Create an AIService endpoint.
+        :param root_url: Root url of API. e.g. "https://ai.keboola.com/"
+        :param token: A key for the Storage API. Can be found in the storage console.
```
vita-stejskal, line 317:
```suggestion
        :param token: A Keboola Storage API token
```

src/keboola_mcp_server/client.py
```diff
@@ -297,3 +299,33 @@ def _search(self, params: Dict[str, Any], **kwargs) -> Dict[str, Any]:
...
         return self._get(url, params=params, **kwargs)
+
+
+class AIServiceClient(Endpoint):
+    """
+    Class handling endpoints for interacting with the Keboola AI Service.
+
+    Attributes:
+        base_url (str): The base URL for this endpoint.
+        token (str): A key for the Storage API.
```
vita-stejskal, line 310: These attributes are made available by `Endpoint` class and just inherited here. They are documented in the `Endpoint` class. Why do you duplicate that documentation here?

src/keboola_mcp_server/component_tools.py
```diff
@@ -153,44 +145,39 @@ class ComponentWithConfigurations(BaseModel):
...
         serialization_alias="configurationSchema",
+    )
+    configuration_row_schema: Optional[dict[str, Any]] = Field(
         default=None,
+        validation_alias=AliasChoices(
+            "configurationRowSchema", "configuration_row_schema", "configuration-row-schema"
+        ),
+        serialization_alias="configurationRowSchema",
     )
     configuration_description: Optional[str] = Field(
```
vita-stejskal, line 175: I believe that this is part of the `documentation` field value. I'd remove this field.

src/keboola_mcp_server/component_tools.py
```diff
@@ -153,44 +145,39 @@ class ComponentWithConfigurations(BaseModel):
 
 
 class Component(ReducedComponent):
```
vita-stejskal, line 147: I think that the newly added fields deserve descriptions (i.e. `description="..."`) that would explain a bit more their meaning to the LLM.

src/keboola_mcp_server/component_tools.py
```diff
@@ -352,8 +339,8 @@ async def _get_component_details(
     :return: The component details
     """
 
-    endpoint = f"branch/{client.storage_client._branch_id}/components/{component_id}"
-    raw_component = await client.get(endpoint)
+    raw_component = client.ai_service_client.get_component_detail(component_id)
+    print(raw_component)
```
vita-stejskal, line 343:
```suggestion
```

src/keboola_mcp_server/component_tools.py
```diff
@@ -611,6 +598,7 @@ async def get_component_configuration_details(
     # Get Configuration Metadata if exists
     endpoint = f"branch/{client.storage_client._branch_id}/components/{component_id}/configs/{configuration_id}/metadata"
     r_metadata = await client.get(endpoint)
+    print(f"metadata: {r_metadata}")
```
vita-stejskal, line 601:
```suggestion
```

src/keboola_mcp_server/component_tools.py
```diff
@@ -619,7 +607,14 @@ async def get_component_configuration_details(
...
         )
-
+    print(
+        {
+            **raw_configuration,
+            "component": component,
+            "component_id": component_id,
+            "metadata": r_metadata,
+        }
+    )
```
vita-stejskal, line 617: Please remove this `print` statement.

tests/test_component_tools.py
```diff
@@ -30,22 +30,34 @@ def mock_components() -> list[dict[str, Any]]:
...
             "name": "AWS S3 Extractor",
             "type": "extractor",
-            "description": "Extract data from AWS S3",
-            "version": "1",
+            "component_categories": [],
+            "documentation_url": "https://components.keboola.com/components/keboola.ex-aws-s3",
+            "documentation": None,
+            "configuration_schema": {},
+            "configuration_row_schema": {},
+            "configuration_description": None,
```
vita-stejskal, line 38: The `configuration_description` field should be removed, I think. I mentioned that in one of my other comment. But all the other newly added fields should have non-empty values. At least in one of the mocked components. Otherwise they are not really tested much.

vita-stejskal (changes requested):
Please, don't forget to update the project version.

src/keboola_mcp_server/component_tools.py
```diff
@@ -352,8 +339,8 @@ async def _get_component_details(
     :return: The component details
     """
 
-    endpoint = f"branch/{client.storage_client._branch_id}/components/{component_id}"
-    raw_component = await client.get(endpoint)
```
vita-stejskal, line 356: You should keep this (the call to SAPI) and use it as a fallback when `component_id` is not found in the components catalog (the AI service call). This can happen for private components which are not part of the catalog, but client may use them in their projects.

@mariankrotil mentioned this in one of the Slack conversations, IIRC.

src/keboola_mcp_server/client.py
ncapek: It's directly copied from Marian's section.

ncapek: I'll remove it.

src/keboola_mcp_server/client.py
ncapek: Done

src/keboola_mcp_server/client.py
```diff
@@ -297,3 +299,33 @@ def _search(self, params: Dict[str, Any], **kwargs) -> Dict[str, Any]:
...
+        Create an AIService endpoint.
+        :param root_url: Root url of API. e.g. "https://ai.keboola.com/"
+        :param token: A key for the Storage API. Can be found in the storage console.
+        """
+        super().__init__(root_url, "", token)
+
+        # set the base url to the root url
+        self.base_url = self.root_url.rstrip("/")
+
+    def get_component_detail(self, component_id: str) -> Dict[str, Any]:
```
radcliffkey, line 324: It would be better to return structured output instead of dict. That's one of the main points of using a custom client over just making 'raw' http requests.

src/keboola_mcp_server/client.py
```diff
@@ -66,9 +67,10 @@ def __init__(
         # The storage_client.jobs endpoint is for storage jobs
         # Use self.jobs_queue instead which provides access to the Job Queue API
         # that handles component/transformation jobs
+        # The AI Service API is used to various endpoints built in KaiBot - such as documentation fetching
         self.storage_client = Client(self.base_storage_api_url, self.token)
-
         self.jobs_queue = JobsQueue(self.base_queue_api_url, self.token)
+        self.ai_service_client = AIServiceClient(self.base_ai_service_api_url, self.token)
```
radcliffkey, line 73: Maybe these individual clients should be added as (read-only) properties for better visibility

src/keboola_mcp_server/component_tools.py
ncapek: Done

src/keboola_mcp_server/component_tools.py
ncapek: Done

src/keboola_mcp_server/component_tools.py
ncapek: Done

src/keboola_mcp_server/client.py
ncapek: We discuseed this with @vita-stejskal and decided that since we are creating the DTOs at the tool level, we don't have to do that here.

src/keboola_mcp_server/client.py
radcliffkey: I can still see the old state

src/keboola_mcp_server/client.py
ncapek: I'm not sure what you mean by read only. Also, Should the AI Service Client really be called this? It seems kind of weird, since it's inheriting from the Endpoint class. At the same time, i don't have a better name for it.

src/keboola_mcp_server/client.py
ncapek: Pushed

vita-stejskal (approved):

src/keboola_mcp_server/component_tools.py
ncapek: I added this, but since there is no private component in my workspace, this is going to be hard to test out.

radcliffkey (approved):

src/keboola_mcp_server/component_tools.py
ncapek: Fallback now tested and working:
image

PR #46, opened 2025-04-23
Kab 991 refactor remove annotated

mariankrotil:
Removed `Annotated` from utility function parameters. Tool function parameters, visible to the bot, still use `Annotated` as required.

### DIFF
--- a/src/keboola_mcp_server/client.py
+++ b/src/keboola_mcp_server/client.py
@@ -147,7 +147,6 @@ class KeboolaClient:
 
         Args:
             endpoint: API endpoint to call
-            data: Request payload
 
         Returns:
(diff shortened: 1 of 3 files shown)


src/keboola_mcp_server/client.py
```diff
@@ -4,7 +4,7 @@
 import os
 import tempfile
 
-from typing import Annotated, Any, Dict, Mapping, Optional, cast, List
+from typing import Any, Dict, Mapping, Optional, cast, List
```
radcliffkey, line 7: While you're at it, don't use Dict/List from typing, use just regular `dict` or `list`

radcliffkey (approved):

PR #50, opened 2025-04-23
Kab 991 refactor quotes for strings

mariankrotil:
This PR standardizes strings, single quotes for string literals and triple double quotes for function docstrings.
Additionally, we configure `black` setting in `pyproject.toml` file to skip quotes formatting.


pyproject.toml
```diff
@@ -37,6 +37,7 @@ keboola-mcp = "keboola_mcp_server.cli:main"
 [tool.black]
 line-length = 100
 target-version = ["py310"]
+skip_string_normalization = true
```
radcliffkey, line 40: If we want to have a rule for strings, can we set black to enforce it?

src/keboola_mcp_server/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Keboola MCP server package."""
 
-__version__ = "0.1.0"
+__version__ = '0.1.0'
```
radcliffkey, line 3: Do we need this? The version is out of sync

mariankrotil: Removed.

pyproject.toml
mariankrotil: Black formatter generally unifies strings by using double quotes. I could not find a rule for single quotes in black formatter setting. Instead I found a possibility to skip string normalization which ignores the quotes of string literals. We can configure `flake8` to check the qoutes.

mariankrotil: Which I do in the following PR.

PR #52, opened 2025-04-23
Kab 991 refactor typing

mariankrotil:
This PR refactors code to use simple types `list` or `dict` instead of using `typing` lib.


tests/test_storage_tools.py
```diff
@@ -9,16 +9,16 @@
 from keboola_mcp_server.sql_tools import TableFqn, WorkspaceManager
 from keboola_mcp_server.storage_tools import (
     BucketDetail,
+    TableColumnInfo,
```
radcliffkey, line 12: It seems I have different isort settings because I get case insensitive sorting :/

radcliffkey (approved):

PR #57, opened 2025-04-25
Kab 991 refactor flake

mariankrotil:
This PR modifies `pyproject.toml` and `ci.yml` pipeline.

- we include `tox` environment within our optional `dev` deps for running tests and code style checks.
- `tox` runs tests using `pytest` and creates coverage and test results files as it was, then it runs code style check that uses flake8 with appropriate settings to conform isort and black and vice versa.
- we add three dependency groups - `dev` (containing build and tox envs), `codestyle` (containing flake8, isort and black dependencies), and `tests` (including pytest packages)
- `ci.yml` uses `tox` instead of separate steps


.github/workflows/ci.yml
```diff
@@ -32,9 +32,12 @@ jobs:
...
           pip install -e ".[dev]"
 
-      - name: Run tests with coverage
+      # Runs tests with coverage and generates coverage.xml file and test-results.xml file
+      # and checks isort & black & flake8 code-style formatting
+      # see setup in pyproject.toml
+      - name: Run tox
         run: |
-          pytest --junitxml ./test-results.xml
+          tox
```
odinuv, line 40: nedavali jsme ten tox nedavno pryc?

mariankrotil: I am not sure. When we first started working on this project, it was not there and tests and style checks had to be run manually one by one. Were there any problems with using tox that contributed to its removal from the CI? I can also run it in separate steps in the CI pipeline if needed.

odinuv: I'm probably mistaking that with something else, Vita or Rado would know better

pyproject.toml
```diff
@@ -69,3 +79,47 @@ show_error_codes = true
 ignore_missing_imports = true
 namespace_packages = true
 explicit_package_bases = true
+
+[tool.flake8]
+max-line-length = 120
+import-order-style = "edited"
+application-package-names = ["keboola_mcp_server"]
+min_python_version = "3.10.0"
+# Skip checking I101 (import order), we want to sort imports in case sensitive manner using simply `isort`
```
radcliffkey, line 88: we can use `--fass` argument for isort if want to enforce flake-like sorting

pyproject.toml
```diff
@@ -69,3 +79,47 @@ show_error_codes = true
...
+extend-ignore = ["I101"]
+# Skip checking F401 in __init__.py files (unused imports), we simplify imports in __init__.py and it is expected
+# not to use them there. 
+per-file-ignores = [
+    "__init__.py: F401",
+]
+
+[tool.tox]
+requires = ["tox>=4.23", "black>=25.1", "isort>=6.0"]
```
radcliffkey, line 97: I would just keep tox here, each tox env can specify its own deps

pyproject.toml
```diff
@@ -69,3 +79,47 @@ show_error_codes = true
...
+
+[tool.tox.env_run_base]
+description = "Run tests"
+extras = ["dev"]
+package = "wheel"
+wheel_build_env = ".pkg"
+commands = [["pytest", "--junitxml", "./test-results.xml"]]
+
+[tool.tox.env.code_style]
+description = "Run code style check - black, isort, and flake"
```
radcliffkey, line 108: add `package = "skip"`

pyproject.toml
```diff
@@ -69,3 +79,47 @@ show_error_codes = true
...
+deps = [
+    "black",
```
radcliffkey, line 110: can't we just use `extras = ["dev"]` since all these deps are already there?

pyproject.toml
```diff
@@ -69,3 +79,47 @@ show_error_codes = true
...
+    "flake8-bugbear",
+    "flake8-colors",
+    "flake8-import-order",
+    "flake8-typing-imports",
+    "Flake8-pyproject",
+    "flake8-quotes",
+    "pep8-naming",
+]
+commands = [
+    ["black", "src/", "tests/", "--check"],
```
radcliffkey, line 122: I'm not sure we need to run all 3 tools here. Can't we just use `flake8` for checking and have `black` and `isort` configured to produce code that will pass? So the devs would run them before commiting or fix the errors manually if the prefer to. Also, relying on flake for checking everything has an advantage for Cursor users: if you use flake8 in Cursor, it logs into the 'Problems' tab. When Cursor generates code, it can see these errors tries to fix them.

radcliffkey:
Not directly related, but we have some mypy settings in the project - is anyone using it?

.github/workflows/ci.yml
radcliffkey: I don't remember removing tox. It is true that tox is more useful for local development than for CI builds, because they always run in a new clean environment anyway (correct me if that's not true). But it still has the advantage that testing, linting etc. is configured in one place and will run with the same config locally and in the CI.

pyproject.toml
mariankrotil: Do we want the flake-like sorting? As it sorts alphabetically without considering lower/upper case as follows:
`from foo import A, a, B, b, C` 
But more readable seems to be:
`from foo import A, B, C, a, b`

mariankrotil:
> Not directly related, but we have some mypy settings in the project - is anyone using it?

Not sure. But it seems none of us

pyproject.toml
mariankrotil: Removed here: 2550c96661550b204e2d06c2a323591c62268383

pyproject.toml
mariankrotil: Added here: 2550c96661550b204e2d06c2a323591c62268383

pyproject.toml
mariankrotil: I think we would install all dependencies in the `code_style` tox environment, but we only need the `flake8` since it is run on the `src/` and `tests/` directories explicitly.

pyproject.toml
mariankrotil: Ok, removed `isort` and `black` checking here: 2550c96661550b204e2d06c2a323591c62268383

mariankrotil:
> > Not directly related, but we have some mypy settings in the project - is anyone using it?
> 
> Not sure. But it seems none of us

It has been removed from the dependencies and pyproject

pyproject.toml
```diff
@@ -20,53 +20,90 @@ dependencies = [
...
+env_list = ["python", "env_flake8"]
+
+[tool.tox.env_run_base]
+description = "Run tests"
+extras = ["tests"]
+package = "wheel"
+wheel_build_env = ".pkg"
+commands = [["pytest", "--junitxml", "./test-results.xml"]]
+
+[tool.tox.env.env_flake8]
```
radcliffkey, line 96:
```suggestion
[tool.tox.env.flake8]
```

radcliffkey (approved):

pyproject.toml
```diff
@@ -20,53 +20,90 @@ dependencies = [
...
+    "__init__.py: F401",
+]
+# For Compatibility with black we skip: E203 whitespace before ':', W503 checks line break before binary operator
+# For Compatibility with isort we skip: I101 Imported names are in the wrong order since we follow isort format using
+# alphabetical case-sensitive sort A, B, a, b 
+extend-ignore = ["E203", "W503", "I101"] 
+
+[tool.tox]
+requires = ["tox>=4.23"]
+env_list = ["python", "env_flake8"]
```
radcliffkey, line 87:
```suggestion
env_list = ["python", "flake8"]
```

PR #59, opened 2025-04-25
KAB-1002 refactor API clients

radcliffkey:
- Make JobsQueue and AI service clients async
- Refactor the basic async client for storage API (so far it only has raw get/post... methods)
- Use pytest mocker fixture in tests

### DIFF
--- a/src/keboola_mcp_server/client.py
+++ b/src/keboola_mcp_server/client.py
@@ -1,12 +1,9 @@
 """Keboola Storage API client wrapper."""
 
 import logging
-import os
-import tempfile
 from typing import Any, Mapping, Optional, cast
 
 import httpx
-from kbcstorage.base import Endpoint
 from kbcstorage.client import Client
 from pydantic import BaseModel, Field
 
@@ -34,7 +31,8 @@ class KeboolaClient:
         storage_api_token: str,
         storage_api_url: str = 'https://connection.keboola.com',
     ) -> None:
-        """Initialize the client.
+        """
+        Initialize the client.
 
         Args:
             storage_api_token: Keboola Storage API token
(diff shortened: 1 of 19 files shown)


src/keboola_mcp_server/client.py
```diff
@@ -95,127 +107,205 @@ async def get(
...
+        :return: A new instance of JobsQueueClient
         """
-        super().__init__(root_url, '', token)
-
-        # set the base url to the root url
-        self.base_url = self.root_url.rstrip('/')
+        return cls(raw_client=RawKeboolaClient(base_api_url=root_url, api_token=token))
 
-    def detail(self, job_id: str) -> dict[str, Any]:
+    async def detail(self, job_id: str) -> dict[str, Any]:
```
mariankrotil, line 298: Having in mind that it is not part of this PR, but, can you please rename this `detail` method to `detail_job` to align with other methods' naming for the api calls. In addition, a new `create` class method is introduced which is intended to create the instance of the Client class, and it could be confused with the other api call methods, hence, adding a `_job` would be more informative.

src/keboola_mcp_server/client.py
```diff
@@ -54,39 +51,54 @@ def __init__(
         )
         ai_service_api_url = f"{self._PREFIX_AISERVICE_API_URL}{storage_api_url.split(self._PREFIX_STORAGE_API_URL)[1]}"
 
-        self.base_storage_api_url = storage_api_url
+        # Initialize clients for individual services
+        self.storage_client_sync = SyncStorageClient(storage_api_url, self.token)
```
mariankrotil, line 55: An idea: What if we designed a wrapper function which would asynchronously encapsulate the `SyncStorageClient's` methods by running them in a new async thread e.g.:
```python
import asyncio
...
async def async_run_sync(sync_fn: Callable, *args, **kwargs):
    return await asyncio.to_thread(sync_fn, *args, **kwargs)
```
Usage:
```python
... = await keboola_client.storage_client.async_run_sync(
                   keboola_client.storage_client_sync.configurations.list, "keboola-component-id")
```

mariankrotil (approved):

src/keboola_mcp_server/client.py
radcliffkey: that's one of the possible solutions, we should think about it. I didn't want to deal with it in this PR since it's already quite big.

src/keboola_mcp_server/client.py
radcliffkey: Done in [a37c07f](https://github.com/keboola/mcp-server/pull/59/commits/a37c07f9211f1febe30a2b750a9b2cf0e529a4b8)

PR #61, opened 2025-04-29
component tools

odinuv:


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -327,4 +333,175 @@ async def create_sql_transformation(
...
+    Retrieves sample configuration examples for a specific component from a JSONL file.
+    USAGE:
+        - Use when you want to see example configurations for a specific component.
+    EXAMPLES:
+        - user_input: `Show me example configurations for component X`
+            -> set the component_id parameter accordingly
+            -> returns a markdown formatted string with configuration examples
+    """
+    import json
+    from pathlib import Path
```
mariankrotil, line 467: We should move these imports to the top of the file.
Assuming it is a partial solution for getting the examples, we should think about how to get the examples dynamically e.g. from a remote server instead of local file.

davidesner:
@mariankrotil I think we will not be merging this one. To me it serves mainly as an example and proof of concept. Next week we should discuss how to collect the examples and present it to the tooling. I think it can be part of the indexing process / or a separate pipeline that collects them.

odinuv:
Yes, this was just a draft, we can either revisit it and finish it or dump it

PR #65, opened 2025-05-02
slight update in readme.md & improved description for SQL transformations generation

ottomansky:
1. placeholder variables in readme.md changed to underscore instead of "-" sign for easier copy pasting
2. Improved description generation for SQL transformations in tools.py

### DIFF
--- a/src/keboola_mcp_server/tools/components/tools.py
+++ b/src/keboola_mcp_server/tools/components/tools.py
@@ -227,8 +227,27 @@ async def create_sql_transformation(
         str,
         Field(
             description=(
-                'The detailed description of the SQL transformation capturing the user intent, explaining the '
-                'SQL query, and the expected output.'
+                'The structured documentation for the SQL transformation. Suggested format:\n\n'
+                '# SQL Transformation Documentation\n\n'
+                '## Overview\n'
+                'Brief description of what this transformation does and its purpose in the overall data flow.\n\n'
+                '### Usage Example\n'
+                '1. Ensure required input mapping is configured (e.g., `in.c.source.table_name`).\n'
+                '2. Add the transformation to a flow in the desired order.\n'
+                '3. Output will be available in `out.c.target.table_name`.\n\n'
+                '## User Parameters Description\n'
+                '| Parameter Name | Description | Required | Default |\n'
+                '|----------------|-------------|----------|--------|\n'
+                '| `source_table` | Input table with raw data | ✅ | N/A |\n'
+                '| `target_table` | Destination table after transformation | ✅ | N/A |\n\n'
+                '> 🧠 **Tip:** Document any variables or shared code dependencies here.\n\n'
+                '## Code Logic\n\n'
+                '### Dependencies\n'
+                'List any used DB-specific functions or external shared code blocks.\n\n'
+                '- DB functions used (e.g., `NOW()`, `REGEXP_REPLACE()`)\n'
+                '- Shared code blocks referenced\n\n'
+                '### Component Workflow\n'
+                'Describe the transformation logic flow from input to output.'
             ),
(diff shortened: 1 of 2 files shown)


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -227,8 +227,27 @@ async def create_sql_transformation(
...
+                '| `source_table` | Input table with raw data | ✅ | N/A |\n'
+                '| `target_table` | Destination table after transformation | ✅ | N/A |\n\n'
+                '> 🧠 **Tip:** Document any variables or shared code dependencies here.\n\n'
+                '## Code Logic\n\n'
+                '### Dependencies\n'
+                'List any used DB-specific functions or external shared code blocks.\n\n'
+                '- DB functions used (e.g., `NOW()`, `REGEXP_REPLACE()`)\n'
+                '- Shared code blocks referenced\n\n'
+                '### Component Workflow\n'
+                'Describe the transformation logic flow from input to output.'
```
ncapek, line 250: Can you explain in more detail precisely what this detailed description is trying to accomplish? Did you have issues with the previous one? Does this drastically improve the performance? If so, we should think more carefully where we want to store this long string - I suspect it should be somehow integrated into the function docstring rather than this description.

Additionally, if you have noticed any changes in agent performance, then it's time to start designing and implementing some kind of evaluation framework.

mariankrotil: See the screenshot how the description looks in cursor. We should format it:
image

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -227,8 +227,27 @@ async def create_sql_transformation(
...
         Field(
             description=(
-                'The detailed description of the SQL transformation capturing the user intent, explaining the '
-                'SQL query, and the expected output.'
+                'The structured documentation for the SQL transformation. Suggested format:\n\n'
+                '# SQL Transformation Documentation\n\n'
+                '## Overview\n'
+                'Brief description of what this transformation does and its purpose in the overall data flow.\n\n'
+                '### Usage Example\n'
+                '1. Ensure required input mapping is configured (e.g., `in.c.source.table_name`).\n'
```
mariankrotil, line 235: Does the format really follow this `in.c.source.table_name`? Should not it be `in.c-source.table_name`?

ottomansky: Yes it should, I missed this one, sorry

src/keboola_mcp_server/tools/components/tools.py
ottomansky: > Can you explain in more detail precisely what this detailed description is trying to accomplish? Did you have issues with the previous one? Does this drastically improve the performance? If so, we should think more carefully where we want to store this long string - I suspect it should be somehow integrated into the function docstring rather than this description.
> 
> Additionally, if you have noticed any changes in agent performance, then it's time to start designing and implementing some kind of evaluation framework.
----------------------------

The original Idea was to improve the description generation using the whole strengths of markdown support in the description field, right now the generation is pretty generic in the original tool. 
I am not a developer by any means, these things are really new for me.

ottomansky:
closed this pull request, lets work on this later

src/keboola_mcp_server/tools/components/tools.py
mariankrotil: I agree, indeed, it is a good idea to utilize the llm generation power to write an extensive description of a new sql transformation, if you found those description generic and since people usually do not want to write it. We should finish this PR i like the idea but i would suggest reformat the structure of the markdown description at least little not to shaddow the instructions and overall tool descriptions.

PR #66, opened 2025-05-02
Kab 1006 update column description

ncapek:
This PR add an update_column_description tool for the MCP server.

### DIFF
--- a/src/keboola_mcp_server/tools/storage.py
+++ b/src/keboola_mcp_server/tools/storage.py
@@ -134,9 +134,9 @@ class TableDetail(BaseModel):
 
 
 class UpdateDescriptionResponse(BaseModel):
-    description: str = Field(..., description="The updated description value.", alias='value')
-    timestamp: datetime = Field(..., description="The timestamp of the description update.")
-    success: bool = Field(default=True, description="Indicates if the update succeeded.")
+    description: str = Field(..., description='The updated description value.', alias='value')
+    timestamp: datetime = Field(..., description='The timestamp of the description update.')
+    success: bool = Field(default=True, description='Indicates if the update succeeded.')
 
 
 @tool_errors()
@@ -220,7 +220,7 @@ async def update_bucket_description(
         'metadata': [{'key': MetadataField.DESCRIPTION.value, 'value': description}],
     }
     response = await client.storage_client.post(endpoint=metadata_endpoint, data=data)
-    description_entry = next(entry for entry in response if entry.get("key") == MetadataField.DESCRIPTION.value)
+    description_entry = next(entry for entry in response if entry.get('key') == MetadataField.DESCRIPTION.value)
 
     return UpdateDescriptionResponse.model_validate(description_entry)
 
@@ -244,7 +244,7 @@ async def update_table_description(
     }
     response = await client.storage_client.post(endpoint=metadata_endpoint, data=data)
     description_entry = next(
-        entry for entry in response.get('metadata') if entry.get("key") == MetadataField.DESCRIPTION.value
+        entry for entry in response.get('metadata') if entry.get('key') == MetadataField.DESCRIPTION.value
     )
 
     return UpdateDescriptionResponse.model_validate(description_entry)
@@ -281,7 +281,7 @@ async def update_column_description(
     description_entry = next(
         entry
         for entry in response.get('columnsMetadata').get(column_name)
-        if entry.get("key") == MetadataField.DESCRIPTION.value
+        if entry.get('key') == MetadataField.DESCRIPTION.value
     )
 
     return UpdateDescriptionResponse.model_validate(description_entry)
--- a/src/keboola_mcp_server/tools/storage.py
+++ b/src/keboola_mcp_server/tools/storage.py
@@ -1,6 +1,7 @@
 """Storage-related tools for the MCP server (buckets, tables, etc.)."""
 
 import logging
+from datetime import datetime
 from typing import Annotated, Any, Mapping, Optional, cast
 
 from mcp.server.fastmcp import Context
@@ -140,7 +141,7 @@ class TableDetail(BaseModel):
 
 class UpdateDescriptionResponse(BaseModel):
     description: str = Field(..., description="The updated description value.", alias='value')
-    timestamp: str = Field(..., description="When the description was updated.")
+    timestamp: datetime = Field(..., description="The timestamp of the description update.")
     success: bool = Field(default=True, description="Indicates if the update succeeded.")
 
 
--- a/src/keboola_mcp_server/tools/storage.py
+++ b/src/keboola_mcp_server/tools/storage.py
@@ -73,7 +73,7 @@ class BucketDetail(BaseModel):
 
     @model_validator(mode='before')
     @classmethod
-    def set_table_count(cls, values):
+    def set_table_count(cls, values: dict[str, Any]) -> dict[str, Any]:
         if isinstance(values.get('tables'), list):
             values['tables_count'] = len(values['tables'])
         else:
@@ -82,7 +82,7 @@ class BucketDetail(BaseModel):
 
     @model_validator(mode='before')
     @classmethod
-    def set_description(cls, values):
+    def set_description(cls, values: dict[str, Any]) -> dict[str, Any]:
         values['description'] = extract_description(values)
         return values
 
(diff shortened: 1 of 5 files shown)


src/keboola_mcp_server/tools/storage.py
```diff
@@ -274,3 +294,34 @@ async def update_table_description(
...
+    Field(description='The response object of the column description update.'),
+]:
+    """Update the description for a given column in a Keboola table."""
+    client = KeboolaClient.from_state(ctx.session.state)
+    metadata_endpoint = f'tables/{table_id}/metadata'
+
+    data = {
+        'provider': 'user',
+        'columnsMetadata': {
+            f'{column_name}/0': [  # TODO: Understand what this index is doing
```
ncapek, line 315: @odinuv I have 2 questions here:
1. How does the indexing work in columnsMetadata?
2. am I using the correct endpoint? I noticed that there is also a dedicated columnMetadata endpoint

src/keboola_mcp_server/tools/storage.py
```diff
@@ -134,45 +138,61 @@ def set_description(cls, values):
...
+
+    @classmethod
+    def _extract_data(cls, raw_input) -> dict:
+        raise NotImplementedError("Subclasses must implement _extract_data")
 
     @model_validator(mode='before')
-    def extract_from_response(cls, values):
-        if isinstance(values, list) and values:
-            data = values[0]  # the response returns a list - elements for each update
+    def extract(cls, values):
```
mariankrotil, line 151: please add return type

src/keboola_mcp_server/tools/storage.py
```diff
@@ -134,45 +138,61 @@ def set_description(cls, values):
...
 
 
-class UpdateTableDescriptionResponse(BaseModel):
-    success: bool = True
+class UpdateBucketDescriptionResponse(BaseUpdateDescriptionResponse):
+    description: str = Field(description='The updated bucket description value')
+    timestamp: str = Field(description='When the description was updated')
+
+    @classmethod
+    def _extract_data(cls, raw_input):
```
mariankrotil, line 168: add return type

mariankrotil: Please the same for all `@classmethod`s

src/keboola_mcp_server/tools/storage.py
```diff
@@ -134,45 +138,61 @@ def set_description(cls, values):
         return values
 
 
-class UpdateBucketDescriptionResponse(BaseModel):
+class BaseUpdateDescriptionResponse(BaseModel):
     success: bool = True
-    description: str = Field(description='The updated description value')
-    timestamp: str = Field(description='When the description was updated')
+    description: str
+    timestamp: str
```
mariankrotil, line 144: We can use `datetime.datetime` as it is done in jobs. Perhaps more structured to the agent.

mariankrotil: Perhaps, we can use `Field` and its `description` to describe `timestamp` ones here.

src/keboola_mcp_server/tools/storage.py
odinuv: i think it's just some kinde of typo, it should behave like the example in here - that is https://keboola.docs.apiary.io/#reference/metadata/table-metadata/create-or-update

```
 "columnsMetadata": {
    "id": [
      {
        "id": "19510",
        "key": "some-key",
        "value": "Some value",
        "provider": "test-component",
        "timestamp": "2017-02-13 23:36:10"
      }
    ]
```

you can use either endpoint, if you're setting just a single column, the https://keboola.docs.apiary.io/#reference/metadata/column-metadata/create-or-update has a simpler structure, the table/metadata endpoint is useful for setting multiple columns at once.

src/keboola_mcp_server/tools/storage.py
```diff
@@ -274,3 +294,34 @@ async def update_table_description(
...
+                    'key': MetadataField.DESCRIPTION.value,
+                    'value': description,
+                    'columnName': column_name,
+                }
+            ]
+        },
+    }
+
+    response = await client.post(endpoint=metadata_endpoint, data=data)
+    print(f"Response: {response}")
```
vita-stejskal, line 326:
```suggestion
```

tests/tools/test_storage.py
```diff
@@ -247,6 +249,23 @@ def mock_update_table_description_response() -> Mapping[str, Any]:
...
+            {
+                'table_id': '1724427984',
+                'column_name': 'test',
+                'key': MetadataField.DESCRIPTION.value,
+                'value': 'Updated column description',
+                'provider': 'user',
+                'timestamp': '2025-04-07T16:47:18+0200',
+            }
+        ]
+    }
```
vita-stejskal, line 266: Is this really a response from `POST v2/storage/tables/{table_id}/metadata` endpoint? It does not look like the example here -- https://keboola.docs.apiary.io/#reference/metadata/table-metadata/create-or-update

src/keboola_mcp_server/tools/storage.py
```diff
@@ -134,45 +138,61 @@ def set_description(cls, values):
         return values
 
 
-class UpdateBucketDescriptionResponse(BaseModel):
+class BaseUpdateDescriptionResponse(BaseModel):
```
vita-stejskal, line 141: Umm, inheriting pydantic models is tricky and it's an overkill in this particular case, I'd say.

* You already have specialized functions (the tools) for dealing with the descriptions. These functions prepare data for the SAPI call and it seems natural to let them handle (deserialize) the SAPI response too.
* There can be just one data object returned from all the "update description" tools, e.g.

vita-stejskal (changes requested):

src/keboola_mcp_server/tools/storage.py
```diff
@@ -134,45 +138,61 @@ def set_description(cls, values):
...
+        raise ValueError('Expected "metadata" field to be a non-empty list.')
+
+
+class UpdateColumnDescriptionResponse(BaseUpdateDescriptionResponse):
+    description: str = Field(description='The updated column description value')
+    timestamp: str = Field(description='When the description was updated')
+
+    @classmethod
+    def _extract_data(cls, raw_input):
+        metadata = raw_input.get('metadata', [])
```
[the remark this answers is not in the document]
ncapek: You're right. I was cheating and only took the subset of the response that I actually needed. That wasn't quite right.

src/keboola_mcp_server/tools/storage.py
ncapek: Added

src/keboola_mcp_server/tools/storage.py
ncapek: Done

src/keboola_mcp_server/tools/storage.py
ncapek: Done

ujovlado:
Hi @ncapek, I changed the base to main. Would you mind rebasing this PR on top of `main`? Please let me know if help is needed.

tests/tools/test_storage.py
ncapek: I fixed it.

tests/tools/test_storage.py
```diff
@@ -1,6 +1,7 @@
 from typing import Any, Mapping, Sequence
 
 import pytest
+from dateutil.parser import isoparse
```
vita-stejskal, line 4: Umm, this seems to come from `python-dateutil` library which is not declared as a dependency in `pyproject.toml` file. Do we really need that dependency? Would not `datetime.fromisoformat()` function from Python Standard Library be sufficient? If you do need full ISO format parsing then please add `python-dateutil` as a dependency to `pyproject.toml` file.

vita-stejskal (approved):

tests/tools/test_storage.py
ncapek: I didnt realize it's an external dependency (i didnt need to install anything). Nevertheless I found a way to make it work with `datetime.fromisoformat()`. Fixed.

tests/tools/test_storage.py
```diff
@@ -88,6 +87,74 @@ def mock_buckets() -> Sequence[Mapping[str, Any]]:
...
+                    'provider': 'user',
+                    'timestamp': '2024-01-01T00:00:00Z',
+                }
+            ]
+        },
+    }
+
+
+@pytest.fixture
+def mock_update_column_description_response() -> Mapping[str, Any]:
```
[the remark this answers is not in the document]
ncapek: I realize it's technically the "same code" but since they support different tests, I thought it would be better for them to have dedicated fixtures. I mostly still think this is the case.

PR #67, opened 2025-05-05
Readme branch

pavel242242:
this is how readme should look like.

Sections to doube check before merging:
- #### Option B: Using Docker
- ## 🧰 Supported Tools
- ## 🧪 Development & Contribution


pavel242242:
@bornedan

bornedan (approved):
I think the README is mergeable now. It will be updated again following further changes in MCP server.

ujovlado (changes requested):
Check the comments bellow. I think we should spit this PR to 2-3. I would vote for not adding Contributing rules, unless we are sure.

Also, few things look weird to me - e.g. "Stay in touch" and other sections.

PR #69, opened 2025-05-06
fix: Main->master branch trigger workflow

Matovidlo:
The `main` branch does not exists

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,10 +4,10 @@ on:
   workflow_dispatch:
   pull_request:
     branches:
-      - main
+      - master
   push:
     branches:
-      - main
+      - master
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -7,7 +7,7 @@ on:
       - main
```
radcliffkey, line 7: here too, right?

AdamVyborny (approved):
Je to ofenzivní, ale co už

.github/workflows/release.yml
AdamVyborny: Good point

ujovlado:
@Matovidlo @radcliffkey please change main branch to `main`.

PR #73, opened 2025-05-06
KAB-1016 fix missing await

radcliffkey:


src/keboola_mcp_server/tools/jobs.py
```diff
@@ -155,21 +155,21 @@ def validate_result_field(cls, current_value: Union[list[Any], dict[str, Any], N
 async def retrieve_jobs(
     ctx: Context,
     status: Annotated[
-        JOB_STATUS | None,
+        JOB_STATUS,
```
radcliffkey, line 158: this typing is correct but calling the tool fails if it used liked this

PR #78, opened 2025-05-07
Clarification in MCP Installation options

pavel-synek:
Proposed a little clarification in the first step of the MCP installation to make it more like a no-brainer.... Smithery and PIP are two different options, so it should be clear those are not steps.


jordanrburger:
Thanks, Pavel. We will be redoing the Readme completely later today. Going to close this PR out.

PR #84, opened 2025-05-09
(feat): Added basic component config verification tooling

radektomasek:


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -23,7 +23,9 @@
     _retrieve_components_configurations_by_ids,
     _retrieve_components_configurations_by_types,
 )
+
 from keboola_mcp_server.tools.sql import get_sql_dialect
+from .validation import validate_component_configuration
```
vita-stejskal, line 28: Please, do not use relative imports.

src/keboola_mcp_server/tools/components/validation.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Args:
+        client: The Keboola client
+        component_id: The ID of the component
+
+    Returns:
+        A tuple of (root_configuration_schema, row_configuration_schema)
+    """
+    try:
+        endpoint = 'v2/storage/components'
```
davidesner, line 28: This will only give you components in a project. There is no reason imo to have another call to storage API here. The root and row schema is available in the component detail that uses the index in the AI service

src/keboola_mcp_server/tools/components/validation.py
```diff
@@ -0,0 +1,145 @@
...
+
+async def validate_component_configuration(
+        ctx: Context,
+        component_id: Annotated[
+            str,
+            Field(
+                description='The ID of the component to validate the configuration against.',
+            ),
+        ],
+        configuration: Annotated[
```
davidesner, line 98: we call this `parameters` in all the related tools (update/create). It may not matter but I would incline to keep it consistent.

PR #87, opened 2025-05-13
KAB-1011 Integration tests - project setup

radcliffkey:
- Set up and teardown of the testing Keboola project
 - Flake8 plugin for pytest

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -e ".[dev]"
+          pip install  ".[dev]"
 
       # Runs tests with coverage and generates coverage.xml file and test-results.xml file
       # and checks flake8 code-style formatting that is compatible with isort & black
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -97,9 +81,48 @@ jobs:
           compression-level: 0  # wheels are ZIP archives
           retention-days: 7
 
+  integration_tests:
```
radcliffkey, line 84: @odinuv this part is vibe-coded so pls look at it - the goal is to not run integration tests in parallel for various python versions since they all work with the same project.

vita-stejskal (approved):
Nice!

integtests/conftest.py
```diff
@@ -8,25 +12,120 @@
...
-def keboola_client(env_file_loaded: bool, mocker) -> KeboolaClient:
+@dataclass(frozen=True)
+class TableDef:
+    bucket_id: str
+    table_name: str
+    table_id: str
+
+
+@pytest.fixture(scope='session')
+def test_buckets() -> list[BucketDef]:
```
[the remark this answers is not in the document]
radcliffkey: done in [db0103f](https://github.com/keboola/mcp-server/pull/87/commits/db0103fc5027d1ce211b69ab8a95065bcad7ee7d)

.github/workflows/ci.yml
radcliffkey: Suggestion from @odinuv on Slack:

> in this case (since there is a limited number of those and they are added consciously) i'd say - make multiple projects and add the storage token as part of the test matrix variable

PR #93, opened 2025-05-19
Kab 998 fastmcp 2x - session state injected via tool decorator

mariankrotil:
Improvement to the Kab 998 fastmcp 2x v2

This PR introduces an alternative way of creating and accessing `context.sesison.state` through decorator functions. Before we customized the FastMCP server and overrode its methods to incorporate the `session.state`. However, Inherited code parts need to be kept up to date with the fast evolving fastmcp codebase, whose new (backward uncompatible) updates can cause errors. Hence, we can use decorator which dynamically inserts the `session.state` to the `context` just by calling the tool function.

Changes:
- updated to fastmcp 2.0
- tests updated wrt changes and to fastmcp 2.0
- add the decorator `@with_session_state()` removed inheritance from `mcp.py`
- refactored the `sql tools` , we moved the `Workspace` to the separate file due to circular imports
- annotated all tools `@with_session_state()`
- use config server setting as a default option.

Usage:
We can use `@with_session_state()` annotation which injects the session `state` created by session state factory to the `context.session` only if the annotated function (tool) includes `Context` type parameter in its arguments, otherwise it raises an error.
```python

@with_session_state()
def tool(..., ctx: Context, ...):
    ....
```


tests/conftest.py
```diff
@@ -9,8 +10,12 @@
     KeboolaClient,
     RawKeboolaClient,
 )
-from keboola_mcp_server.mcp import StatefulServerSession
-from keboola_mcp_server.tools.sql import WorkspaceManager
+from keboola_mcp_server.mcp import SessionState
+from keboola_mcp_server.tools.workspace import WorkspaceManager
+
+
+class StatefulServerSession(ServerSession):
```
vita-stejskal, line 17: This is wrong IMO. It's a random contract used just in the tests. IMO on the line 49 you should just mock `ServerSession` directly.

mariankrotil: Sure, removed here: 1f3435986e96921f8ec57796dca4c4c7bfac520d
Using mocking of `ServerSession`.

mariankrotil:
> please check the readme file if the intructions are still correct

I have tried the `uv` commands with my local project and everything seems working.

src/keboola_mcp_server/cli.py
```diff
@@ -56,19 +71,28 @@ def main(args: Optional[list[str]] = None) -> None:
...
+        LOG.info(f'Creating server with config: {config}')
+        keboola_mcp_server = create_server(config)
+        await keboola_mcp_server.run_async(transport=parsed_args.transport)
     except Exception as e:
         LOG.exception(f'Server failed: {e}')
         sys.exit(1)
 
 
+def main(args: Optional[list[str]] = None) -> None:
+    import asyncio
```
radcliffkey, line 93: you can put this to the top with the 'global' imports. Also, please change relative imports (`from .config import Config`) to absolute.

mariankrotil: Done here: 6863ab34124db61ad566d647385196ce20f8ed2e

pyproject.toml
```diff
@@ -4,14 +4,14 @@ build-backend = "hatchling.build"
...
-version = "0.18.0"
+version = "0.19.0"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
 license = "MIT"
 authors = [{ name = "Keboola", email = "devel@keboola.com" }]
 dependencies = [
-    "mcp[cli] == 1.6.0",
+    "fastmcp ~= 2.2",
```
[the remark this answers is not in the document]
mariankrotil: Updated here: 1544969caa7047fc5cb9d46fba36cdb71d55ea7b

src/keboola_mcp_server/mcp.py
```diff
@@ -1,183 +1,305 @@
...
         super().add_tool(
             fn=fn,
             name=name,
             description=description or textwrap.dedent(fn.__doc__ or '').strip(),
+            tags=tags,
+            annotations=annotations,
         )
+
+
+def keboola_session_params_factory(ctx: Optional[Context] = None) -> SessionParams:
```
vita-stejskal, line 81: Why are `keboola_session_params_factory()` function and `keboola_session_state_factory()` function public? They are used only within this file. They could be private and have their names start with undescore.

src/keboola_mcp_server/mcp.py
```diff
@@ -1,183 +1,305 @@
...
+        LOG.error(f'Failed to initialize Storage API Workspace manager: {e}')
+        raise
+
+    return state
+
+
+def with_session_state(
+    session_state_field_name: str = 'state',
+    session_state_factory: SessionStateFactory = keboola_session_state_factory,
+    session_params_factory: SessionParamsFactory = keboola_session_params_factory,
```
vita-stejskal, line 139: This is totally confusing. This decorator uses `keboola_session_state_factory()` and `keboola_session_params_factory()` functions as the defaults for the `session_state_factory` and `session_params_factory`, but `with_session_state_from()` decorator uses `default_session_state_factory()` and `default_session_params_factory()` as the defaults. And the factories behave differently. There should only be one default implementation for each factory type.

src/keboola_mcp_server/mcp.py
```diff
@@ -1,183 +1,305 @@
...
+            'Did you forget to call it? Use @session_state_from()'
+        )
+
+    session_state_factory = session_state_factory or default_session_state_factory()
+    session_params_factory = session_params_factory or default_session_params_factory()
+    injector = custom_session_state_injector or default_session_state_injector()
+    return injector(session_state_field_name, session_state_factory, session_params_factory)
+
+
+def default_session_state_factory() -> SessionStateFactory:
```
vita-stejskal, line 198: This function could also be private, I think.

src/keboola_mcp_server/mcp.py
```diff
@@ -1,183 +1,305 @@
...
+    def _(params: SessionParams | None = None) -> SessionState:
+        return params or {}
+
+    return _
+
+
+def default_session_params_factory() -> SessionParamsFactory:
```
vita-stejskal, line 205: This function could also be private, I think.

src/keboola_mcp_server/mcp.py
```diff
@@ -1,183 +1,165 @@
...
+                )
+            # convert positional args to kwargs using inspect.signature in case context is passed as positional arg
+            updated_kwargs = inspect.signature(fn).bind(*args, **kwargs).arguments
+            ctx = updated_kwargs.get(ctx_kwarg) if ctx_kwarg else None
+
+            if not isinstance(ctx, Context):
+                raise TypeError(f'The "ctx" argument must be of type Context, got {type(ctx)}.')
+
+            if not getattr(ctx.session, 'state', None):
+                # This is here to allow mocking the context.session.state in tests.
```
radcliffkey, line 149: Is it really just for tests? Don't the tools need to access `ctx.session.state` regardless of the tests?

radcliffkey (approved):
I am approving so I don't block it. But I'm still quite confused about how it works. There are many concept like lifespan, context, session, state, config. It would be nice to add some explanation/diagram of how these concepts are related.

src/keboola_mcp_server/mcp.py
mariankrotil: Other option would be inserting the state every tool call. But once the `session.state` is maintained for the client connection by the server I think it is reasonable to initialize it only once per session. So not sure if it is for mocking only.

mariankrotil:
> I am approving so I don't block it. But I'm still quite confused about how it works. There are many concept like lifespan, context, session, state, config. It would be nice to add some explanation/diagram of how these concepts are related.

`Context` object provides access to MPC capabilities, such as log messages to the client, access resources, asking the client agent, server instance,.... so on

`lifespan` is created when the server starts and is destroyed with server. It is a server state to which you can access from server everywhere, and is intended to access resources like `Redis` database, configuration, and so on. It is provided within the `Context`.

`session` (storing temporal objects) is created for each client-server connection and maintained by the server (TTL), also attached to the `Context`

`state` we inject state to the session which holds our initialized `Clients` with credentials for each user which sends those credentials within request

`config` is our internal  server configuration (From my perspective it serves mainly for local run servers)

mariankrotil:
I like those simplification changes, I approve as well

src/keboola_mcp_server/mcp.py
vita-stejskal: As @mariankrotil says reusing the `ctx.session.state` is there solely for the purpose of mocking the session state in the tests.

PR #97, opened 2025-05-20
Add Data App Creation

Nweaver412:
This still has to go through testing before implemented, but just a proof of concept for creating and generating Data Apps

**Create Data App**
- Added `create_data_app` function that enables creation of new data apps with customizable configurations
- Implemented script generation using the model data app creation
- Connected to `_get_data_context` to get context for the data/what type of data app may make sense
- Added support for custom streamlit configurations and other params
- Only built out for 'None' auth
- TODO add compatibility for added packages

**Data Context Management**
- Implemented `_get_data_context` to gather comprehensive information about available data
- The goal of this is to be able to pass detail to the model to make sure the streamlit generation makes sense for the data in the related buckets

**Models and Configuration**
- Added new data models:
  - `DataApp`: Core model for data app configuration
  - `DataAppConfiguration`: Extended component configuration for data apps

**Notes**
- The `_start_data_app` function is currently a placeholder and will be implemented in a future PR
- Also need to add some `_get_data_app` functionality
- Some TODO items remain for library management in the script generation process

### DIFF
--- a/src/keboola_mcp_server/tools/components/tools.py
+++ b/src/keboola_mcp_server/tools/components/tools.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Annotated, Sequence
+from typing import Annotated, Sequence, Optional, List, Dict, Any
 
 from mcp.server.fastmcp import Context, FastMCP
 from pydantic import Field
@@ -9,6 +9,8 @@ from keboola_mcp_server.tools.components.model import (
     ComponentConfiguration,
     ComponentType,
     ComponentWithConfigurations,
+    DataApp,
+    DataAppConfiguration,
 )
 from keboola_mcp_server.tools.components.utils import (
     _get_component_details,
@@ -18,7 +20,8 @@ from keboola_mcp_server.tools.components.utils import (
     _retrieve_components_configurations_by_ids,
     _retrieve_components_configurations_by_types,
 )
-from keboola_mcp_server.tools.sql import get_sql_dialect
+from keboola_mcp_server.tools.storage import retrieve_buckets, retrieve_bucket_tables
+from keboola_mcp_server.tools.sql import get_sql_dialect, query_table
 
 LOG = logging.getLogger(__name__)
 
@@ -55,6 +58,12 @@ def add_component_tools(mcp: FastMCP) -> None:
 
     mcp.add_tool(create_sql_transformation)
     LOG.info(f'Added tool: {create_sql_transformation.__name__}.')
+
+    mcp.add_tool(create_data_app)
+    LOG.info('Added tool: create_data_app.')
+
+    mcp.add_tool(_start_data_app)
+    LOG.info('Added tool: _start_data_app.')
 
     LOG.info('Component tools initialized.')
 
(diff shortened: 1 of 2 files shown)


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -331,4 +340,325 @@ async def create_sql_transformation(
         raise e
 
 
+async def _get_data_context(ctx: Context) -> Dict[str, Any]:
```
jordanrburger, line 343: I'm not sure about this. I would imagine that for many projects, it won't fit into context. It shouldn't be necessary overall either. In most instances, there should already be some notion of what tables are important based on the preceding conversation.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -56,6 +59,12 @@ def add_component_tools(mcp: FastMCP) -> None:
     mcp.add_tool(create_sql_transformation)
     LOG.info(f'Added tool: {create_sql_transformation.__name__}.')
 
+    mcp.add_tool(create_data_app)
+    LOG.info('Added tool: create_data_app.')
+
+    mcp.add_tool(_start_data_app)
```
mariankrotil, line 65: Do not start the tool name with `_`, as it evokes that it is a helper function, moreover, agent sees the tool name as `_start_data_app` unless you specify explicitly. Regarding the starting data app tool, we should think about creating a new PR and split those changes.

src/keboola_mcp_server/tools/components/tools.py
mariankrotil: Yes definitely agree with Jordan. Yes, we could rely on the agent's capabilities to decide what he needs or which tools he uses. To help him with that we can provide prompts with instructions for that.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -331,4 +340,325 @@ async def create_sql_transformation(
...
+        List of script lines
+    """
+    # Get data context from Keboola project
+    data_context = await _get_data_context(ctx)
+    
+    # Create a prompt for Claude that includes data context
+    # TODO: Make sure libraries that need to be used but are not in the list below are added to packages
+    # Use: - If any libraries are missing, add them to the packages list in config
+
+    prompt = f"""Generate a Streamlit script for a data app named '{name}' that {description}.
```
mariankrotil, line 443: We should create a separate `.py` file for keeping the prompt instruction templates, and move the prompt there.

ncapek:
Is this a draft or is it intended to be merged into main? `feature/component-cu-tools` is being just barely fixed to make it passable.  It's going to require a large refactor as it's currently very bloated. I'm **very** worried that piling on more code into component tools is going to make this problem even worse.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -331,4 +340,325 @@ async def create_sql_transformation(
...
+    """
+    Get context about available data in the Keboola project.
+    
+    Args:
+        ctx: The MCP server context
+        
+    Returns:
+        Dictionary containing information about available data
+    """
+    try:
```
ncapek, line 353: We specifically designed the `tool_errors` decorator so we don't have to have error handling logic in the tools. In general, I think the tools should be only coordinators of code, i.e. keep them as lean as possible. This is not your fault, since the pattern was copied from `feature/component-cu-tools`. 

One solution to getting out of this mess would be to create some kind of style guide where we would specify what standards and intended uses of existing tools are expected in future code. Otherwise we are going to have to explain everything to all new participants individually.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -331,4 +340,325 @@ async def create_sql_transformation(
...
+async def _get_data_context(ctx: Context) -> Dict[str, Any]:
+    """
+    Get context about available data in the Keboola project.
+    
+    Args:
+        ctx: The MCP server context
+        
+    Returns:
+        Dictionary containing information about available data
+    """
```
ncapek, line 352: We use the reST docstring format in this repo.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -331,4 +340,325 @@ async def create_sql_transformation(
...
+                    # Get a sample of the table data
+                    sample_data = await query_table(
+                        ctx,
+                        table_id=table.table_id,
+                        limit=5,
+                        columns=[]
+                    )
+                    
+                    # Get table info
+                    table_info = {
```
ncapek, line 384: Couldn't you re-use or extend the TableDetail class in `tools/storage.py`? It would easily allow you to construct it with `TableDetail.model_validate()` so you don't have to do all of this data processing inside of a tool definition.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -331,4 +340,325 @@ async def create_sql_transformation(
...
+                        'columns': list(sample_data[0].keys()) if sample_data else [],
+                        'sample_rows': [
+                            {k: str(v) for k, v in row.items()} 
+                            for row in sample_data
+                        ] if sample_data else []
+                    }
+                    bucket_info['tables'].append(table_info)
+                except Exception as e:
+                    LOG.warning(f"Could not get sample data for table {table.table_id}: {e}")
+                    bucket_info['tables'].append({
```
ncapek, line 397: This looks like a lot of duplication of code you wrote just a couple of lines above this.

ncapek (commented):
Key takeaways - I think it would be a good idea to create a style guide so everyone knows what is expected in new code, refactor component tools completely, and ideally ban all cursor generated code.

src/keboola_mcp_server/tools/components/tools.py
Nweaver412: Should this also go in the tools/components folder, or somewhere else?

PR #98, opened 2025-05-21
KAB-1030 Add user agent to API calls

radcliffkey:
also remove sync client from production deps

### DIFF
--- a/src/keboola_mcp_server/client.py
+++ b/src/keboola_mcp_server/client.py
@@ -2,6 +2,7 @@
 
 import importlib.metadata
 import logging
+import os
 from typing import Any, Mapping, Optional, Union, cast
 
 import httpx
@@ -16,7 +17,7 @@ JsonStruct = Union[JsonDict, JsonList]
 
 
 class KeboolaClient:
-    """Helper class to interact with Keboola Storage API and Job Queue API."""
+    """Class holding clients for Keboola APIs: Storage API, Job Queue API, and AI Service."""
 
     STATE_KEY = 'sapi_client'
     # Prefixes for the storage and queue API URLs, we do not use http:// or https:// here since we split the storage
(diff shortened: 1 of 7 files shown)


src/keboola_mcp_server/client.py
```diff
@@ -55,10 +55,33 @@ def __init__(
...
+        )
+        self.jobs_queue_client = JobsQueueClient.create(
+            root_url=queue_api_url, token=self.token, headers=self._get_headers()
+        )
+        self.ai_service_client = AIServiceClient.create(
+            root_url=ai_service_api_url, token=self.token, headers=self._get_headers()
+        )
+
+    @classmethod
+    def _get_user_agent(cls) -> str:
```
ncapek, line 69: What is this for?

radcliffkey: To create the user agent string to be sent to Keboola APIs for purposes of tracking requests coming from the MCP server

vita-stejskal (approved):

src/keboola_mcp_server/client.py
```diff
@@ -55,10 +56,35 @@ def __init__(
...
+            root_url=storage_api_url, token=self.token, headers=self._get_headers()
+        )
+        self.jobs_queue_client = JobsQueueClient.create(
+            root_url=queue_api_url, token=self.token, headers=self._get_headers()
+        )
+        self.ai_service_client = AIServiceClient.create(
+            root_url=ai_service_api_url, token=self.token, headers=self._get_headers()
+        )
+
+    @classmethod
```
mariankrotil, line 69: Shouldnt we use `@staticmethod`?

mariankrotil (approved):

src/keboola_mcp_server/client.py
radcliffkey: It could be static. I like `classmethod` because with static methods, you have to use them like `KeboolaClient._static_method()` even within `KeboolaClient`. With `classmethod` you can use `cls` or `self`, making it clear that you're using a method of the class itself, not some external class.

PR #99, opened 2025-05-21
Kab 998 Tests for server client communication

mariankrotil:
We are adding integration tests for end2end mcp tests - communication and connection between mcp server and client. Testing all transports and multiple client connections. Checkout from [KAB-998-fastmcp-2x-v3](https://github.com/keboola/mcp-server/tree/KAB-998-fastmcp-2x-v3)

changes:
- add test_mcp_server.py testing server client communication
- refactor conftest.py to provide the test environment variables


integtests/test_mcp_server.py
```diff
@@ -0,0 +1,325 @@
...
+AsyncContextClientRunner = Callable[
+    [Literal['sse', 'streamable-http'], str, dict[str, str] | None], _AsyncGeneratorContextManager[Client]
+]
+
+
+LOG = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def run_server_remote() -> AsyncContextServerRemoteRunner:
```
jordanrburger, line 31: Can't we utilize their new in-memory testing here? https://gofastmcp.com/patterns/testing#in-memory-testing

mariankrotil: Yes, we are using their in-memory testing in the `test_stdio_setup()` which assess server-client connection and communication using `stdio`  transport . We could also leverage using this `stdio` server-client setup elsewhere in the integration tests to test tool calls results through client-server communication. However, when we want to test the client connection to the remote server and its communication using `sse` or `streamable-http` transport we need to create the server on our own as a distinct thread, since I could not find (or they do not provide) in-memory testing for remote connections.

integtests/test_mcp_server.py
```diff
@@ -0,0 +1,359 @@
...
+    workspace_schema: str,
+    storage_api_url: str,
+):
+
+    setup = {
+        'storage_token': storage_api_token,
+        'workspace_schema': workspace_schema,
+        'storage_api_url': storage_api_url,
+    }
+    config = Config.from_dict(setup)
```
vita-stejskal, line 187: Umm, you could simply just create `Config` instance like this `Config(storage_token=storage_api_token, ...)`.

integtests/test_mcp_server.py
```diff
@@ -0,0 +1,359 @@
...
+
+    mocker.patch('keboola_mcp_server.mcp.os.environ', setup)
```
vita-stejskal, line 189: Do you really need this? The `setup` values are in `config` already and so the server should have all it needs.

integtests/test_mcp_server.py
```diff
@@ -0,0 +1,359 @@
...
+    run_client: AsyncContextClientRunner,
+    assert_basic_setup: Callable[[FastMCP, Client], Awaitable[None]],
+    assert_mcp_tool_call: Callable[[Client], Awaitable[None]],
+    storage_api_token: str,
+    workspace_schema: str,
+    storage_api_url: str,
+):
+
+    # we pass empty dict and test if it is set from the headers
+    config = Config.from_dict({})
```
vita-stejskal, line 267:
```suggestion
    config = Config()
```

integtests/test_mcp_server.py
radcliffkey: also on other places below

integtests/test_mcp_server.py
radcliffkey: Integration tests are using env vars prefixed by  `INTEGTEST`, e.g. `INTEGTEST_STORAGE_API_URL` to prevent using 'production' values you might have in your `.env` file. That was sufficient for testing individual tool which don't read env vars themselves but context is passed to them. In this case the server might actually use the 'production' env values instead of the integtest ones. Do I understand it correctly that that is the reason for this mocking? Maybe we should do the overriding right after loading the `.env` file to be sure.

mariankrotil: As @rado says, the values from config were overwritten by development environment variables which were configured from the `.env` file and then loaded by the server when using `os.environ`, I should have rather used an empty `{}` dict but I have changed the test so they use only the `INTEG` prefixed variables

mariankrotil: 8ebbbd586747c504616a3932873282b540aece3b

integtests/test_mcp_server.py
mariankrotil: 8ebbbd586747c504616a3932873282b540aece3b

integtests/test_mcp_server.py
```diff
@@ -0,0 +1,359 @@
...
+
+@pytest.fixture(scope='session')
+def assert_mcp_tool_call(configs: list[ConfigDef]):
+    """
+    Assert that the MCP tool call is correct.
+    """
+    _component_details_tool_name = 'get_component_details'
+
+    async def _assert_mcp_tool_call(client: Client):
+        for config in configs:
```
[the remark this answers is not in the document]
mariankrotil: 8ebbbd586747c504616a3932873282b540aece3b

integtests/test_mcp_server.py
```diff
@@ -0,0 +1,359 @@
...
+        client_resources = await client.list_resources()
+
+        assert len(client_tools) == len(server_tools)
+        assert len(client_prompts) == len(server_prompts)
+        assert len(client_resources) == len(server_resources)
+        assert all(expected == ret_tool.name for expected, ret_tool in zip(server_tools.keys(), client_tools))
+        assert all(expected == ret_prompt.name for expected, ret_prompt in zip(server_prompts.keys(), client_prompts))
+        assert all(
+            expected == ret_resource.name for expected, ret_resource in zip(server_resources.keys(), client_resources)
+        )
```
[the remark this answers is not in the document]
mariankrotil: 8ebbbd586747c504616a3932873282b540aece3b

integtests/test_mcp_server.py
```diff
@@ -0,0 +1,359 @@
...
+        if exception_from_client:
+            # we need to raise the exception from the client TaskGroup otherwise it will inform only
+            # about task group error
+            raise exception_from_client
+
+    return _run_client
+
+
+@pytest.fixture
+def assert_basic_setup():
```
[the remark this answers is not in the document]
mariankrotil: 8ebbbd586747c504616a3932873282b540aece3b

integtests/test_mcp_server.py
```diff
@@ -0,0 +1,359 @@
...
+    """fixture returning a _run_server_remote function"""
+
+    @asynccontextmanager
+    async def _run_server_remote(
+        server: FastMCP, transport: Literal['sse', 'streamable-http']
+    ) -> AsyncGenerator[str, None]:
+        """
+        Run the server in a async context manager which will ensure that the server is properly closed after the test.
+        The server is created with the given transport and port.
+        :yield: The url of the server.
```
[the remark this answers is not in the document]
mariankrotil: 8ebbbd586747c504616a3932873282b540aece3b

integtests/test_mcp_server.py
mariankrotil: 8ebbbd586747c504616a3932873282b540aece3b

mariankrotil:
> @mariankrotil , these tests look awesome! Thanks for adding them. I'd reorganize them a bit though to avoid the problems I mentioned in my comments.

Thanks I refactored the code.
- added fixture for replacing the os dev environment variables (expected by the server) to the test environment variables in the pytest thread in order to prevent errors using two different `tokens`.
- I converted the `assert_..` fixture methods to `_assert_...` helper functions
- `run_server` and `run_client` fixture moved to `contest`
- added comments, improved code readability

radcliffkey (approved):

integtests/conftest.py
```diff
@@ -64,21 +80,53 @@ class ProjectDef:
...
 
 
-def _storage_client() -> SyncStorageClient:
+@pytest.fixture(scope='session')
+def env_file_preloaded() -> bool:
+    return load_dotenv()
+
+
+@pytest.fixture(scope='session')
+def env_file_loaded(
```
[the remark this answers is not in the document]
mariankrotil: Done

PR #101, opened 2025-05-21
feat: Allow updating production Helm chart, cleanup Docker tags

wokoman:
This should ensure CI/CD for production deployments.

- `v*` tags produce `production-*` tag for production stacks.
- `v*` tags also produce `latest` tag.
- `v*-dev.*` tags prodduce `canary-orion-*` tag for Orion stack.

Removed some never-true tags from `docker/metadata-action` step.


Matovidlo (approved):
LGTM, so this means that update of production tag will also update testing stack right ? :slightly_smiling_face:

PR #107, opened 2025-05-23
Kab 1008 Validate Storage Configuration

mariankrotil:
We add validation of storage configuration generated by bot which the bot pass as tool argument in a dict type, we use @odinuv designed storage json schema for `storage` configuration within the component configuration.

We propose two approaches:
- using `jsonschema` (edited code from odin's repo) 
- using `pydantic.BaseModel.model_validate` method. (The pydantic models were generated from the odin's json schema using `datamodel-code-generator` library)

We will add tests once we decide which approach we are gonna use. We need to assess both approaches on real data. If the 2nd approach would work I would go for pydantic models


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -842,7 +846,7 @@ async def update_component_row_configuration(
         f'Updating configuration row: {name} for component: {component_id}, configuration id {configuration_id} '
         f'and row id {configuration_row_id}.'
     )
-
+    storage = validate_storage_pydantic(storage) if storage else None
```
ncapek, line 849: Shouldn't the `ConfigurationValidationError` just be passed into the `tools_error` decorator along with the recovery instructions?

mariankrotil: I would like to. I will modify it on Monday.

src/keboola_mcp_server/validators/validate.py
```diff
@@ -0,0 +1,104 @@
...
+            e, input_data=storage, json_schema=StorageSchema.model_json_schema()
+        )
+
+
+def _validate_json_against_schema(json_data: JsonStruct, schema: dict[str, Any]) -> bool:
+    """Validate JSON data against the provided schema."""
+    try:
+        jsonschema.validate(instance=json_data, schema=schema)
+        return True
+    except json.JSONDecodeError as e:
```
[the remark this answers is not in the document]
mariankrotil: removed 4bc0d92bbdd3361489a6f8c7eb20de457648988f

src/keboola_mcp_server/validators/exceptions.py
```diff
@@ -0,0 +1,112 @@
+"""
+Exceptions for the validators.
+"""
+
+from typing import Optional
+
+from keboola_mcp_server.client import JsonStruct
+
+
+class ConfigurationValidationError(Exception):
```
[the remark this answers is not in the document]
mariankrotil: Ok. Done here: 4bc0d92bbdd3361489a6f8c7eb20de457648988f

src/keboola_mcp_server/validators/storage_schema.py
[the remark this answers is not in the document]
mariankrotil: Ok removed, using only `jsonschema`

mariankrotil:
This is a screenshot how it looks now in Cursor ai:
image

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -0,0 +1,131 @@
...
+        str_repr += f'{self._RECOVERY_INSTRUCTIONS}\n'
+        if self.invalid_json:
+            str_repr += f'\nInvalid input data:\n{json.dumps(self.invalid_json, indent=2)}\n'
+        return str_repr.rstrip()
+
+
+def validate_storage(storage: JsonDict, initial_message: Optional[str] = None) -> JsonDict:
+    """
+    Validate the storage configuration using jsonschema.
+
```
ncapek, line 88: Empty line redundant

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -0,0 +1,131 @@
...
+        raise e  # unsupported error
+
+
+def _load_schema(json_schema_name: ConfigurationSchemaResourceName) -> dict[str, Any]:
+    try:
+        with resources.files(RESOURCES_PATH).joinpath(json_schema_name.value).open('r', encoding='utf-8') as f:
+            return json.load(f)
+    except FileNotFoundError:
+        LOG.exception(f'Schema {json_schema_name.value} not found in the package resources.')
+        raise FileNotFoundError(f'Schema {json_schema_name.value} not found in the package resources.')
```
ncapek, line 128: Does this error handling really have added value? If it fails we can just read the error from the original exception, no?

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -0,0 +1,131 @@
...
+
+from keboola_mcp_server.client import JsonDict
+
+LOG = logging.getLogger(__name__)
+
+
+RESOURCES_PATH = 'keboola_mcp_server.resources'
+
+
+class ConfigurationSchemaResourceName(str, Enum):
```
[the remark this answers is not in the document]
mariankrotil: Maybe schema for flows?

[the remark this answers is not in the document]
ncapek: My point is that currently it seems like overkill to store a single string value inside of an Enum class.

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -0,0 +1,131 @@
+"""
+Validator functions for Component Configuration data that are generated by agents.
```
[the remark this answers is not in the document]
mariankrotil: By using `_` for file prefix, I tried to distinguish between tool and utility submodules, such as validations, which are additionally treated as a private submodule and not meant for using outside this package. I could have considered creating a utils directory to put it there. What do you suggest?

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -0,0 +1,131 @@
...
+        raise RecoverableValidationError.create_from_values(e, invalid_json=json_data, initial_message=initial_message)
+    except jsonschema.SchemaError as e:
+        err_msg = (
+            f'The validation schema is not valid: {e} \n'
+            f'initial_message: {initial_message}\n'
+            f'schema: {schema}\n'
+            f'json_data: {json_data}'
+        )
+        LOG.error(f'{err_msg}')
+        raise jsonschema.SchemaError(f'{e}\n{err_msg}')  # this is not an Agent error, the schema is not valid
```
vita-stejskal, line 117: Umm, do we really want to fail the validation if the schema is invalid? This would effectively make it impossible for AI assistants to create/modify a component, if its developers failed to provide a valid configuration schema. And that is too restrictive, I think. I'd just log the exception (btw. it's better to use `LOG.exception()` to send the traceback in the log) and returned.

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -0,0 +1,131 @@
...
+        json_data=expected_input_data,
+        schema=schema,
+        initial_message=initial_message or STORAGE_VALIDATION_INITIAL_MESSAGE,
+    )
+    return storage
+
+
+def _validate_json_against_schema(
+    json_data: JsonDict, schema: dict[str, Any], initial_message: Optional[str] = None
+) -> bool:
```
vita-stejskal, line 102: Can this function ever return `False`? It not, I'd just remove the return type and return nothing. Also, `schema` parameter is `JsonDict` too, is it not?

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -0,0 +1,131 @@
...
+        return str_repr.rstrip()
+
+
+def validate_storage(storage: JsonDict, initial_message: Optional[str] = None) -> JsonDict:
+    """
+    Validate the storage configuration using jsonschema.
+
+    """
+    schema = _load_schema(ConfigurationSchemaResourceName.STORAGE)
+    expected_input_data = {'storage': storage.get('storage', storage)}
```
vita-stejskal, line 91: Why do you do this?

vita-stejskal: Should you do just `{'storage': storage}` or should you actually `return expected_input_data`?

mariankrotil: Due to the bot, the bot can pass either `{'storage':  {..}}` or without the `storage` key `{}` We want to accept both.

src/keboola_mcp_server/tools/_validate.py
mariankrotil: returning None, using JsonDict

PR #109, opened 2025-05-26
Kab 1064 update tools readme

ncapek:
Component tools have been updated into main and this PR aims to reflect these changes in the related files. Also upgrade to FastMCP 2.0 slightly broke the tools documentation script, hence the fix.

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -299,18 +299,18 @@ Executes an SQL SELECT query to get the data from the underlying database.
 Creates a component configuration using the specified name, component ID, configuration JSON, and description.
 
 CONSIDERATIONS:
-    The configuration JSON object must follow the root_configuration_schema of the specified component.
-    Make sure the configuration parameters always adhere to the root_configuration_schema,
-    which is available via the component_detail tool.
-    The configuration JSON object should adhere to the component's configuration examples if found.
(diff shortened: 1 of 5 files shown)


TOOLS.md
```diff
@@ -286,14 +285,160 @@ Executes an SQL SELECT query to get the data from the underlying database.
...
+    which is available via the component_detail tool.
+    The configuration JSON object should adhere to the component's configuration examples if found.
+
+USAGE:
+    - Use when you want to create a new root configuration for a specific component.
+
+EXAMPLES:
+    - user_input: `Create a new configuration for component X with these settings`
+        -> set the component_id and configuration parameters accordingly
+        -> returns the created component configuration if successful.
```
vita-stejskal, line 313: This markdown is broken. It should not be indented and the arrows `->` should be replaced with dashes `-`. You can look [here](https://github.com/keboola/mcp-server/blob/KAB-1064-update-tools-readme/TOOLS.md#create_component_root_configuration) to see how this is rendered. Please update the tools' docstrings. It seems to be a problem in all the component tools which we did not spot before merging the `feature/component-cu-tools` branch.

vita-stejskal (changes requested):

PR #112, opened 2025-05-26
Kab 1008 Validation of row and root parameter configurations

mariankrotil:
We add Row and Root parameter schema validation for CREATE and UPDATE component tools. The Row and root schemas can be fetched from the AI service along the component given the component id. When the schemas are not provided we skip the validation and proceeding with the update/creation
We modify:
- add validation of row / root parameter configuration along with storage.
- validation incorporated within the tools
- recovery exceptions for bot

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "0.25.0"
+version = "0.26.0"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/src/keboola_mcp_server/prompts/add_prompts.py
+++ b/src/keboola_mcp_server/prompts/add_prompts.py
@@ -1,0 +1,39 @@
+"""Module to add prompts to the Keboola MCP server."""
+
+from keboola_mcp_server.mcp import KeboolaMcpServer
+
+
+def add_keboola_prompts(mcp: KeboolaMcpServer) -> None:
+    """Add all Keboola-specific prompts to the MCP server.
+
+    The prompt names and descriptions are automatically derived from the function
+    names and docstrings.
+    """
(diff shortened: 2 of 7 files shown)


src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -285,55 +285,88 @@ def _get_transformation_configuration(
...
-            row_parameters = validate_row_parameters(row_parameters, component_id, schema)
-        else:
-            LOG.warning(f'No row parameter schema provided for component {component_id}, skipping validation.')
-    return {'storage': storage, 'row_parameters': row_parameters}
+    if not schema:
+        LOG.warning(f'No schema provided for component {component_id}, skipping validation.')
+        return parameters
+
+    # we expect the parameters to be a dictionary of parameter configurations without the "parameters" key
+    expected_parameters = parameters.get('parameters', parameters)
```
vita-stejskal, line 367: This should be done in `validate_parameters()` function so that its behavior is consistent with the behavior of `validate_storage()` function.

src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -285,55 +285,88 @@ def _get_transformation_configuration(
...
+    assert isinstance(expected_parameters, dict)
```
vita-stejskal, line 368: The `jsonschema` validator covers this problem or does it not? Either way raising `AssertionError` is wrong in this context, the `RecoverableValidationError` would be the correct one to raise.

vita-stejskal (commented):
Almost there, thanks for you patience.

src/keboola_mcp_server/tools/components/utils.py
mariankrotil: done: 66848c0d09aa24ccebe5cac187ae3f5031a5f44b
it was due to the type checker

vita-stejskal (approved):

PR #119, opened 2025-05-27
KAB-1068: stop reading the server config from the URL query part unless --accept-secrets-in-url CLI parameter is set

vita-stejskal:


src/keboola_mcp_server/mcp.py
```diff
@@ -151,7 +151,8 @@ async def _inject_session_state(*args, **kwargs) -> Any:
                 config = config.replace_by(os.environ)
                 if http_rq := _get_http_request():
                     config = config.replace_by(http_rq.headers)
-                    config = config.replace_by(http_rq.query_params)
+                    if config.accept_secrets_in_url:
+                        config = config.replace_by(http_rq.query_params)
```
mariankrotil, line 155: This part of the code seems potentially vulnerable. It looks like we could receive and set `accept_secrets_in_url` from the headers in a request and then inject parameters like `storage_token` into the URL. If the intended behavior is to let the user decide whether to use headers or query parameters within a single server context, then it is ok.

vita-stejskal: Good thinking! Fixed in b28958e, thanks.

PR #120, opened 2025-05-28
update reduced component flag field

ncapek:
Update the the `flags` field in `ReducedComponent` model


davidesner:
@ncapek I added another fix related to this which adds the additional flag that marks the input mapping presence.

src/keboola_mcp_server/tools/components/model.py
```diff
@@ -34,12 +34,13 @@ class ReducedComponent(BaseModel):
         serialization_alias='componentType',
     )
 
-    flags: list[str] = Field(
+    component_flags: list[str] = Field(
         default_factory=list,
         description='List of developer portal flags.',
+        validation_alias=AliasChoices('flags', 'component_flags', 'componentFlags', 'component-flags')
```
mariankrotil, line 40: what about serialization alias? do we want the default `component_flags`?

ncapek: Done

PR #123, opened 2025-05-28
Kab 1057 flow get schema tool

ncapek:
This PR adds a tool for fetching the flow schema to help the agent construct valid flows.


ncapek:
Something went wrong with the commit history. Let me fix that.

src/keboola_mcp_server/resources/flow_schema.json
```diff
@@ -0,0 +1,134 @@
+{
```
mariankrotil, line 1: use `kebab` case for the resource file name to align with `storage-schema` or rename storage schema

PR #124, opened 2025-05-29
Kab 1057 flow get schema tool v2

ncapek:
This PR implements the `get_flow_schema` tool and updates the tool docstrings to help coordinate the agent decision making.

### DIFF
--- a/src/keboola_mcp_server/tools/flow.py
+++ b/src/keboola_mcp_server/tools/flow.py
@@ -1,12 +1,14 @@
 """Flow management tools for the MCP server (orchestrations/flows)."""
 
+import json
 import logging
+from importlib import resources
 from typing import Annotated, Any, Dict, List, Optional
 
 from fastmcp import Context, FastMCP
 from pydantic import Field
 
-from keboola_mcp_server.client import KeboolaClient
+from keboola_mcp_server.client import JsonDict, KeboolaClient
 from keboola_mcp_server.errors import tool_errors
 from keboola_mcp_server.mcp import with_session_state
 from keboola_mcp_server.tools.components.model import (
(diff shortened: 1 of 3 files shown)


src/keboola_mcp_server/tools/flow.py
```diff
@@ -175,9 +162,9 @@ async def update_flow(
 @with_session_state()
 async def retrieve_flows(
     ctx: Context,
-    flow_ids: Annotated[Optional[List[str]], Field(
-        description='Optional list of specific flow configuration IDs'
-    )] = None,
+    flow_ids: Annotated[
+        Optional[List[str]], Field(description='Optional list of specific flow configuration IDs')
```
mariankrotil, line 166: Optional caused problems with agents to call the tool. We should could use `Sequence[str], ... = tuple()` or default factory `list`

src/keboola_mcp_server/tools/flow.py
```diff
@@ -77,46 +46,59 @@ def add_flow_tools(mcp: FastMCP) -> None:
...
+    ],
 ) -> Annotated[FlowConfiguration, Field(description='Created flow configuration')]:
-    """Creates a new flow configuration in Keboola orchestrator.
+    """Creates a new flow configuration in Keboola.
 
     Flow configurations are special - they store phases/tasks directly under 'configuration',
     not under 'configuration.parameters' like other components.
 
-    The schema above shows the required structure for phases and tasks."""
+    Prior to calling create flow, agents should call the `get_flow_schema()` to fetch the flow schema, if not already
```
mariankrotil, line 85: We should leave the decision on the bot though, and we should not include instructions in the tool description at all.  Instead we could provide hints or guide where to get those information.

src/keboola_mcp_server/tools/flow.py
```diff
@@ -132,26 +114,31 @@ async def update_flow(
...
+    ],
+    tasks: Annotated[List[Dict[str, Any]], Field(description='Updated list of task definitions.')],
+    change_description: Annotated[str, Field(description='Description of changes made')],
 ) -> Annotated[FlowConfiguration, Field(description='Updated flow configuration')]:
-    """Updates an existing flow configuration."""
+    """Updates an existing flow configuration.
+
+    Flow configurations are special - they store phases/tasks directly under 'configuration',
+    not under 'configuration.parameters' like other components.
+
```
mariankrotil, line 131: I would remove this message and handle it by our own, within the function, we could safely use  `tasks.get('parameters', tasks)` or `phases.get('parameters', phases)` to unwrap this.

The same in other tool's description

src/keboola_mcp_server/tools/flow.py
vita-stejskal: I'd structure the description using `CONSIDERATIONS:`, `USAGE:`, `EXAMPLES:` sections, see for example `create_sql_transformation` tool. So this could be something like:
```
"""
    Creates a new flow configuration in Keboola. A flow is a special Keboola component that orchestrates other components.
    The components within a flow are represented by tasks and are grouped into phases. Tasks within a phase
    can run in parallel, while phases can depend on each other and run sequentially.

    CONSIDERATIONS:
    - Make sure that `phases` and `tasks` parameters conform to the JSON schema of the flow configuration.
    - Make sure that each phase and task has `id` and `name` besides of other parameters.

    USAGE:
    - Use this tool to create a new flow for orchestrating components.

    EXAMPLES:
    - user_input: Orchestrate all my JIRA extractors.
        - fill `tasks` parameter with the tasks for the JIRA extractors
        - determine dependencies between the JIRA extractors
        - fill `phases` parameter by grouping tasks into phases
"""
```

src/keboola_mcp_server/tools/flow.py
```diff
@@ -77,46 +46,59 @@ def add_flow_tools(mcp: FastMCP) -> None:
     LOG.info('Flow tools initialized.')
 
 
+@tool_errors()
+@with_session_state()
+async def get_flow_schema(ctx: Context) -> Annotated[str, Field(description='The configuration schema of Flow')]:
+    """Gets the schema for Flow configurations. Should be called prior to calling 'create_flow' and used as a
+    template for creating/updating flows."""
```
vita-stejskal, line 53: I'd change this to:
```
"""Returns the JSON schema that defines the structure of Flow configurations."""
```

src/keboola_mcp_server/tools/flow.py
ncapek: Resolved by merging from branch by @vita-stejskal

src/keboola_mcp_server/tools/flow.py
ncapek: I pushed a new version of the docstrings

vita-stejskal (approved):

PR #131, opened 2025-06-02
Kab 1092 validate flow schema

ncapek:
This PR add flow configuration validation against validation schema. It also simplified the Flow models to remove unused fields.


ncapek:
FYI: I think it's time to think about a file organization refactor. `tools/components/model.py` contains models related to flows. `_validate.py` contains validation functions for a variety of different object types. And `tools/components/utils.py` contain wrappers of `_validate.py` function which I don't really see much value in. Additionally functions like `validate_row_parameters_configuration` etc in the `utils.py` module do not only validate - they modify the json in some cases => they should probably be renamed to reflect this.

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -113,9 +114,23 @@ def validate_parameters(parameters: JsonDict, schema: JsonDict, initial_message:
     return {'parameters': expected_input}  # normalized to {"parameters" : {...}}
 
 
-def _validate_json_against_schema(
-    json_data: JsonDict, schema: JsonDict, initial_message: Optional[str] = None
-):
+def validate_flow_configuration_against_schema(flow: JsonDict, initial_message: Optional[str] = None) -> JsonDict:
```
mariankrotil, line 117: What if we receive `flow` as {parameters: {config}} instead of {config} from the agent do we handle it somewhere?

src/keboola_mcp_server/tools/components/model.py
```diff
@@ -339,20 +339,13 @@ class FlowPhase(BaseModel):
...
         description='List of phase IDs this phase depends on',
         validation_alias=AliasChoices('dependsOn', 'depends_on', 'depends-on'),
-        serialization_alias='dependsOn'
-    )
-    behavior: Optional[dict[str, Any]] = Field(None, description='Phase behavior configuration')
-    child_behavior: Optional[dict[str, Any]] = Field(
-        None,
-        description='Child task behavior configuration',
-        validation_alias=AliasChoices('childBehavior', 'child_behavior', 'child-behavior'),
-        serialization_alias='childBehavior'
```
mariankrotil, line 355: Why are we removing this? Wont it be causing error when the `FlowPhase` is created from the JSON data provided by the agent?

src/keboola_mcp_server/tools/flow.py
```diff
@@ -99,6 +100,8 @@ async def create_flow(
         'phases': [phase.model_dump(by_alias=True) for phase in processed_phases],
         'tasks': [task.model_dump(by_alias=True) for task in processed_tasks],
     }
+    flow_configuration = cast(JsonDict, flow_configuration)
+    validate_flow_configuration_against_schema(flow_configuration)
```
mariankrotil, line 104: We can consider providing the message to the validation method to closely specify the validation error when appears, maybe.

src/keboola_mcp_server/tools/_validate.py
ncapek: I can't imagine how this would happen. It's not compliant to the schema and `retrieve_flows` return flow summaries, which does not conform to this format.

src/keboola_mcp_server/tools/flow.py
mariankrotil: perhaps something like this? Because when the validation fails, the exact location error of violated schema is returned, we should somehow inform the bot that the fields are validated together but he has to provide them separated, maybe.
```
The fields "phases" or "tasks" are invalid.
The provided flow configuration does not follow the flow schema. The flow configuration consists of phases and tasks which is validated together as a single unit.
```

vita-stejskal:
> tools/components/model.py contains models related to flows.

See [KAB-1084](https://keboola.atlassian.net/browse/KAB-1084) task.

[KAB-1084]: https://keboola.atlassian.net/browse/KAB-1084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/keboola_mcp_server/tools/flow.py
mariankrotil: Yes it points to there. But I was wondering whether the bot can recognize phases and tasks and correctly pass it to the tool because when the error occurs the schema uses concatenation of phases and tasks, so it points to that concatenation.

src/keboola_mcp_server/tools/components/model.py
ncapek: @odinuv @radektomasek @davidesner Can any of you comment on the purpose of the `behaviour` and `child_behavior` fields? It is not being used anywhere in the codebase, therefore I removed it. But it might serve some other purpose I'm not aware of at this time.

PR #133, opened 2025-06-03
KAB-1030 Add metadata for configs created/updated by MCP

radcliffkey:


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -535,12 +534,21 @@ async def create_component_root_configuration(
...
-        f'Created new configuration for component "{component_id}" with configuration id '
-        f'"{new_configuration.configuration_id}".'
+        f'Created new configuration for component "{component_id}" with configuration id "{configuration_id}".'
     )
 
+    try:
+        await client.storage_client.configuration_metadata_update(
+            component_id=component_id,
+            configuration_id=configuration_id,
+            metadata={'KBC.MCP.createdBy': 'true'},
```
vita-stejskal, line 547: Do we also want to record the config versions created/updated by the MCP server? I'm not sure what was the outcome of the design discussion that we had. But I think that we agreed that we will have a metadata key (e.g. `KBC.MCP.updatedVersions`) which value would be a growing list of the configuration versions (e.g. `1 5 12`) that were created/updated by the MCP server.

Also, I think that similar code should also be added to the other create/update root/row configuration tools.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -535,12 +534,21 @@ async def create_component_root_configuration(
...
+        )
+    except HTTPStatusError as e:
+        LOG.error(f'Failed to set "KBC.MCP.createdBy" metadata for configuration {configuration_id}: {e}')
```
vita-stejskal, line 550: `LOG.exception` maybe to have the stacktrace logged as well.

integtests/components/test_tools.py
```diff
@@ -79,3 +85,155 @@ async def test_retrieve_components_by_types(mcp_context: Context, configs: list[
...
+        assert isinstance(metadata, list)
+        metadata_dict = {item['key']: item['value'] for item in metadata if isinstance(item, dict)}
+        assert 'KBC.MCP.createdBy' in metadata_dict
+        assert metadata_dict['KBC.MCP.createdBy'] == 'true'
+    finally:
+        # Delete the configuration
+        await client.storage_client.configuration_delete(
+            component_id=component_id,
+            configuration_id=result.configuration_id
+        )
```
odinuv, line 149: please call the configuration_delete twice - to delete the configuration from the recycle bin as well

src/keboola_mcp_server/client.py
```diff
@@ -284,7 +287,7 @@ async def put(
     async def delete(
         self,
         endpoint: str,
-    ) -> JsonStruct:
+    ) -> JsonStruct | None:
```
odinuv, line 290: afaik delete never returns JsonStruct

odinuv (approved):
lgtm

integtests/components/test_tools.py
```diff
@@ -79,3 +85,155 @@ async def test_retrieve_components_by_types(mcp_context: Context, configs: list[
...
+
+        # Verify the configuration exists in the backend by fetching it
+        config_detail = await client.storage_client.configuration_detail(
+            component_id=component_id,
+            configuration_id=updated_result.configuration_id
+        )
+
+        assert config_detail['name'] == updated_name
+        assert config_detail['description'] == updated_description
+        assert 'configuration' in config_detail
```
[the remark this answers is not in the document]
radcliffkey: The only difference is that now I'd get assertion error instead of key error, which makes sense in tests

src/keboola_mcp_server/client.py
```diff
@@ -425,6 +439,37 @@ async def configuration_list(self, component_id: str) -> list[JsonDict]:
 
         return cast(list[JsonDict], await self.get(endpoint=endpoint))
 
+    async def configuration_metadata_get(self, component_id: str, configuration_id: str) -> JsonList:
```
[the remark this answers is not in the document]
radcliffkey: This makes clear that it is a separate API call. But we might add and metadata option to `configuration_detail()` later if we find it useful

PR #135, opened 2025-06-03
Kab 1096 add project id endpoint

ncapek:
This PR adds an endpoint to fetch the data about tokens and projects.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "0.27.3"
+version = "0.27.4"
 description = "MCP server for interacting with Keboola Connection"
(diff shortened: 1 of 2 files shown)


vita-stejskal (approved):

vita-stejskal:
> @vita-stejskal I would have preferred to parse out the data and return the `project_id` only but I figured it's a waste to dump the remaining data. I just need to figure out where the parsing function should live now.

The `verify_token()` function looks good and I agree that it's better if it returns whatever data it gets from SAPI. We are likely to need that information later on anyway. We already use that info in the integration tests (on `KAB-1069-dynamic-workspace` branch).

> I just need to figure out where the parsing function should live now.

I am not sure what parsing you talk about. Whoever calls the `verify_token()` function can pick whatever information they need. What do you want to parse?

PR #137, opened 2025-06-03
Kab 1097 simplify and link response objects

ncapek:
This PR implement updates to the flow response tools. That is: 1) simplify the response objects to return only response status + some basic information, 2) return the link to the created object. The links are constructed from the `project_id`, hence a project manager class is also added.


src/keboola_mcp_server/tools/workspace.py
```diff
@@ -288,3 +288,24 @@ async def get_quoted_name(self, name: str) -> str:
     async def get_sql_dialect(self) -> str:
         workspace = await self._get_workspace()
         return workspace.get_sql_dialect()
+
+
+class ProjectManager:
```
ncapek, line 293: `workspace.py` is not an ideal place to put this but creating a separate module just for this one class seemed even worse.

src/keboola_mcp_server/tools/flow.py
```diff
@@ -168,10 +185,14 @@ async def update_flow(
         flow_configuration=flow_configuration,  # Direct configuration
     )
 
-    updated_flow_response = FlowConfigurationResponse.from_raw_config(updated_raw_configuration)
+    flow_id = str(updated_raw_configuration.get('id', ''))  # Could this just be configuration_id instead?
+    project_id = await project_manager.get_project()
+    base_url = client.storage_client.base_api_url.split('/v2')[0]
+    flow_link = get_flow_url(base_url=base_url, project_id=project_id, flow_id=flow_id)
```
mariankrotil, line 191: We could consider creating a dedicated class or module that encapsulates the logic for building these links. It would provide methods for generating URLs using `client`, `project_id` and `base_url` for various tools like components, jobs, or storage. I believe this will eventually be needed for other tools as well, so it makes sense to centralize this functionality early.

src/keboola_mcp_server/tools/workspace.py
mariankrotil: Do we really need `ProjectManager` for this? I am not sure how computationally expensive token verification is, but perhaps we could handle it directly through the client, either by retrieving it or caching it there.

src/keboola_mcp_server/tools/workspace.py
```diff
@@ -288,3 +288,24 @@ async def get_quoted_name(self, name: str) -> str:
...
+    def from_state(cls, state: Mapping[str, Any]) -> 'ProjectManager':
+        instance = state[cls.STATE_KEY]
+        assert isinstance(instance, ProjectManager), f'Expected ProjectManager, got: {instance}'
+        return instance
+
+    def __init__(self, client: KeboolaClient):
+        self.project = None
+        self.client = client
+
+    async def get_project(self) -> str | int:
```
mariankrotil, line 306: we should rename it to `get_project_id` to capture its purpose

ncapek: Renamed

src/keboola_mcp_server/tools/workspace.py
```diff
@@ -420,3 +420,24 @@ async def get_quoted_name(self, name: str) -> str:
...
+    def from_state(cls, state: Mapping[str, Any]) -> 'ProjectManager':
+        instance = state[cls.STATE_KEY]
+        assert isinstance(instance, ProjectManager), f'Expected ProjectManager, got: {instance}'
+        return instance
+
+    def __init__(self, client: KeboolaClient):
+        self.project = None
+        self.client = client
+
+    async def get_project_id(self) -> str | int:
```
radcliffkey, line 438: it always returns `str`

src/keboola_mcp_server/tools/workspace.py
```diff
@@ -420,3 +420,24 @@ async def get_quoted_name(self, name: str) -> str:
...
+        if self.project is None:
+            # This should probably be handled for exceptions somehow - unclear how at this time
```
radcliffkey, line 440: I'm not a big fan of these todo comments. I think if we cannot retrieve the project ID, we should throw an exception, because it should not happen and without the project ID, you cannot construct a working link. Alternatively, you can catch exceptions and log an error here and return `None`. But then you'd need to handle it in the tools and throw an exception there.

src/keboola_mcp_server/tools/workspace.py
```diff
@@ -420,3 +420,24 @@ async def get_quoted_name(self, name: str) -> str:
...
+    STATE_KEY = 'project_manager'
+
+    @classmethod
+    def from_state(cls, state: Mapping[str, Any]) -> 'ProjectManager':
+        instance = state[cls.STATE_KEY]
+        assert isinstance(instance, ProjectManager), f'Expected ProjectManager, got: {instance}'
+        return instance
+
+    def __init__(self, client: KeboolaClient):
+        self.project = None
```
radcliffkey, line 435:
```suggestion
        self._project_id = None
```

src/keboola_mcp_server/tools/workspace.py
radcliffkey, line 438: this can be done as a `project_id` property https://docs.python.org/3/library/functions.html#property

src/keboola_mcp_server/tools/workspace.py
```diff
@@ -420,3 +420,24 @@ async def get_quoted_name(self, name: str) -> str:
...
+        self.client = client
+
+    async def get_project_id(self) -> str | int:
+        if self.project is None:
+            # This should probably be handled for exceptions somehow - unclear how at this time
+            project_data = await self.client.storage_client.verify_token()
+            self.project = project_data.get('owner', {}).get('id')
```
radcliffkey, line 442: We shouldn't silently allow the project_id to be `None`. Also, `str(self.project)` would return string `'None'`.

src/keboola_mcp_server/tools/workspace.py
ncapek: Fixed

ncapek: I foresee a future where we support multiple projects, so it didn't seem like the right approach to make it a property.

src/keboola_mcp_server/tools/workspace.py
ncapek: Fixed

src/keboola_mcp_server/tools/workspace.py
ncapek: It's removed now

src/keboola_mcp_server/tools/workspace.py
ncapek: Project manager is gone altogether and tools will always fetch the project id.

src/keboola_mcp_server/client.py
```diff
@@ -624,6 +628,20 @@ async def flow_create(
...
         )
 
+    async def flow_delete(self, configuration_id: str, skip_trash: bool = False) -> None:
+        """
+        Deletes a flow configuration.
+
+        :param configuration_id: The id of the flow configuration.
+        :param skip_trash: If True, the configuration is deleted without moving to the trash.
+            (Technically it means the API endpoint is called twice.)
+        :raises ValueError: If the configuration_id is invalid.
```
radcliffkey, line 638: does it really? wouldn't it be just http error - bad request or something? I now see that I have made a similar docstring error in  `configuration_delete` where I don't really validate `component_id` or `configuration_id`

src/keboola_mcp_server/tools/flow.py
```diff
@@ -104,16 +127,22 @@ async def create_flow(
...
 
     new_raw_configuration = await client.storage_client.flow_create(
         name=name, description=description, flow_configuration=flow_configuration  # Direct configuration
     )
 
-    flow_response = FlowConfigurationResponse.from_raw_config(new_raw_configuration)
+    flow_id = str(new_raw_configuration['id'])
+    flow_name = new_raw_configuration['name']
+    project_id = await client.storage_client.project_id()
+    base_url = client.storage_client.base_api_url.split('/v2')[0]
```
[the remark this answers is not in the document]
ncapek: How would you name it? Since it's being taken from the `storage_client`, the base url is everything including the "/v2", no?

src/keboola_mcp_server/client.py
ncapek: Yeah, this was just copy-pasted from your implementation.

src/keboola_mcp_server/client.py
```diff
@@ -624,6 +628,20 @@ async def flow_create(
...
+        """
+        endpoint = f'branch/{self.branch_id}/components/{ORCHESTRATOR_COMPONENT_ID}/configs/{configuration_id}'
```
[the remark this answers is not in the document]
ncapek: Done

src/keboola_mcp_server/tools/flow.py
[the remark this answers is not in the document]
ncapek: Nevermind. Fixing now.

tests/tools/test_flows.py
```diff
@@ -363,9 +375,11 @@ async def test_create_flow(
             tasks=sample_tasks,
         )
 
-        assert isinstance(result, FlowConfiguration)
-        assert len(result.phases) == 2
-        assert len(result.tasks) == 2
+        assert isinstance(result, FlowToolResponse)
+        assert result.description == 'Test flow description'
+        assert result.timestamp == parser.isoparse('2025-05-25T06:33:41+0200')
```
radcliffkey, line 380: where does this come from? Maybe we should have constant for it.

src/keboola_mcp_server/client.py
radcliffkey: can you fix it please?

ncapek: Fixed

src/keboola_mcp_server/tools/flow.py
```diff
@@ -357,3 +393,24 @@ def has_cycle(phase_id: Any, _visited: set, rec_stack: set, path: list[Any]) ->
...
+def get_flow_url(base_url: str, project_id: str, flow_id: str | int) -> str:
+    return f'{base_url}/admin/projects/{project_id}/flows/{flow_id}'
+
+
+def get_flows_url(base_url: str, project_id: str) -> str:
+    return f'{base_url}/admin/projects/{project_id}/flows'
+
+
+def get_flow_urls(base_url: str, project_id: str, flow_id: str | int, flow_name: str) -> list[FlowURL]:
+    flow_url = FlowURL.model_validate(
```
radcliffkey, line 407: `flow_url = FlowURL(type=..)`

ncapek: I don't understand. The type is specified in the dict that is passed to `.model_validate()`

integtests/test_flow_tools.py
```diff
@@ -0,0 +1,173 @@
...
+    try:
+        assert isinstance(created, FlowToolResponse)
+        assert created.description == flow_description
+        assert created.success is True
+        assert len(created.links) == 3
+
+        flows = await retrieve_flows(mcp_context)
+        assert any(f.name == flow_name for f in flows)
+        found = [f for f in flows if f.name == flow_name][0]
+        assert found.id == created.flow_id
```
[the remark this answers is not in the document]
ncapek: Actually, the `created` object does not return the name, so this can't be tested as is.

src/keboola_mcp_server/tools/flow.py
ncapek: Fixed

radcliffkey (approved):

PR #142, opened 2025-06-05
Kab 1108 release tool params

mariankrotil:
We update the parameters because the agent fails to call the tools correctly on the first attempt.
Also, fix the test by defining the client before it's used.
See screenshots
Before
Screenshot 2025-06-05 171740
image
After
image


mariankrotil:
With having the idea in mind, we are closing this PR. The current approach worked well; however, it shows a slightly hacky way to encourage agent to use the input parameters within calling the tool. As the method lies in wrapping the input parameters of `dict` types into a separate `Pydantic` models having just one field of that input parameter `dict` type, we decided to wait for new advanced testing scenarios which could reveal new problems or confirm this one, potentially resulting in handling the problem via applying the current approach.  Moreover, hopefully, it could be solved by development on the cursor client side too.

PR #144, opened 2025-06-06
Kab 1108 storage necessity validation

mariankrotil:
We add validation of storage config necessity - During testing, we encountered the bot tried to update transformation without storage, it has to be forbidden.

See the error message. 
Screenshot 2025-06-06 140816
Screenshot 2025-06-06 152715

### DIFF
--- a/src/keboola_mcp_server/tools/components/utils.py
+++ b/src/keboola_mcp_server/tools/components/utils.py
@@ -325,15 +325,18 @@ ROW_PARAMETERS_VALIDATION_INITIAL_MESSAGE = (
 
 def validate_storage_configuration(
     storage: Optional[JsonDict],
+    component: Component,
     initial_message: Optional[str] = None,
 ) -> JsonDict:
     """
-    Validates the storage configuration and extracts the storage key contents.
+    Validates the storage configuration and checks if it is necessary for the component.
     :param storage: The storage configuration to validate received from the agent.
+    :param component: The component for which the storage is provided
     :param initial_message: The initial message to include in the error message.
     :return: The contents of the 'storage' key from the validated configuration,
               or an empty dict if no storage is provided.
     """
+    _validate_storage_necessity(storage=storage, component=component)
     if not storage or storage is None or storage.get('storage', {}) is None:
         LOG.warning('No storage configuration provided, skipping validation.')
         return {}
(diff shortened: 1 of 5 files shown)


src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -343,45 +346,57 @@ def validate_storage_configuration(
...
+    """
+    if not storage or storage.get('storage') == {}:
+        # necessary storage for writer and transformation components
+        if component.component_type in ['writer', 'transformation']:
+            raise ValueError(
+                f'Storage configuration cannot be empty for component {component.component_id} of type '
+                f'{component.component_type}. Please provide the storage configuration within the tool call.'
+            )
+
+
```
mariankrotil, line 364: @odinuv
Are the ui-flags related to storage - like hasInputTableMapping provided in the component info - considered stable?
Can I rely on them when validating storage configuration?

If so, we could add checks such as component.has_input_table_mappings, component.has_file_output_mappings, etc. And if any of them are True, we would require a storage configuration.
Does this apply to both row and root configurations?

src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -325,15 +325,18 @@ def _get_transformation_configuration(
...
     """
-    Validates the storage configuration and extracts the storage key contents.
+    Validates the storage configuration and checks if it is necessary for the component.
     :param storage: The storage configuration to validate received from the agent.
+    :param component: The component for which the storage is provided
     :param initial_message: The initial message to include in the error message.
     :return: The contents of the 'storage' key from the validated configuration,
               or an empty dict if no storage is provided.
     """
+    _validate_storage_necessity(storage=storage, component=component)
```
ncapek, line 339: Why doesn't this just return a bool? Seems like there is a lot of overlap between the checking in `_validate_storage_necessity` and the following line of code.

tests/tools/components/test_utils_validation.py
```diff
@@ -17,13 +17,46 @@
...
-def test_validate_storage_configuration_output(input_storage: Optional[JsonDict], output_storage: Optional[JsonDict]):
+def test_validate_storage_configuration_output(
+    mock_component: dict, input_storage: Optional[JsonDict], output_storage: Optional[JsonDict]
+):
     """testing normalized and returned structures {storage: {...}} vs {...}"""
-    result = utils.validate_storage_configuration(input_storage)
+    component_raw = mock_component.copy()
+    component_raw['type'] = 'extractor'  # set storage to {} to pass the validation for storage necessity
+    component = Component.model_validate(component_raw)
+    result = utils.validate_storage_configuration(input_storage, component)
```
ncapek, line 27: I don't think anything name `validate` should be returning a value. The naming would suggest that it raises an error instead. Also, the `validate_storage_configuration` does some kind of normalization inside, so the naming is off for that reason as well.

mariankrotil: Okay, what do you suggest? I think the validate function should check whether the data conforms to a predefined schema and return the validated data if it passes validation, and potentially normalized it, similar to how `pydantic.BaseModel.model_validate` works. In case it does not follow the schema, the error is raised.

mariankrotil: As we discussed in person, we clarified that the mentioned points are related and part of the codebase for our JSON Schema validation of storage, flow, and parameter. However, we agreed on and implemented a few refactoring changes to simplify the validation codebase. As a result, schema validation remains in` _validate.py`, while the logic for handling possible agent input/output has been moved to `components/utils.py`. This module encapsulates the schema validation methods and adds additional logic related to components.

src/keboola_mcp_server/tools/_validate.py
```diff
@@ -164,34 +167,40 @@ def _sanitize_required_and_properties(
...
+        json_data=storage,
         schema=schema,
         initial_message=initial_message,
     )
-    return normalized_storage_data
+    return storage
 
 
-def validate_parameters(parameters: JsonDict, schema: JsonDict, initial_message: Optional[str] = None) -> JsonDict:
+def validate_parameters_configuration_against_schema(
```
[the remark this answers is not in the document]
mariankrotil: To follow the `flow` schema validation function name we may leave it as it is.

src/keboola_mcp_server/tools/components/utils.py
davidesner: @mariankrotil hasInputTableMapping comes from there right https://github.com/keboola/mcp-server/blob/4f07a58fffef53518eae725599354f2bd3285484/src/keboola_mcp_server/tools/components/model.py#L91 ? Or the logic was moved somewhere to ai service. If it has any of these flags `'genericDockerUI-tableInput', 'genericDockerUI-simpleTableInput'` then it has input mapping (displayed in the UI). That doesn't mean that the input mapping is required.

Having storage={} is actually fine, it's just can't be null, otherwise the UI will break. SQL transformation should have output mapping in 99% of cases. Python transformation does not need an input mapping at all (storage={}). So that check on component type transformation is too strict.

With writers we can be sure that the input mapping is required.

src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -325,63 +332,84 @@ def _get_transformation_configuration(
...
+        LOG.warning(
+            'No storage configuration provided for component %s of type %s.',
+            component.component_id,
+            component.component_type,
+        )
+        storage_cfg = {}
+    # Only for SQL transformations
+    if component.component_id in [SNOWFLAKE_TRANSFORMATION_ID, BIGQUERY_TRANSFORMATION_ID]:
+        # For SQL transformations, we want to have input or output config in the storage
+        if not storage_cfg or ('output' not in storage_cfg and 'input' not in storage_cfg):
```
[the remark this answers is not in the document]
mariankrotil: Yes, I hope I captured it there. But I check if neither of them. If `output` and `input` are not in the config, then we require at least one of them to be there. so if there is only `output` it is ok, if there is only `input` it is ok. Does it make sense?

src/keboola_mcp_server/tools/components/utils.py
radcliffkey, line 1: It seems a little inconsistent to put some validation functions here and some to `validation`. I suggest putting them all to `validation`.

PR #148, opened 2025-06-09
Kab 1083 project info tool

ncapek:
This PR add a tool to fetch the project info.


src/keboola_mcp_server/links.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_project_links(self, project_id: str) -> list[Link]:
+        """Return a list of relevant links for a project."""
+        project_url = self.get_project_url(project_id)
+        return [Link(type='ui-detail', title='Project Dashboard', url=project_url)]
+
+    def get_flow_links(self, project_id: str, flow_id: str | int, flow_name: str) -> list[Link]:
+        """Get a list of relevant links for a flow, including detail, dashboard, and documentation."""
+        flow_url = Link(type='ui-detail', title=f'Flow: {flow_name}', url=self.get_flow_url(project_id, flow_id))
+        flows_url = Link(
```
mariankrotil, line 41: Do not insist but to align with `documentation_url` we could consider `detail_url` and `dashboard_url`

src/keboola_mcp_server/tools/project.py
```diff
@@ -0,0 +1,91 @@
...
+    )
+    links: list[Link] = Field(..., description='The links relevant to the tool call.')
+
+
+@tool_errors()
+@with_session_state()
+async def get_project_info(
+    ctx: Context,
+) -> Annotated[ProjectInfo, Field(description='Structured project info.')]:
+    """Return structured project information pulled from multiple endpoints."""
```
mariankrotil, line 64: Probably in a new PR. We could highlight what the project info contains in the docstring (e.g., sql_dialect) and remove the get_sql_dialect tool as proposed. After that, we should update the docstrings of the tools that reference `get_sql_dialect` directly.

src/keboola_mcp_server/tools/project.py
```diff
@@ -0,0 +1,91 @@
...
+    client = KeboolaClient.from_state(ctx.session.state)
+    storage = client.storage_client
+
+    token_data = await storage.verify_token()
+    project_data = token_data.get('owner', {})
+    organization_id = token_data.get('organization', {}).get('id', '')
+
+    metadata = await storage.get('branch/default/metadata')
+    description = next((item['value'] for item in metadata if item.get('key') == MetadataField.PROJECT_DESCRIPTION), '')
```
mariankrotil, line 73: If the description could be `None` in some cases e.g. when the key is not present, we might handle such case, otherwise the `ProjectInfo.model_validate` fails as it does not accept `project_description`  to be `None`.

ncapek: I have the empty string as default at the end of the line.

src/keboola_mcp_server/links.py
```diff
@@ -0,0 +1,45 @@
...
+    type: URLType = Field(..., description='The type of the URL.')
+    title: str = Field(..., description='The name of the URL.')
+    url: str = Field(..., description='The URL.')
+
+
+class LinksManager:
+
+    FLOW_DOCUMENTATION_URL = 'https://help.keboola.com/flows/'
+
+    def __init__(self, base_url: str):
```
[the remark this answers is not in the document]
ncapek: Good idea

src/keboola_mcp_server/links.py
ncapek: Partially fixed. I named the variables `flow_detail_url` and `flows_dashboard_url`.

src/keboola_mcp_server/links.py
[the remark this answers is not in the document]
ncapek: Done

src/keboola_mcp_server/tools/project.py
mariankrotil: Ok, agree, it would require experimenting with agent and tools, hence, before the release we should keep with our current tools as you suggest.

mariankrotil (approved):

src/keboola_mcp_server/tools/project.py
```diff
@@ -0,0 +1,89 @@
...
+
+    sql_dialect = await WorkspaceManager.from_state(ctx.session.state).get_sql_dialect()
+    links = links_manager.get_project_links()
+
+    combined = {
```
radcliffkey, line 79: This is an unnecessarily obscure way ho to create the object. You already set all the fields except `project_id` and `project_name`, so why not set them all explicitly. Then you 1) don't need to use aliases 2) get some static code checks in your IDE instead of waiting for runtime failure of `model_validate`
```
ProjectInfo(
    project_id=...
    project_name=...
    ...
)
```

src/keboola_mcp_server/tools/project.py
```diff
@@ -0,0 +1,89 @@
...
+    project_tools = [get_project_info]
+
+    for tool in project_tools:
+        LOG.info(f'Adding tool {tool.__name__} to the MCP server.')
+        mcp.add_tool(tool)
+
+    LOG.info('Project tools initialized.')
+
+
+class ProjectInfo(BaseModel):
```
radcliffkey, line 28: please remove the aliases. These can be useful if the object is being deserialized from JSON sent over the network, but this is not the case - we have the object under our control.

ncapek: Removed

src/keboola_mcp_server/tools/project.py
ncapek: Fixed

radcliffkey (approved):

PR #154, opened 2025-06-10
KAB-1120 add a patch so that missing session doesn't return 404, but 401

tomasfejfar:
KAB-1120

The goal here is to gently nudge the client to reauthorize if session is somehow lost or broken.

### DIFF
--- a/patches/mcp-session-status-fix.patch
+++ b/patches/mcp-session-status-fix.patch
@@ -1,0 +1,31 @@
+diff --git a/src/mcp/server/sse.py b/src/mcp/server/sse.py
+index 1234567..abcdefg 100644
+--- a/src/mcp/server/sse.py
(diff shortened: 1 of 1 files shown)


patches/mcp-session-status-fix.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/src/mcp/server/sse.py b/src/mcp/server/sse.py
```
odinuv, line 1: This probably not intended to be merged this way

tomasfejfar: I didn't know how it's supposed to work tbh. I just kinda modeles it based on the rest of the patches.

odinuv: we decided on standup to postpone this

PR #155, opened 2025-06-11
Disable release to production

odinuv:
Keeps releases to canary enabled.

To enable releases to production, simply revert this PR.


radcliffkey:
On standup we agreed to do one more release today and then we can do the freeze

wokoman (approved):
Good idea.

If we need something urgently, we can manually bump it through `kbc-stacks`.

jordanrburger (approved):

PR #156, opened 2025-06-11
KAB-1123-transformation-fix-invalid-blocks

mariankrotil:
We renamed a field in the sql transformation configuration model and updated description and instructions of the tool to be more understandable for the agent.  Renamed `script` to `sql_statements` and add explicit description that `each sql statement must be executable and follow current dialect`


mariankrotil:
image

PR #166, opened 2025-06-13
Kab 1118 simplify config tool responses

ncapek:
This PR modifies the outputs of component configuration create/update tools to return minimalistic outputs to conserve token usage. Also adds links to the modified objects.


src/keboola_mcp_server/tools/components/model.py
```diff
@@ -1,14 +1,25 @@
...
 TransformationType = Literal['transformation']
 AllComponentTypes = Union[ComponentType, TransformationType]
 
 
+class ComponentToolResponse(BaseModel):
+    component_id: str = Field(..., description='The ID of the component.')
+    configuration_id: str = Field(..., description='The ID of the configuration.')
+    description: str = Field(..., description='The description of the configuration.')
+    timestamp: datetime = Field(..., description='The timestamp of the operation.')
+    success: bool = Field(default=True, description='Indicates if the operation succeeded.')
```
mariankrotil, line 19: Are we actually using `success = False` in the code? If not, is it meant as information for the agent, or can we remove it?

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -374,26 +384,33 @@ async def create_sql_transformation(
...
 
-    component = await _get_component(client=client, component_id=component_id)
-    new_transformation_configuration = ComponentConfigurationResponse.model_validate(
-        new_raw_transformation_configuration
-        | {
-            'component_id': component_id,
-            'component': component,
-        }
-    )
+    configuration_id = new_raw_transformation_configuration['id']
```
mariankrotil, line 387: We may use `new_transformation = ComponentConfigurationResponse.model_validate(new_raw_transformation_configuration)` to validate the result and work with a typed object `new_transformation.configuration_id` thereafter instead of the dict.

mariankrotil (approved):
Probably just a useless comment since you likely intended to gradually update main with the corresponding task id, but I noticed you are using the pattern below, while in [PR#165](https://github.com/keboola/mcp-server/pull/165) you are using `LinksManager.from_client` Which one should we be using? I would stick with the second option.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -380,26 +385,30 @@ async def create_sql_transformation(
...
 
-    LOG.info(
-        f'Created new transformation "{component_id}" with configuration id '
-        f'"{new_transformation_configuration.configuration_id}".'
+    LOG.info(f'Created new transformation "{component_id}" with configuration id ' f'"{configuration_id}".')
+
+    links = links_manager.get_component_configuration_links(
+        component_id=component_id,
+        configuration_id=configuration_id,
+        configuration_name=str(name),
```
vita-stejskal, line 401: Is `str(...)` really needed here?

src/keboola_mcp_server/tools/components/tools.py
vita-stejskal: OTOH it sounds a bit too heavy weight approach for just picking up the `configuration_id`.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -490,26 +502,32 @@ async def update_sql_transformation_configuration(
...
         component_id=sql_transformation_id,
         configuration_id=configuration_id,
-        configuration_version=updated_transformation_configuration.version,
+        configuration_version=updated_raw_configuration.get('version'),
+    )
+
+    links = links_manager.get_component_configuration_links(
+        component_id=sql_transformation_id,
+        configuration_id=str(configuration_id),
+        configuration_name=str(updated_raw_configuration.get('name', '')),
```
vita-stejskal, line 515: The same question here and below -- is the `str(...)` conversion really needed here?

tests/tools/components/test_tools.py
```diff
@@ -303,14 +303,10 @@ async def test_create_transformation_configuration(
...
-    )
-
-    assert isinstance(new_transformation_configuration, ComponentConfigurationResponse)
-    assert new_transformation_configuration.model_dump() == expected_config.model_dump()
-
-    keboola_client.ai_service_client.get_component_detail.assert_called_once_with(component_id=expected_component_id)
+    assert isinstance(new_transformation_configuration, ComponentToolResponse)
+    assert new_transformation_configuration.component_id == expected_component_id
+    assert new_transformation_configuration.configuration_id == mock_configuration['id']
+    assert new_transformation_configuration.description == mock_configuration['description']
```
vita-stejskal, line 309: There are other fields in `ComponentToolResponse` -- timestamp, success, links -- you should assert their values too. Specifically, the `timestamp` and `links` fields seem should have their values tested. Testing the `timestamp` is likely to require mocking `datetime.utcnow()` function though.

src/keboola_mcp_server/tools/components/model.py
vita-stejskal: The `success` field does look strange indeed.

AFAICT the former (not-simplified) tool responses did not contain it and responses from other (non-component / config related) tools do not contain it either. I'd remove it unless we are sure that it has positive effect on agents. In which case, I'd mention that effect in a comment next to the field.

vita-stejskal (approved):

tests/tools/components/test_tools.py
radcliffkey, line 1: please change the test names according to the functions being tested. sorry for the conflicts caused by the renaming :)

integtests/tools/components/test_tools.py
```diff
@@ -194,14 +237,29 @@ async def test_update_config(mcp_context: Context, configs: list[ConfigDef]):
...
         assert updated_config.component_id == component_id
         assert updated_config.configuration_id == created_config.configuration_id
-        assert updated_config.parameters == updated_parameters
-        assert updated_config.storage == updated_storage
-        assert updated_config.version > created_config.version
+        assert updated_config.description == updated_description
+        assert updated_config.success is True
+        assert updated_config.timestamp is not None
+
+        # Verify link structure
```
radcliffkey, line 248: let's keep the style used in other tests, i.e.
```
assert set(updated_config.links) == {
        Link(...),
        Link(...),
```

PR #167, opened 2025-06-13
Kab 1131 mcp writer storage validation

mariankrotil:
In this PR, we validate writer storage configuration give row-based and root-based setting:
- If row configurations are enabled, the root-level storage must be empty, but each row must have a non-empty input storage configuration.
- If row configurations are not enabled, a non-empty root-level input storage configuration is required.
We add tests as well.

Additionally, we simplify transformation storage validation and update tests.


src/keboola_mcp_server/tools/validation.py
```diff
@@ -267,6 +267,7 @@ def validate_storage_configuration(
     storage: Optional[JsonDict],
     component: Component,
     initial_message: Optional[str] = None,
+    validating_row_storage: bool = False,
```
radcliffkey, line 270: perhaps `validate_row_storage`? Pls add doctring.

PR #168, opened 2025-06-13
KAB-1129 mcp server doesn t forward exceptionids from sapi client

tomasfejfar:
This PR forwards the exceptionId for HTTP500 errors, so that we can later ask/instruct LLMs to forward the exceptionIds to support.

### DIFF
--- a/ERROR_HANDLING.md
+++ b/ERROR_HANDLING.md
@@ -1,0 +1,520 @@
+# Error Handling in Keboola MCP Server
+
+This document provides a comprehensive overview of how error handling is implemented throughout the Keboola MCP Server application. The application employs a multi-layered approach to error handling, ensuring robust operation, meaningful error messages, and recovery instructions for both developers and AI agents.
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Core Error Handling Architecture](#core-error-handling-architecture)
+3. [Tool-Level Error Handling](#tool-level-error-handling)
+4. [HTTP Client Error Handling](#http-client-error-handling)
+5. [Validation Error Handling](#validation-error-handling)
+6. [OAuth Error Handling](#oauth-error-handling)
+7. [Server-Level Error Handling](#server-level-error-handling)
+8. [Configuration Error Handling](#configuration-error-handling)
+9. [Logging and Monitoring](#logging-and-monitoring)
+10. [Error Recovery Patterns](#error-recovery-patterns)
+11. [Testing Error Handling](#testing-error-handling)
+
+## Overview
+
+The Keboola MCP Server implements a sophisticated error handling system designed to:
+
+- **Provide meaningful error messages** to AI agents with recovery instructions
+- **Log errors comprehensively** for debugging and monitoring
+- **Handle different types of failures** gracefully (network, validation, authentication, etc.)
+- **Maintain system stability** even when individual operations fail
+- **Support both human developers and AI agents** with appropriate error context
+
+## Core Error Handling Architecture
+
+### 1. Custom Exception Classes
+
+The application defines custom exception classes to provide structured error handling:
+
+#### `ToolException` Class
+```python
+class ToolException(Exception):
+    """Custom tool exception class that wraps tool execution errors."""
+
+    def __init__(self, original_exception: Exception, recovery_instruction: str):
+        super().__init__(f'{str(original_exception)} | Recovery: {recovery_instruction}')
+```
+
+**Purpose**: Wraps original exceptions with AI-friendly recovery instructions
+**Location**: `src/keboola_mcp_server/errors.py`
+
+#### `RecoverableValidationError` Class
+```python
+class RecoverableValidationError(jsonschema.ValidationError):
+    """An instance was invalid under a provided schema using a recoverable message for the Agent."""
+```
+
+**Purpose**: Provides detailed validation error messages with recovery instructions
+**Location**: `src/keboola_mcp_server/tools/validation.py`
+
+### 2. Decorator-Based Error Handling
+
+The application uses a decorator pattern for consistent error handling across all tools:
+
+#### `@tool_errors` Decorator
+```python
+def tool_errors(
+    default_recovery: Optional[str] = None,
+    recovery_instructions: Optional[dict[Type[Exception], str]] = None,
+) -> Callable[[F], F]:
+```
+
+**Features**:
+- **Automatic exception catching**: Wraps all tool functions in try-catch blocks
+- **Recovery instruction mapping**: Maps specific exception types to recovery messages
+- **Comprehensive logging**: Logs all exceptions with context
+- **Exception transformation**: Converts exceptions to `ToolException` with recovery instructions
+
+**Usage Example**:
+```python
+@tool_errors(
+    default_recovery="Please check your input parameters and try again.",
+    recovery_instructions={
+        ValueError: "Check that data has valid types.",
+        HTTPStatusError: "Verify your authentication credentials and API endpoint."
+    }
+)
+async def my_tool_function(ctx: Context, param: str) -> Result:
+    # Tool implementation
+    pass
+```
+
+## Tool-Level Error Handling
+
+### 1. Tool Function Error Handling
+
+All tool functions in the application are decorated with `@tool_errors` to ensure consistent error handling:
+
+**Components with tool error handling**:
+- Component tools (`src/keboola_mcp_server/tools/components/tools.py`)
+- Job tools (`src/keboola_mcp_server/tools/jobs.py`)
+- Storage tools (`src/keboola_mcp_server/tools/storage.py`)
+- Flow tools (`src/keboola_mcp_server/tools/flow.py`)
+- SQL tools (`src/keboola_mcp_server/tools/sql.py`)
+- Project tools (`src/keboola_mcp_server/tools/project.py`)
+- Documentation tools (`src/keboola_mcp_server/tools/doc.py`)
+
+### 2. Error Handling Flow in Tools
+
+1. **Exception occurs** in tool function
+2. **Logging**: Exception is logged with full context using `logging.exception()`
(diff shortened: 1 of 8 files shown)


src/keboola_mcp_server/client.py
```diff
@@ -120,6 +122,37 @@ def __init__(
...
+            original_exception = httpx.HTTPStatusError(
+                f"Server error '{response.status_code} {response.reason_phrase}' for url '{response.url}'",
+                request=response.request,
+                response=response
+            )
+            
+            raise KeboolaHTTPException(original_exception, exception_id, error_details)
+        else:
+            # For all other HTTP errors, use standard HTTPStatusError
+            response.raise_for_status()
```
odinuv, line 154: nevím no, asi bych šel už rovnou do Application+UserException konceptu? Pak by se dal udělat někde ten top level try-catch, ale tohle chce asi trošku probrat s nějakým python exception guru. Vím, že ten systém vyjímek nefunguje stejně v Pythonu, ale nevím jak

ERROR_HANDLING.md
```diff
@@ -0,0 +1,520 @@
+# Error Handling in Keboola MCP Server
```
mariankrotil, line 1: I like the idea of documenting our specific implemented approaches aiding in maintenance of the code for new programmers, however, both documents `.md` files are quite huge incorporating tons of texts resulting in info overload. Thus, we may consider shortening it and include only relevant and important information. If we insisted on the exclusive content, we should leverage moving them in a separate directory of documentations.

HTTP_500_ERROR_IMPROVEMENT_PLAN.md
```diff
@@ -0,0 +1,508 @@
+# HTTP 500 Error Handling Improvement Plan
```
mariankrotil, line 1: Do we really need to include the comprehensive plan of the improvement with all the phases and intermediate steps? If so, we may think of a dedicated directory for such files.

integtests/test_http_error_handling.py
```diff
@@ -0,0 +1,246 @@
...
+        """Test that Storage API 404 errors maintain standard behavior (no exception ID enhancement)."""
+        # Try to access a non-existent bucket
+        with pytest.raises(ToolException) as exc_info:
+            await get_bucket_detail("non.existent.bucket", mcp_context)
+        
+        tool_exception = exc_info.value
+        # Should not include exception ID for non-500 errors
+        assert "For support reference Exception ID:" not in tool_exception.recovery_message
+        # Should include standard recovery message
+        assert "Please try again later." in tool_exception.recovery_message
```
mariankrotil, line 37: Should we use a constant out of this `"Please try again later."` making it easier to change and experiment with it in a future development.

src/keboola_mcp_server/errors.py
```diff
@@ -36,15 +67,22 @@ async def wrapped(*args, **kwargs):
...
                 if recovery_instructions:
                     for exc_type, msg in recovery_instructions.items():
                         if isinstance(e, exc_type):
                             recovery_msg = msg
                             break
+                
+                # Special handling for KeboolaHTTPException (HTTP 500 errors only)
+                if isinstance(e, KeboolaHTTPException) and e.original_exception.response.status_code == 500:
+                    if e.exception_id:
+                        recovery_msg = f"{recovery_msg or 'Please try again later.'} For support reference Exception ID: {e.exception_id}"
```
mariankrotil, line 81: We may use a `default_recovery_message` for "Please try again later".  And, please, split the recovery instruction message from the support exception id message by a `newline`.

Additionally, we should discuss whether we want to add the `"For support reference Exception ID: {e.exception_id}"` after or before the recovery message as it is for the bot decision.

src/keboola_mcp_server/client.py
vita-stejskal: @odinuv Could you please give us a bit more details about the "Application+UserException koncept"? I assume that it's a different way how to handle 4xx and 5xx errors. But what is the difference?

src/keboola_mcp_server/errors.py
```diff
@@ -2,15 +2,46 @@
...
 from typing import Any, Callable, Optional, Type, TypeVar, cast
 
+import httpx
+
 LOG = logging.getLogger(__name__)
 
 F = TypeVar('F', bound=Callable[..., Any])
 
 
+class KeboolaHTTPException(httpx.HTTPStatusError):
```
vita-stejskal, line 12: Umm, I could be wrong, but this all seems unnecessarily complicated. You pass a lot of stuff around just to create the exception message, which you partially create in `RawKeboolaClient._handle_http_error()` function anyway. Why don't you simply create the full message there and just raise `httpx.HTTPStatusError`.

src/keboola_mcp_server/errors.py
vita-stejskal, line 81: Please revert all the changes in `tool_errors()` wrapper and add `"For support reference Exception ID: {e.exception_id}"` to the message created for the `httpx.HTTPStatusError`.

I'd also rephrase the text a little bit to `"When contacting Keboola support please refer to the exception ID: {e.exception_id}."`

The `Please try again later.` instruction is IMO generally pretty useless and often can be misleading. Many errors are a result of a wrong LLM input fed to a tool and it makes no sense to "try later", because it won't help unless the LLM corrects the input that it feeds to the tool.

tests/test_errors.py
```diff
@@ -1,10 +1,492 @@
...
+
+# --- Test Cases for KeboolaHTTPException ---
+
+
+class TestKeboolaHTTPException:
+    """Test suite for KeboolaHTTPException class."""
+
+    def test_create_with_exception_id_for_500_error(self, mock_httpstatus_error_500):
+        """Test that KeboolaHTTPException includes exception ID for HTTP 500 errors."""
+        from keboola_mcp_server.errors import KeboolaHTTPException
```
mariankrotil, line 78: We should move these imports to the top (the same for other below)

tests/test_errors.py
```diff
@@ -1,10 +1,492 @@
...
 
+import httpx
 import pytest
 
 from keboola_mcp_server.errors import ToolException, tool_errors
 
 
+# --- New test fixtures for KeboolaHTTPException ---
+@pytest.fixture
+def mock_http_request():
```
vita-stejskal, line 12: Please use typing hints for the return values, e.g. `-> httpx.Request:`.

src/keboola_mcp_server/errors.py
```diff
@@ -2,15 +2,46 @@
...
+        message = self._build_error_message()
+        super().__init__(message, request=original_exception.request, response=original_exception.response)
+    
+    def _build_error_message(self) -> str:
+        """Build a comprehensive error message including exceptionId for HTTP 500 errors."""
+        base_message = str(self.original_exception)
+        
+        # Only enhance HTTP 500 errors with exception ID
+        if self.original_exception.response.status_code == 500 and self.exception_id:
+            base_message += f" (Exception ID: {self.exception_id})"
```
mariankrotil, line 30: consider dividing the individual lines using newline

src/keboola_mcp_server/client.py
odinuv: The concept is that application exception is 5xx and doesn't display much information (because the user can't do much about it either), in the sense of application - if there is end-user fixable error we convert it to something inheriting from UserException and then there is top-level try-catch which converts that to 4xx response and everything else to 5xx.

The thing is that it's not just blind conversion of 4xx client errors to 4xx responses, an example would be that e.g. the workspace that is created in the background when a user initiates the session may, for some reason return 4xx error, but since it's happening in the background with no user input, the user can't fix it, thus it should be applicationError with 5xx

tomasfejfar:
> @odinuv Could you please give us a bit more details about the "Application+UserException koncept"? I assume that it's a different way how to handle 4xx and 5xx errors. But what is the difference?

The other exception is one which has a message that can be directly passed to the user. It's true at places where we are sure that the exception might appear and its message is safe to display. User exception is very likely to be caused by user action or users mistake. 

 in contrast an application exception is thrown when an unexpected thing happens. Any exception that is not caught and rethrown is usually an application exception. 

So you are right but it is different concept for HTTP statuses. To difference being that you can throw meaningful exception in your code, that is not http exception and layers above can decide what to do with it. 

For example code throwing user exception will be handled as 400 in API and in background job it will result in meaningful job error message. While application exception will be 500 in API and "job failed with execution abcdefgh, contact support".

tomasfejfar:
I'll try to fix your comments.

src/keboola_mcp_server/client.py
```diff
@@ -120,6 +122,37 @@ def __init__(
...
+                # Fallback to basic error handling
+                exception_id = None
+                error_details = None
+            
+            # Create enhanced exception for HTTP 500 errors
+            original_exception = httpx.HTTPStatusError(
+                f"Server error '{response.status_code} {response.reason_phrase}' for url '{response.url}'",
+                request=response.request,
+                response=response
+            )
```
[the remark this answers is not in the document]
tomasfejfar: I think I resolved this

src/keboola_mcp_server/errors.py
tomasfejfar: I removed the recovery message as well if not provided.

tests/test_errors.py
tomasfejfar: Resolved

vita-stejskal:
@radcliffkey @mariankrotil @ncapek I heavily cleaned the changes in this PR. All unit and integration tests should pass now. It's ready to be reviewed. Could you please give it a look. Thanks

HTTP_500_ERROR_IMPROVEMENT_PLAN.md
vita-stejskal: Removed in afe613e.

integtests/test_http_error_handling.py
vita-stejskal: Cleaned up in 19ef907. The message is no longer in use.

PR #171, opened 2025-06-19
allow running github actions in ci for pull_requests

mariankrotil:
We enable running github actions on pull requests in hte ci pipeline to allow it on PRs of forks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,6 @@
 name: CI
 
-on:
-  push:
-  pull_request:
-
+on: [ push, pull_request ]
 concurrency: ci-${{ github.ref }}
 
 permissions:


.github/workflows/ci.yml
```diff
@@ -1,6 +1,9 @@
 name: CI
 
-on: [ push ]
+on:
+  push:
```
radcliffkey, line 4: should the `:` be there?

mariankrotil: Not sure, but after `:` we specify additional properties for the given event (e.g. specific branch prefixes and so on). So I think leaving it empty results in running actions every push and pull request? @odinuv

odinuv: yes, it's either 
`on: [push, pull_request]` 
or 
```
on:
  push:
  pull_request:
```

PR #174, opened 2025-06-19
Add environment variable configuration support

janvanicek:
This PR adds support for configuring the Keboola MCP server using environment variables instead of command line parameters.

## Changes
- Added support for `KBC_STORAGE_API_URL` environment variable
- Updated CLI to read from environment variables first, with command line params taking precedence
- Updated README with environment variable configuration examples for both Claude Desktop and Cursor AI
- Modified Dockerfile to use environment variable approach

## Reasons
- **Docker MCP Toolkit compatibility** - the toolkit only supports environment variables, not command line parameters
- Command line parameters still work for backward compatibility


README.md
```diff
@@ -118,6 +123,23 @@ In this mode, Claude or Cursor automatically starts the MCP server for you. **Yo
 }
 ```
 
+**Option 2: Using environment variable**
+```json
+{
+  "mcpServers": {
+    "keboola": {
+      "command": "uvx",
+      "args": ["keboola_mcp_server"],
```
mariankrotil, line 132: what about using only `"command": "uvx keboola_mcp_server"`

README.md
```diff
@@ -118,6 +123,23 @@ In this mode, Claude or Cursor automatically starts the MCP server for you. **Yo
 }
 ```
 
+**Option 2: Using environment variable**
```
mariankrotil, line 126: We should consider using only one and the simpler option in the `readme`, as offering two choices may confuse or discourage the user. I like and would use the option you added if it works properly. The same holds for choices below.

README.md
vita-stejskal: @mariankrotil That may or may not work based on how the `command` setting is interpreted / used by the particular AI assistant apps.

README.md
vita-stejskal: I agree with @mariankrotil . I'd simply remove the `--api-url` CLI parameter from the examples and replaced it with `KBC_STORAGE_API_URL` environment variable.

vita-stejskal: Fixed in 44e1755.

PR #176, opened 2025-06-20
KAB-1171 Skip Actions for Forks

mariankrotil:
skip actions requiring secrets, run job of integration tests only for PRs of non-forks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,13 +42,16 @@ jobs:
 
       - name: Publish test results
         uses: dorny/test-reporter@v1
-        if: success() || failure() # run this step even if a previous step failed
+        # run this step even if a previous step failed, but only for push events or non-fork pull requests
+        if: (success() || failure()) && (github.event_name == 'push' || github.event.pull_request.head.repo.fork == false)
         with:
           name: Test results (${{ matrix.python-version }})
           path: ./test-results.xml
           reporter: 'java-junit'
 
       - name: Upload coverage to Codecov
+        # run this step only for push events or non-fork pull requests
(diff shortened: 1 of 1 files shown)


Matovidlo (approved):
LGTM let's try it in worst case well fix again :slightly_smiling_face:

.github/workflows/ci.yml
```diff
@@ -10,6 +10,8 @@ permissions:
 jobs:
   build:
     name: Build, test and package
+    # run this job only for push (avoiding pull requests from the same repo) or for pull requests from diff repo
```
radcliffkey, line 13:
```suggestion
    # run this job only for push (avoiding pull requests from the same repo) or for pull requests from different repo
```

radcliffkey (commented):
There should be a comment explaining **why** some events are run `only for push events or pull requests from the same repo`, either at each event or on a single place above.
IIUC it would say something like `... because this action requires secrets which the other repos could not access`

.github/workflows/ci.yml
mariankrotil: done

.github/workflows/ci.yml
```diff
@@ -42,13 +44,16 @@ jobs:
 
       - name: Publish test results
         uses: dorny/test-reporter@v1
-        if: success() || failure() # run this step even if a previous step failed
+        # run this step even if a previous step failed, but only for push events or pull requests from the same repo
+        if: (success() || failure()) && (github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository)
```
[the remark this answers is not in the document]
mariankrotil: I am not sure. Maybe it is necessary to explicitly state that the action runs when the previous step fails, and thus, also when succeeds.

PR #186, opened 2025-06-26
Ai 1157 sql transformation create table

mariankrotil:
We updated the description of the create_sql_transformation tool to include a instroction:
```- By default, SQL transformation must create at least one table to produce a result; omit only if the user
      explicitly indicates that no table creation is needed.```


See the result. However, the agent often creates semantically incorrect SQL queries but which fixes after the job failure.
![Screenshot 2025-06-26 161718](https://github.com/user-attachments/assets/d7b99a96-9494-44a3-8b81-b9986f640193)

### DIFF
--- a/src/keboola_mcp_server/tools/components/tools.py
+++ b/src/keboola_mcp_server/tools/components/tools.py
@@ -321,18 +321,19 @@ async def create_sql_transformation(
         Sequence[str],
         Field(
             description=(
-                'An empty list or a list of created table names if and only if they are generated within SQL '
-                'statements (e.g., using `CREATE TABLE ...`).'
+                'A list of created table names if they are generated within the SQL query statements '
+                '(e.g., using `CREATE TABLE ...`).'
             ),
         ),
     ] = tuple(),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/keboola_mcp_server/tools/components/tools.py [hunk 1]
```diff
@@ -311,32 +311,32 @@ async def create_sql_transformation(
         Sequence[TransformationConfiguration.Parameters.Block.Code],
         Field(
             description=(
-                'The executable SQL query code blocks, each containing a descriptive name and a sequence of '
-                'semantically related sql statements written in the current SQL dialect. Each sql statement is'
-                'executable and a separate item in the list of sql statements.'
+                'The SQL query code blocks, each containing a descriptive name and a sequence of '
+                'semantically related inependently executable sql_statements written in the current SQL dialect.'
```
radcliffkey, line 315:
```suggestion
                'semantically related independently executable SQL statements written in the current SQL dialect.'
```

For Cursor, I recommend _Code Spell Checker_ extension

radcliffkey (approved):

PR #187, opened 2025-06-29
Ai 131 integ tests

ncapek:
Add more integtests to increase coverage


integtests/tools/test_jobs.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@pytest.mark.asyncio
+async def test_get_job(mcp_context: Context):
+    """Tests`get_job`."""
+
+    # First get a list of jobs to find one to examine in detail
+    jobs_list = await list_jobs(ctx=mcp_context, limit=5)
+
+    if len(jobs_list.jobs) > 0:
```
mariankrotil, line 127: If there are not any jobs in the project, we also pass this test, which is not good. In this test method, we should expect that there are jobs in our testing project (maybe we should add the jobs to the project in the initialization of integration tests as it is done with e.g. configs). Otherwise, we could change functionality of that tool but we would not notice.

integtests/tools/test_sql.py
```diff
@@ -0,0 +1,98 @@
...
+    """Tests basic functionality of SQL tools: get_sql_dialect and query_data."""
+
+    dialect = await get_sql_dialect(ctx=mcp_context)
+    assert dialect in ['Snowflake', 'Bigquery']
+
+    buckets_listing = await list_buckets(ctx=mcp_context)
+
+    tables_listing = await list_tables(bucket_id=buckets_listing.buckets[0].id, ctx=mcp_context)
+    table = await get_table(table_id=tables_listing.tables[0].id, ctx=mcp_context)
+    LOG.error(table)
```
mariankrotil, line 25: maybe leftover?

ncapek: Yes. removed.

integtests/tools/components/test_tools.py
```diff
@@ -863,3 +880,118 @@ async def test_update_sql_transformation(mcp_context: Context, keboola_project:
...
+        assert isinstance(component, SuggestedComponent)
+
+
+@pytest.mark.asyncio
+async def test_get_config_examples_with_invalid_component(mcp_context: Context):
+    """Tests that `get_config_examples` handles non-existent components properly."""
+
+    result = await get_config_examples(ctx=mcp_context, component_id='completely-non-existent-component-12345')
+
+    assert result == ''
```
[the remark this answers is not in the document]
ncapek: cc @vita-stejskal

integtests/tools/test_jobs.py
```diff
@@ -0,0 +1,278 @@
...
+from integtests.conftest import ConfigDef, ProjectDef
+from keboola_mcp_server.client import KeboolaClient
+from keboola_mcp_server.links import Link
+from keboola_mcp_server.tools.components.tools import create_config
+from keboola_mcp_server.tools.jobs import JobDetail, ListJobsOutput, get_job, list_jobs, run_job
+
+LOG = logging.getLogger(__name__)
+
+
+async def _wait_for_job_in_list(
```
mariankrotil, line 16: Would not be enough to use only `sleep` with a sufficient number of seconds?

Also there are third-party libraries that perform this `retry` logic using only annotations - `retry` or `tenacity` but it might be too much just for tests.

mariankrotil (approved):

integtests/tools/test_jobs.py
```diff
@@ -0,0 +1,278 @@
...
+        result = await list_jobs(
+            ctx=mcp_context,
+            component_id=component_id,
+            config_id=config_id,
+            limit=10,
+            sort_by='startTime',
+            sort_order='desc',
+        )
+
+        job_ids = [job.id for job in result.jobs]
```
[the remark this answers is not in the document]
ncapek: Done

integtests/tools/test_sql.py
```diff
@@ -0,0 +1,98 @@
...
+
+    invalid_sql = 'INVALID SQL SYNTAX SELECT * FROM'
+
+    with pytest.raises(ValueError, match='Failed to run SQL query'):
+        await query_data(sql_query=invalid_sql, ctx=mcp_context)
+
+
+@pytest.mark.asyncio
+async def test_query_data_non_select_query(mcp_context: Context):
+    """Tests that `query_data` handles non-SELECT queries and returns a message."""
```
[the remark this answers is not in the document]
ncapek: So this test should simply be removed?

PR #189, opened 2025-07-02
AI-331: use gzip to compress the authorization code, access and refresh tokens

vita-stejskal:
The authorization code, access and refresh tokens issued by the MCP server are still encrypted JSON Web Tokens, but their contents is `gzip`-ed to make the tokens shorter. The `gzip`-ed tokens are approximately 60% of their original size.

Apparently, even the modern Windows versions still have a limitation on the size of the command used for launching a process. This affects the max size of the redirect URI used for passing the authorization code back to the running CursorAI process. CursorAI uses `cursor://` protocol in its redirect URIs and the browsers handle that by starting another CursorAI process and passing it the redirect URL as a command-line parameter. The Windows OSes require this whole command to fit into a certain length (max chars).

Since the authorization code generated by our MCP server needs to contain all the information need later on to swap the authorization code for the access/refresh tokens it is a bit lengthy. The OAuth protocol itself allows that and does not restrict the tokens size in any way. But there are practical limits such as the max command-line size on Windows.

This could be of interest to @odinuv @tomasfejfar @romanbracinik .

### DIFF
--- a/src/keboola_mcp_server/oauth.py
+++ b/src/keboola_mcp_server/oauth.py
@@ -1,13 +1,15 @@
+import gzip
+import json
 import logging
 import math
 import secrets
 import time
 from http.client import HTTPException
-from typing import Any, cast
+from typing import Any, Mapping, cast
 from urllib.parse import urljoin
 
 import httpx
-import jwt
+import jwt.api_jws
 from fastmcp.server.auth.auth import OAuthProvider
 from mcp.server.auth.provider import (
     AccessToken,
@@ -154,7 +156,7 @@ class SimpleOAuthProvider(OAuthProvider):
             'client_id': client.client_id,
             'expires_at': time.time() + 5 * 60,  # 5 minutes from now
         }
-        state_jwt = jwt.encode(state, self._jwt_secret)
+        state_jwt = self._encode(state)
 
         # create the authorization URL
         url_params = {
@@ -181,7 +183,7 @@ class SimpleOAuthProvider(OAuthProvider):
         """
         # Validate the state first to prevent calling OAuth server with invalid authorization code.
         try:
-            state_data = jwt.decode(state, self._jwt_secret, algorithms=['HS256'])
+            state_data = self._decode(state)
         except jwt.InvalidTokenError:
             LOG.debug(f'[handle_oauth_callback] Invalid state: {state}', exc_info=True)
             raise HTTPException(400, 'Invalid state parameter')
@@ -240,7 +242,7 @@ class SimpleOAuthProvider(OAuthProvider):
             'oauth_access_token': access_token.model_dump(),
             'oauth_refresh_token': refresh_token.model_dump(),
         }
-        auth_code_jwt = jwt.encode(auth_code, self._jwt_secret)
+        auth_code_jwt = self._encode(auth_code)
 
         mcp_redirect_uri = construct_redirect_uri(
             redirect_uri_base=redirect_uri, code=auth_code_jwt, state=state_data['state'],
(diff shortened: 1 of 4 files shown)


src/keboola_mcp_server/oauth.py
```diff
@@ -571,3 +573,19 @@ def _ceil_to_hour(seconds: int) -> int:
...
+        return json_encrypted
+        # return jwt.encode(data, key or self._jwt_secret)
+
+    def _decode(self, data: str, *, key: str | None = None) -> dict[str, Any]:
+        json_gzip = jwt.api_jws.decode(data, key or self._jwt_secret, algorithms=['HS256'])
+        json_bytes = gzip.decompress(json_gzip)
+        json_str = json_bytes.decode('utf-8')
+        json_obj = json.loads(json_str)
+        return json_obj
+        # return jwt.decode(data, key or self._jwt_secret, algorithms=['HS256'])
```
mariankrotil, line 591: Perhaps leftoverr

tests/test_oauth.py
```diff
@@ -62,6 +61,7 @@ async def test_load_authorization_code(
             oauth_provider: SimpleOAuthProvider
     ):
         client_info = OAuthClientInformationFull(client_id='foo-client-id', redirect_uris=[AnyUrl('foo://bar')])
-        auth_code_str = jwt.encode(auth_code, key=key)
+        auth_code_str = oauth_provider._encode(auth_code, key=key)
+        print(f'len(auth_code_str)={len(auth_code_str)}')
```
mariankrotil, line 65: Remove the debug print

mariankrotil (approved):
Not sure if it is tested tested, but we could add a test assessing `_encode` and `_decode` methods that would compare the dictionaries: one before the process with one after the process.

src/keboola_mcp_server/oauth.py
vita-stejskal: Removed in cd82fa3.

tests/test_oauth.py
vita-stejskal: Removed in cd82fa3.

vita-stejskal:
> Not sure if it is tested tested, but we could add a test assessing `_encode` and `_decode` methods that would compare the dictionaries: one before the process with one after the process.

The encrypting and decrypting is tested by `TestSimpleOAuthProvider.test_load_authorization_code()` function.

src/keboola_mcp_server/oauth.py
```diff
@@ -527,8 +529,13 @@ def _read_oauth_tokens(self, data: dict[str, Any], scopes: list[str]) -> tuple[A
...
             scopes=scopes,
-            # this is slightly different from 'expires_at' kept by the OAuth server
-            expires_at=current_time + expires_in,
+            # The expires_in refers to the access token.
+            # There is no way of knowing when the refresh token expires.
+            # The Keboola OAuth server issues refresh tokens that expire in 1 month and access tokens that
+            # expire in 1 hour.
+            # We derive the lifespan of a refresh token from the lifespan of an access token and make it approximately
+            # 1 week long under the default circumstances.
+            expires_at=current_time + self._ceil_to_hour(min(168 * expires_in, 168 * 3600)),
```
mariankrotil, line 538: If the `expires_in` is `0` then the both tokens expire immediately, is it intended?

mariankrotil: I just found that we handle such problem a few lines above.

mariankrotil:
> @mariankrotil I've sneaked in a fix for the lifespan of refresh tokens. Please see https://github.com/keboola/mcp-server/commit/075ac8fabd677fbecd10c75d0a2b376956db47e7. I'm going to deploy that version to canary-orion for testing.

Looks reasonable.

PR #190, opened 2025-07-02
Ai 1172 Add Global Search Method To The StorageClient

mariankrotil:
Add global search client method for the global search endpoint.
Add client method for checking if project has global storage enabled.

### DIFF
--- a/src/keboola_mcp_server/client.py
+++ b/src/keboola_mcp_server/client.py
@@ -3,7 +3,7 @@
 import importlib.metadata
 import logging
 import os
-from typing import Any, Literal, Mapping, Optional, Union, cast
+from typing import Any, Iterable, Literal, Mapping, Optional, Union, cast
 
 import httpx
 from pydantic import BaseModel, Field
@@ -17,8 +17,8 @@ JsonStruct = Union[JsonDict, JsonList]
 
 ComponentResource = Literal['configuration', 'rows', 'state']
 
-# Feature flag for the global search if it is enabled in the project
-GlobalSearchFeatureFlag = Literal['global-search']
+# Project features that can be checked with the is_enabled method
+ProjectFeature = Literal['global-search']
 # Input types for the global search endpoint parameters
 GlobalSearchBranchTypes = Literal['production', 'development']
 GlobalSearchTypes = Literal[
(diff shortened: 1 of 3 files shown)


src/keboola_mcp_server/client.py
```diff
@@ -937,6 +985,16 @@ async def project_id(self) -> str:
         raw_data = cast(JsonDict, await self.get(endpoint='tokens/verify'))
         return str(raw_data['owner']['id'])
 
+    async def has_global_search(self) -> bool:
```
vita-stejskal, line 988: This could probably be a bit more generic and declared as:
```
async def is_enabled(self, features: str | Iterable[str]) -> bool:
    """
    Checks if a feature or a set of features is enabled in the active project.
    :params features: The features to check.
    :returns: True if all the features are enabled, False otherwise.
    """
```

Also the `features` parameter could probably by of some `enum` type rather than just plain string.
```
class ProjectFeatures(Enum, str):
    GLOBAL_SEARCH = 'global-search'
```

If you find using the `enum` too restrictive then please use a `Literal` type alias.
```
ProjectFeatures = Literal[
    'global-search'
]
```
We are soon going to need to add "WAII" feature. I don't remember its exact string value though.

Also, please, write a unit test for this function. It should be fairly simple.

vita-stejskal (approved):
Please increment the project version.

src/keboola_mcp_server/client.py
ncapek: We had a similar discussion with Rado regarding `MetadataField`. Ideally we could use `StrEnum` but that's not yet supported in python 3.10, so we went with a normal class intead.

src/keboola_mcp_server/client.py
```diff
@@ -819,6 +828,39 @@ async def flow_update(
             updated_description=description,
         )
 
+    async def global_search(
+        self,
+        query: str,
+        limit: int = 100,
+        offset: int = 0,
+        types: list[GlobalSearchTypes] | None = None,
+        branch_ids: list[str] | None = None
```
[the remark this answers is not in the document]
mariankrotil: Removed.

src/keboola_mcp_server/client.py
mariankrotil: Done. Using `is_enabled` method and only:
```
ProjectFeatures = Literal[
    'global-search'
]
```

PR #196, opened 2025-07-05
add data type information to get_table tool

ncapek:
Add column data type to `get_table` tool

### DIFF
--- a/src/keboola_mcp_server/tools/storage.py
+++ b/src/keboola_mcp_server/tools/storage.py
@@ -103,6 +103,9 @@ class TableColumnInfo(BaseModel):
         validation_alias=AliasChoices('quotedName', 'quoted_name', 'quoted-name'),
         serialization_alias='quotedName',
     )
+    data_type: Optional[str] = Field(None, description='Data type of the column.')
+    base_type: Optional[str] = Field(None, description='Base data type of the column.')
+    nullable: Optional[bool] = Field(None, description='Whether the column can contain null values.')
 
 
 class TableDetail(BaseModel):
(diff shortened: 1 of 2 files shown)


ncapek:
Currently there are some issues with workspaces and table tools due to the patch from yesterday, hence integration tests are failing at the moment.

src/keboola_mcp_server/tools/storage.py
```diff
@@ -212,12 +215,42 @@ async def get_table(
 
     raw_table = await client.storage_client.table_detail(table_id)
     raw_columns = cast(list[str], raw_table.get('columns', []))
-    column_info = [
-        TableColumnInfo(name=col, quoted_name=await workspace_manager.get_quoted_name(col)) for col in raw_columns
-    ]
+    column_metadata = cast(dict[str, list[dict[str, Any]]], raw_table.get('columnMetadata', {}))
+
+    def extract_column_metadata(column_name: str) -> dict[str, Any]:
```
mariankrotil, line 220: I would not create embedded helper function within a tool function, but rather, define a regular helper function in this `storage` module

src/keboola_mcp_server/tools/storage.py
```diff
@@ -103,6 +103,9 @@ class TableColumnInfo(BaseModel):
         validation_alias=AliasChoices('quotedName', 'quoted_name', 'quoted-name'),
         serialization_alias='quotedName',
     )
+    data_type: Optional[str] = Field(None, description='Data type of the column.')
+    base_type: Optional[str] = Field(None, description='Base data type of the column.')
+    nullable: Optional[bool] = Field(None, description='Whether the column can contain null values.')
```
mariankrotil, line 108: What if it is a foreign key? Is it possible record this property as well? I think it might be helpful.

mariankrotil (approved):

src/keboola_mcp_server/tools/storage.py
radcliffkey, line 108: I'm not a big fan of `Optional[bool]` because now you have a third state you have to handle. Perhaps this should be `True` by default.

vita-stejskal: I afraid, that the foreign key relations among the tables are not captured in the Keboola artifacts.

src/keboola_mcp_server/tools/storage.py
```diff
@@ -23,30 +23,42 @@
...
-    mcp.add_tool(FunctionTool.from_function(list_buckets, serializer=listing_output_serializer))
-    mcp.add_tool(FunctionTool.from_function(get_table))
-    mcp.add_tool(FunctionTool.from_function(list_tables, serializer=listing_output_serializer))
+    mcp.add_tool(FunctionTool.from_function(list_buckets, serializer=exclude_none_serializer))
+    mcp.add_tool(FunctionTool.from_function(get_table, serializer=exclude_none_serializer))
+    mcp.add_tool(FunctionTool.from_function(list_tables, serializer=exclude_none_serializer))
     mcp.add_tool(FunctionTool.from_function(update_bucket_description))
-    mcp.add_tool(FunctionTool.from_function(update_table_description, serializer=listing_output_serializer))
-    mcp.add_tool(FunctionTool.from_function(update_column_description, serializer=listing_output_serializer))
+    mcp.add_tool(FunctionTool.from_function(update_table_description, serializer=exclude_none_serializer))
```
ncapek, line 30: These tools have very structured output with few fields. They probably don't need to serializer.

src/keboola_mcp_server/tools/storage.py
vita-stejskal: Fixed in 7b6d58f.

src/keboola_mcp_server/tools/storage.py
```diff
@@ -212,12 +215,42 @@ async def get_table(
...
+
+        return result
+
+    column_info = []
+    for col in raw_columns:
+        metadata = extract_column_metadata(col)
+        column_info.append(TableColumnInfo(
+            name=col,
+            quoted_name=await workspace_manager.get_quoted_name(col),
+            data_type=metadata.get('data_type'),
```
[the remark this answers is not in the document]
vita-stejskal: The default values for `native_type` and `nullable` added in df0ddff.

PR #197, opened 2025-07-05
update description in query_data tool

ncapek:
Updating the description of the `query_data` tool to fix issues of agent using made up values of categorical columns.


ncapek:
Parking for now until issues with workspaces and tables get resolved.

PR #199, opened 2025-07-07
Ai 1172 Typed response for global search client method

mariankrotil:
Make typed response in the global search client method 
Add integ tests for global search.

### DIFF
--- a/src/keboola_mcp_server/client.py
+++ b/src/keboola_mcp_server/client.py
@@ -3,6 +3,7 @@
 import importlib.metadata
 import logging
 import os
+from datetime import datetime
 from typing import Any, Iterable, Literal, Mapping, Optional, Union, cast
 
 import httpx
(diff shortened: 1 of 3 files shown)


radcliffkey:
General note: please write more meaningful PR titles, don't just leave the pre-filled one. This could be something like _Typed response for global search client method_

src/keboola_mcp_server/client.py
```diff
@@ -355,6 +355,27 @@ async def delete(
         return await self.raw_client.delete(endpoint=endpoint)
 
 
+class GlobalSearchResponse(BaseModel):
+    """The SAPI global search response."""
+
+    all: int = Field(description='Total number of found results.')
+    items: list[dict[str, Any]] = Field(
```
radcliffkey, line 362: could this be typed as well?

src/keboola_mcp_server/client.py
```diff
@@ -355,6 +355,27 @@ async def delete(
...
+        description='List of search results containing the items of the GlobalSearchType.'
+    )
+    by_type: dict[str, JsonPrimitive] = Field(description='Search results grouped by type.', alias='byType')
```
radcliffkey, line 365: why `JsonPrimitive`? Please make it more clear in the description what the key is and what the value is.

src/keboola_mcp_server/client.py
```diff
@@ -355,6 +355,27 @@ async def delete(
...
+    by_project: dict[str, JsonPrimitive] = Field(
```
radcliffkey, line 366: again, why `JsonPrimitive`? Isn't project name a string?

src/keboola_mcp_server/client.py
mariankrotil: done here: 37084b937faa7a0be778c54e9f900fc2c780d5e0

src/keboola_mcp_server/client.py
mariankrotil: done here 37084b937faa7a0be778c54e9f900fc2c780d5e0
using only `str`

PR #200, opened 2025-07-07
feat: Introduce automatic versioning for tagged workflows

Matovidlo:
* Current limitation is that `tagged` versions have changed version in `pyproject.toml`.
* The main branch have to have manually changed version

### DIFF
--- a/docs/VERSION_MANAGEMENT.md
+++ b/docs/VERSION_MANAGEMENT.md
@@ -1,0 +1,69 @@
+# Version Management
+
+This project automatically updates the version in `pyproject.toml` when a new git tag is created.
+
+## How It Works
+
+When you create a git tag with the format `v1.2.3`, the GitHub Actions workflow will:
+
+1. Extract the version number from the tag (removing the `v` prefix)
+2. Update the `version` field in `pyproject.toml`
+3. Commit and push the changes back to the repository
+4. Continue with the Docker build and push process
+
+## Usage
+
+### Creating a New Release
+
+1. **Create and push a new tag:**
+   ```bash
+   git tag v1.2.3
+   git push origin v1.2.3
+   ```
+
+2. **The workflow will automatically:**
+   - Update `pyproject.toml` version to `1.2.3`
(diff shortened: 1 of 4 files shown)


vita-stejskal:
I am afraid that changing `pyproject.toml` is not the most efficient way how to automate / simplify versioning in python projects. I think that we should look at [setuptools-git-versioning](https://github.com/dolfinus/setuptools-git-versioning) and try to use that. However, no matter what version automation system we use, we should carefully consider its implications for the project and the way we work.

docs/VERSION_MANAGEMENT.md
```diff
@@ -0,0 +1,69 @@
+# Version Management
+
+This project automatically updates the version in `pyproject.toml` when a new git tag is created.
+
+## How It Works
+
+When you create a git tag with the format `v1.2.3`, the GitHub Actions workflow will:
```
mariankrotil, line 7: I like it. Just a question. We are used to triggering a release to the pypi by tagging our main branch. Wouldn't be there then a conflict with a regular PR with version update when we do not want to release but only increase the version?

Matovidlo: I am not sure, I kicked it off just to show how I see how tagging for regular workflow could work.
We should fix
1. Pypi versioning
2. Check vita's comment which does not like this approach.
If we are on a same page I would continue on it, this is low prio task. For me the integrations tests are much bigger issue but let's have some discussion out of standup

vita-stejskal:
I am sorry for being a bit assertive now, but there is a lot of uncertainty and controversy around this change and I am going to close this PR. The reasons are:

* The current system with manually updating the project version `pyproject.toml` file works fine. There have been no serious complaints about it.
* It's unclear how the semantic versioning would work when not done manually in `pyproject.toml`.
* We have more pressing issues to sort out.
* This is not meant to kill any further discussion or changes in this area, but it needs to be thought through carefully and discussed thoroughly. It's much more of a project/release management problem than a technical problem. The discussions leading to a consensus will be much more demanding in this case than the technical solution which is likely to be fairly easy once we decide how the project versioning should work.

PR #202, opened 2025-07-09
AI 1172 Add Global Search Tool

mariankrotil:
Add global search tool accepting name-based query, types to search for and parameter for limiting the number of returned examples and for pagination.


integtests/test_client.py
```diff
@@ -36,5 +36,5 @@ async def test_global_search_with_results(self, storage_client: AsyncStorageClie
         assert isinstance(ret, GlobalSearchResponse)
         assert ret.all == len(tables)
         assert len(ret.items) == len(tables)
-        assert all(isinstance(item, GlobalSearchResponse.GlobalSearchResponseItem) for item in ret.items)
+        assert all(isinstance(item, GlobalSearchResponse.Item) for item in ret.items)
```
ncapek, line 39: This is already covered by `assert isinstance(ret, GlobalSearchResponse)`

src/keboola_mcp_server/tools/doc.py
```diff
@@ -31,6 +33,53 @@ class DocsAnswer(BaseModel):
     source_urls: list[str] = Field(description='List of URLs to the sources of the answer.')
 
 
+class GlobalSearchAnswer(BaseModel):
```
ncapek, line 36: Maybe call this `GlobalSearchResult`?

src/keboola_mcp_server/tools/doc.py
```diff
@@ -31,6 +33,53 @@ class DocsAnswer(BaseModel):
...
+        created: datetime = Field(description='The date and time the entity was created.')
+        additional_info: dict[str, Any] = Field(description='Additional information about the item.')
+
+        @classmethod
+        def from_api_response(cls, item: GlobalSearchResponse.Item) -> 'GlobalSearchAnswer.Item':
+            """Creates an Item from the API response."""
+            add_info = {}
+            if item.type == 'table':
+                bucket_info = item.full_path.get('bucket', {})
+                add_info['bucket_id'] = bucket_info.get('id', 'unknown')
```
ncapek, line 54: These default values look suspicious. I expect that if you are able to get `bucket_info` it is guaranteed to have an 'id' field.

src/keboola_mcp_server/tools/doc.py
```diff
@@ -31,6 +33,53 @@ class DocsAnswer(BaseModel):
...
+            return cls(name=item.name, type=item.type, id=item.id, created=item.created, additional_info=add_info)
+
+    total_count: int = Field(description='Total number of items found in the global search.')
+    counts_by_type: dict[str, int] = Field(description='Number of items found by type.')
+    items: list[Item] = Field(
+        description=('List of items found in the global search, sorted by relevance and creation time.')
+    )
+
+    @classmethod
+    def from_api_response(cls, response: GlobalSearchResponse) -> 'GlobalSearchAnswer':
```
ncapek, line 74: If we are going to be using these factory methods from structured API objects, we should consider using `model_construct` to bypass validation. It should speed up creating of pydantic objects:
https://docs.pydantic.dev/2.4/concepts/models/

src/keboola_mcp_server/client.py
```diff
@@ -359,7 +359,7 @@ async def delete(
 class GlobalSearchResponse(BaseModel):
     """The SAPI global search response."""
 
-    class GlobalSearchResponseItem(BaseModel):
+    class Item(BaseModel):
         id: str = Field(description='The id of the item.')
         name: str = Field(description='The name of the item.')
         type: GlobalSearchTypes = Field(description='The type of the item.')
```
radcliffkey, line 365: I have missed it before, but shouldn't it be `GlobalSearchType`? It's not very common for a class name to be plural.

src/keboola_mcp_server/server.py
```diff
@@ -170,6 +171,7 @@ async def oauth_callback_handler(request: Request) -> Response:
 
     add_component_tools(mcp)
     add_doc_tools(mcp)
+    add_search_tools(mcp)
```
radcliffkey, line 174: pls keep the alphabetic order of tools

src/keboola_mcp_server/client.py
radcliffkey: the same for `GlobalSearchBranchTypes`

src/keboola_mcp_server/tools/doc.py
radcliffkey: https://docs.pydantic.dev/2.4/concepts/models/#creating-models-without-validation

But we should check whether the following is true:

Pydantic also provides the `model_construct()` method, which allows models to be created without validation. This can be useful in at least a few cases:
    - when working with complex data that is already known to be valid (for performance reasons)
    - when one or more of the validator functions are non-idempotent, or
    - when one or more of the validator functions have side effects that you don't want to be triggered.

src/keboola_mcp_server/tools/doc.py
radcliffkey: But if that field should be there and is missing, then we should throw an exception, because it's a wrong API response.

src/keboola_mcp_server/tools/doc.py
ncapek: We can have another discussion about the proper prefix across all tools. We're not being consistent anyway, e.g. `get_component` returns `Component`, not `ComponentOutput`.

src/keboola_mcp_server/client.py
mariankrotil: Renamed.

src/keboola_mcp_server/server.py
mariankrotil: done.

src/keboola_mcp_server/tools/search.py
```diff
@@ -0,0 +1,148 @@
...
+
+LOG = logging.getLogger(__name__)
+
+MAX_GLOBAL_SEARCH_LIMIT = 100
+DEFAULT_GLOBAL_SEARCH_LIMIT = 50
+
+
+def add_search_tools(mcp: FastMCP) -> None:
+    """Add tools to the MCP server."""
+    doc_tools = [
```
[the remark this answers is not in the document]
mariankrotil: b30f13c289e1125f452890e481b0f758fb6a67c8

src/keboola_mcp_server/tools/doc.py
mariankrotil: @radcliffkey regarding this `when working with complex data that is already known to be valid (for performance reasons)` we can use `model_construct` as the data are already validated via `GlobalSearchResponse` from the client response

src/keboola_mcp_server/tools/search.py
```diff
@@ -0,0 +1,148 @@
...
+            ],
+        )
+
+
+@tool_errors()
+@with_session_state()
+async def global_search(
+    ctx: Context,
+    name_prefixes: Annotated[list[str], Field(description='Name prefixes to look for inside entity name.')],
+    entity_types: Annotated[
```
vita-stejskal, line 108: You use the word "items" when referring to the "things" found in the project. So, this should IMO be `item_types`.

src/keboola_mcp_server/client.py
```diff
@@ -21,8 +21,8 @@
 # Project features that can be checked with the is_enabled method
 ProjectFeature = Literal['global-search']
 # Input types for the global search endpoint parameters
-GlobalSearchBranchTypes = Literal['production', 'development']
-GlobalSearchTypes = Literal[
+GlobalSearchBranchType = Literal['production', 'development']
+GlobalSearchType = Literal[
```
vita-stejskal, line 25: This should really be called `ItemType` or `ArtifactType`. It does not refer to the type of a search, but to the types of the items to find or that were found.

src/keboola_mcp_server/client.py
```diff
@@ -21,8 +21,8 @@
 # Project features that can be checked with the is_enabled method
 ProjectFeature = Literal['global-search']
 # Input types for the global search endpoint parameters
-GlobalSearchBranchTypes = Literal['production', 'development']
-GlobalSearchTypes = Literal[
+GlobalSearchBranchType = Literal['production', 'development']
```
vita-stejskal, line 24: This should probably just be `BransgType` I think.

src/keboola_mcp_server/tools/search.py
```diff
@@ -0,0 +1,148 @@
...
+                if item.type == 'configuration-row':
+                    # as row_config is identified by root_config id and component id.
+                    configuration_info = item.full_path['configuration']
+                    add_info['configuration_id'] = configuration_info['id']
+                    add_info['configuration_name'] = configuration_info['name']
+            return cls.model_construct(name=item.name, id=item.id, created=item.created, additional_info=add_info)
+
+    group_type: GlobalSearchType = Field(description='The type of the items in the group.')
+    group_count: int = Field(description='Number of items in the group.')
+    group_items: list[GroupItem] = Field(
```
vita-stejskal, line 64: I'd rename these fields and call them -- `items`, `items_type` and `items_count`.

src/keboola_mcp_server/tools/search.py
```diff
@@ -0,0 +1,148 @@
...
+            group_count=len(group_items),
+            group_items=[GlobalSearchItemsGroup.GroupItem.from_api_response(item) for item in group_items],
+        )
+
+
+class GlobalSearchOutput(BaseModel):
+    """A result of a global search query for multiple name substrings."""
+
+    counts: dict[str, int] = Field(description='Number of items found for each type.')
+    type_groups: list[GlobalSearchItemsGroup] = Field(description='List of results grouped by type.')
```
vita-stejskal, line 86: I'd call this `items_by_type` and changed the description to say `search results`.

vita-stejskal: Another alternative would be to add separate field for each item type. The JSON would look even simpler, although the pydantic model would be a bit more hard-coded.

```
{
  "counts": {...},
  "buckets": [
    {"id": "...", ...}
  ],
  "tables": [
    ...
  ]
}
```

mariankrotil: > Another alternative would be to add separate field for each item type. The JSON would look even simpler, although the pydantic model would be a bit more hard-coded.
I also thought about this. And I could use a dictionary like: `items_by_type: dict[GlobalSearchType, ItemsGroup]`

src/keboola_mcp_server/tools/search.py
```diff
@@ -0,0 +1,148 @@
...
+        global_search,
+    ]
+    for tool in search_tools:
+        LOG.info(f'Adding tool {tool.__name__} to the MCP server.')
+        mcp.add_tool(FunctionTool.from_function(tool))
+
+    LOG.info('Search tools initialized.')
+
+
+class GlobalSearchItemsGroup(BaseModel):
```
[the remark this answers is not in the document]
mariankrotil: I agree. Renamed to `ItemsGroup` 1db2bd960be6ec49697ca7768bd81eb4f07e239b

src/keboola_mcp_server/tools/search.py
mariankrotil: Sure, definitely. 1db2bd960be6ec49697ca7768bd81eb4f07e239b

src/keboola_mcp_server/tools/search.py
mariankrotil: renamed here: 1db2bd960be6ec49697ca7768bd81eb4f07e239b

src/keboola_mcp_server/tools/doc.py
```diff
@@ -11,6 +11,9 @@
 
 LOG = logging.getLogger(__name__)
 
+MAX_GLOBAL_SEARCH_LIMIT = 100
+DEFAULT_GLOBAL_SEARCH_LIMIT = 50
```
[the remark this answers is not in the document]
mariankrotil: 1db2bd960be6ec49697ca7768bd81eb4f07e239b

src/keboola_mcp_server/client.py
mariankrotil: renamed here 1db2bd960be6ec49697ca7768bd81eb4f07e239b

src/keboola_mcp_server/client.py
mariankrotil: Renamed here:  1db2bd960be6ec49697ca7768bd81eb4f07e239b

src/keboola_mcp_server/tools/search.py
mariankrotil: 1db2bd960be6ec49697ca7768bd81eb4f07e239b

src/keboola_mcp_server/tools/search.py
```diff
@@ -59,54 +59,51 @@ def from_api_response(cls, item: GlobalSearchResponse.Item) -> 'GlobalSearchItem
...
-    counts: dict[str, int] = Field(description='Number of items found for each type.')
-    type_groups: list[GlobalSearchItemsGroup] = Field(description='List of results grouped by type.')
+    counts: dict[str, int] = Field(description='Number of items in total and for each type.')
+    groups: dict[ItemType, ItemsGroup] = Field(description='Search results.')
 
     @classmethod
     def from_api_responses(cls, response: GlobalSearchResponse) -> 'GlobalSearchOutput':
-        """Creates a GlobalSearchResult from the API responses."""
+        """Creates a GlobalSearchOutput from the API responses."""
         items_by_type = defaultdict(list)
```
vita-stejskal, line 89: Adding a typing annotation here would be nice.

vita-stejskal (approved):

PR #204, opened 2025-07-09
AI-1172 Move Find Component Id tool to Search tool module

mariankrotil:
Do we want to cover this tool in the search tool module?


vita-stejskal (approved):
I think it makes sense to move this tool among the search tools. It's in fact a search tool and the component tools are many and large.

PR #205, opened 2025-07-12
Ai 1280 conditional flows

ncapek:
This PR implements the following changes to flow tools:
1) refactor related to new models in Configuration models
2) add the conditional flow config schema
3) updates tools to handle conditional flows


src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -220,15 +213,15 @@ async def list_flows(
 async def get_flow(
     ctx: Context,
     configuration_id: Annotated[str, Field(description='ID of the flow configuration to retrieve.')],
-) -> Annotated[FlowConfigurationResponse, Field(description='Detailed flow configuration.')]:
+) -> Annotated[Flow, Field(description='Detailed flow configuration.')]:
     """Gets detailed information about a specific flow configuration."""
 
     client = KeboolaClient.from_state(ctx.session.state)
```
ncapek, line 219: I don't like this pattern repeating in every single tool. We could probably upgrade the decorator to return these commonly constructed objects directly.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -220,15 +213,15 @@ async def list_flows(
...
     links_manager = await ProjectLinksManager.from_client(client)
-    raw_config = await client.storage_client.flow_detail(configuration_id)
 
-    flow_response = FlowConfigurationResponse.model_validate(raw_config)
-    links = links_manager.get_flow_links(flow_response.configuration_id, flow_name=flow_response.configuration_name)
-    flow_response.links = links
+    raw_config = await client.storage_client.flow_detail(configuration_id)
```
ncapek, line 222: The tools can be ever further simplified by extracting all of this processing logic into a `FlowService` and calling the method directly in the tool. But that's likely overkill.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -187,48 +200,59 @@ async def update_flow(
...
+            if e.response.status_code == 404:
+                LOG.info(f'Could not find flow {configuration_id} under flow type {type_}.')
+            else:
+                raise
+
+    if not found_type:
+        raise ValueError(f'Flow configuration "{configuration_id}" not found.')
+
+    api_flow = APIFlowResponse.model_validate(raw_config)
+    links = links_manager.get_flow_links(api_flow.configuration_id, flow_name=api_flow.name)
```
odinuv, line 254: the flows and conditional flows have different url (flows and flows-v2)

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -187,48 +200,59 @@ async def update_flow(
...
+            raw_config = await client.storage_client.flow_detail(
+                config_id=configuration_id,
+                flow_type=type_,
+            )
+            found_type = type_
+            LOG.info(f'Found flow {configuration_id} under flow type {type_}.')
+            break
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 404:
+                LOG.info(f'Could not find flow {configuration_id} under flow type {type_}.')
```
odinuv, line 246: this is a bit redundant since it's logged again on line 250

src/keboola_mcp_server/tools/flow/conditional_flow_model.py
```diff
@@ -0,0 +1,174 @@
+from typing import List, Literal, Optional, Union
```
ncapek, line 1: @vita-stejskal This file was an attempt to make conditional flows work with pydantic objects. I would recommend removing this completely and trying to make it work with the original approach using `dict[str, Any]`

src/keboola_mcp_server/tools/flow/api_models.py
```diff
@@ -0,0 +1,85 @@
+"""
```
ncapek, line 1: @vita-stejskal This file is part of refactoring - to use API DTOs from the client to create domain objects.

src/keboola_mcp_server/links.py
```diff
@@ -51,19 +51,27 @@ def get_project_links(self) -> list[Link]:
         return [self.get_project_detail_link()]
 
     # --- Flows ---
-    def get_flow_detail_link(self, flow_id: str | int, flow_name: str) -> Link:
-        return Link.detail(title=f'Flow: {flow_name}', url=self._url(f'flows/{flow_id}'))
-
-    def get_flows_dashboard_link(self) -> Link:
-        return Link.dashboard(title='Flows in the project', url=self._url('flows'))
+    def get_flow_detail_link(self, flow_id: str | int, flow_name: str,
```
ncapek, line 54: @vita-stejskal These changes are just to patch the broken link to support both flow types.

src/keboola_mcp_server/client.py
```diff
@@ -36,6 +36,9 @@
 ]
 
 ORCHESTRATOR_COMPONENT_ID = 'keboola.orchestrator'
+CONDITIONAL_FLOW_COMPONENT_ID = 'keboola.flow'
+FLOW_TYPE = Literal['keboola.flow', 'keboola.orchestrator']
+FLOW_TYPES: Sequence[FLOW_TYPE] = (CONDITIONAL_FLOW_COMPONENT_ID, ORCHESTRATOR_COMPONENT_ID)
```
ncapek, line 41: @vita-stejskal Changes to the client should work fine, they just add a parameter to handle the `flow_component_id` passed to the endpoint.

pyproject.toml
```diff
@@ -16,7 +16,7 @@ dependencies = [
     "httpx ~= 0.28",
     "jsonschema ~= 4.23",
     "pyjwt ~= 2.10",
-    "json-log-formatter ~= 1.0",
```
ncapek, line 19: Can be ignored. I needed to update the `pydantic[email]` for the experiment with conditional flow pydantic models.

integtests/tools/flow/test_tools.py
```diff
@@ -7,7 +7,7 @@
 from keboola_mcp_server.client import ORCHESTRATOR_COMPONENT_ID, KeboolaClient
```
ncapek, line 7: In general, all tests need to be re-thought since we are now supporting multiple flow types.

src/keboola_mcp_server/tools/flow/model.py
```diff
@@ -1,29 +1,25 @@
...
 
 
 class ListFlowsOutput(BaseModel):
     """Output of list_flows tool."""
-
-    flows: List['ReducedFlow'] = Field(description='The retrieved flow configurations.')
-    links: List[Link] = Field(
-        description='The list of links relevant to the flows.',
-    )
+    flows: list['FlowSummary'] = Field(description='The retrieved flow configurations.')
```
ncapek, line 13: @vita-stejskal This was also part of refactoring to make flow tools consistent with component tools.

src/keboola_mcp_server/tools/flow/model.py
```diff
@@ -33,12 +29,11 @@ class FlowPhase(BaseModel):
...
     """Represents a task in a flow configuration."""
-
-    id: Union[int, str] = Field(description='Unique identifier of the task')
+    id: int | str = Field(description='Unique identifier of the task')
     name: str = Field(description='Name of the task')
-    phase: Union[int, str] = Field(description='ID of the phase this task belongs to')
-    enabled: bool = Field(default=True, description='Whether the task is enabled')
-    continue_on_failure: bool = Field(
+    phase: int | str = Field(description='ID of the phase this task belongs to')
+    enabled: Optional[bool] = Field(default=True, description='Whether the task is enabled')
```
ncapek, line 35: @vita-stejskal I had to make this `Optional` to be able to fetch a `keboola.flow` using the `get_flow` tool. I didn't have to check if this follows from the schema change.

src/keboola_mcp_server/tools/flow/model.py
```diff
@@ -49,89 +44,116 @@ class FlowTask(BaseModel):
...
+            configuration=config,
+            change_description=api_config.change_description,
+            configuration_metadata=api_config.metadata,
+            created=api_config.created,
+            updated=api_config.updated,
+            links=links
+        )
+
+
+class FlowSummary(BaseModel):
```
ncapek, line 123: @vita-stejskal Also part of refactoring. `ReducedFlow` -> `FlowSummary` for listing

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -41,11 +55,33 @@ def add_flow_tools(mcp: FastMCP) -> None:
 
 
 @tool_errors()
-async def get_flow_schema() -> Annotated[str, Field(description='The configuration schema of Flow.')]:
-    """Returns the JSON schema that defines the structure of Flow configurations."""
+@with_session_state()
+async def get_flow_schema(
```
ncapek, line 59: @vita-stejskal In general there were many discussions on what the expected behaviour of various flow tools should be. I believe it's still up for debate. But I went with: "If conditional flows are disabled, always return the old flow schema. Otherwise, let the agent decide which one it wants to fetch."

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -56,6 +92,7 @@ async def create_flow(
     description: Annotated[str, Field(description='Detailed description of the flow purpose.')],
     phases: Annotated[list[dict[str, Any]], Field(description='List of phase definitions.')],
     tasks: Annotated[list[dict[str, Any]], Field(description='List of task definitions.')],
+
```
ncapek, line 95: @vita-stejskal `create_flow` has not been properly updated. I guess the easiest way to go is to add another version of the tool called `create_conditional_flow`. We could even decide to hide it for projects where conditional flows are disabled.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -109,10 +146,71 @@ async def create_flow(
...
+    flow_links = links_manager.get_flow_links(flow_id=flow_id, flow_name=flow_name, flow_type=ORCHESTRATOR_COMPONENT_ID)
+    tool_response = FlowToolResponse.model_validate(new_raw_configuration | {'links': flow_links})
+
+    LOG.info(f'Created flow "{name}" with configuration ID "{flow_id}" (type: {ORCHESTRATOR_COMPONENT_ID})')
+    return tool_response
+
+
+@tool_errors()
+@with_session_state()
+async def create_conditional_flow(
```
ncapek, line 158: @vita-stejskal My first attempt at this - it is not working well right now.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -187,48 +285,64 @@ async def update_flow(
...
+    links = [
+        links_manager.get_flows_dashboard_link(ORCHESTRATOR_COMPONENT_ID),
+        links_manager.get_flows_dashboard_link(CONDITIONAL_FLOW_COMPONENT_ID),
+    ]
     return ListFlowsOutput(flows=flows, links=links)
 
 
 @tool_errors()
 @with_session_state()
 async def get_flow(
```
ncapek, line 312: @vita-stejskal For `get_flow` the logic is "First try to fetch it as the new version and return. If not found, fetch it under the previous version and return."

src/keboola_mcp_server/tools/flow/utils.py
```diff
@@ -3,26 +3,37 @@
...
+)
+from keboola_mcp_server.tools.flow.api_models import APIFlowResponse
+from keboola_mcp_server.tools.flow.model import FlowPhase, FlowSummary, FlowTask
 
 LOG = logging.getLogger(__name__)
 
 RESOURCES = 'keboola_mcp_server.resources'
-FLOW_SCHEMA_RESOURCE = 'flow-schema.json'
+FLOW_SCHEMAS: Mapping[FLOW_TYPE, str] = {
+    CONDITIONAL_FLOW_COMPONENT_ID: 'flow-schema.json',
```
ncapek, line 23: @vita-stejskal !!!! This is flipped. Please be careful !!!!!

src/keboola_mcp_server/tools/project.py
```diff
@@ -75,6 +61,8 @@ async def get_project_info(
     )
 
     sql_dialect = await WorkspaceManager.from_state(ctx.session.state).get_sql_dialect()
+    project_features = cast(JsonDict, token_data.get('features', {}))
+    conditional_flows_disabled = 'conditional-flows-disabled' in project_features
```
ncapek, line 65: @vita-stejskal I was told by @odinuv that projects will have this field under "features" - disabling conditional flows will be opt-in.

ncapek (commented):
Summary of changes for @vita-stejskal . In general, `get_flow` and `list_flows` are working fine. `get_flow_schema` has the mapping flipped. Project links related to flow were patched but I haven't tested those thoroughly. The main things to fix now are `create_flow` and `update_flow`. Also the tests need to be re-designed.

src/keboola_mcp_server/tools/flow/conditional_flow_model.py
ncapek: I got rid of this.

src/keboola_mcp_server/tools/flow/tools.py
ncapek: I got rid of this.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -85,35 +126,122 @@ async def create_flow(
...
+    - user_input: Create a flow with complex conditional logic and retry mechanisms.
+    - user_input: Build a data pipeline with sophisticated error handling and notifications.
+    """
+    flow_type = CONDITIONAL_FLOW_COMPONENT_ID
+    processed_phases = [ConditionalFlowPhase.model_validate(phase) for phase in phases]
+    processed_tasks = [ConditionalFlowTask.model_validate(task) for task in tasks]
+    flow_configuration = {
+        'phases': [phase.model_dump(exclude_unset=True, by_alias=True) for phase in processed_phases],
+        'tasks': [task.model_dump(exclude_unset=True, by_alias=True) for task in processed_tasks],
+    }
```
mariankrotil, line 219: Would not be easier to just check the json schema against the `flow_configuration` and avoid using the pydantic models for Phases and Tasks? Or experiment with the Pydantic models in the parameters?

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -85,35 +126,122 @@ async def create_flow(
...
+        id=api_config.id,
+        description=api_config.description,
+        timestamp=datetime.now(timezone.utc),
+        success=True,
+        links=flow_links
+    )
 
-    LOG.info(f'Created flow "{name}" with configuration ID "{flow_id}"')
+    LOG.info(f'Created conditional flow "{name}" with configuration ID "{api_config.id}" (type: {flow_type})')
     return tool_response
```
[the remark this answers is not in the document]
ncapek: Which part do you mean? The logging or the FlowToolResponse creation? Neither of them feel like they're justified.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -122,6 +250,15 @@ async def create_flow(
 async def update_flow(
     ctx: Context,
     configuration_id: Annotated[str, Field(description='ID of the flow configuration to update.')],
+    flow_type: Annotated[
```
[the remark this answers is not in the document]
ncapek: Are you sure? To me the flow type feels only like a switch to toggle the settings.

src/keboola_mcp_server/tools/flow/utils.py
```diff
@@ -3,31 +3,100 @@
...
+                raise ValueError(f'Phase {phase.id} depends on non-existent phase {dep_id}')
+
+    for task in tasks:
+        if task.phase not in phase_ids:
+            raise ValueError(f'Task {task.id} references non-existent phase {task.phase}')
+
+    _check_legacy_circular_dependencies(phases)
+
+
+def _check_legacy_circular_dependencies(phases: list[FlowPhase]) -> None:
```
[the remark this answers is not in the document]
ncapek: I reverted back to the original implementation.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -85,35 +126,122 @@ async def create_flow(
...
+    - **Retry configurations**: RetryConfiguration with RetryStrategyParams and RetryOnCondition
+    - **Task configurations**: JobTaskConfiguration, NotificationTaskConfiguration, VariableTaskConfiguration
+
+    CONSIDERATIONS:
+    - Do not create conditions, unless user asks for them explicitly
+    - All IDs must be unique and clearly defined.
+    - The `phases` and `tasks` parameters must conform to the keboola.flow JSON schema.
+    - The phases cannot be empty.
+    - This tool automatically validates the structure before creation.
+    - The enhanced models provide better type safety and validation.
```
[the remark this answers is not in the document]
ncapek: Removed

src/keboola_mcp_server/tools/flow/tools.py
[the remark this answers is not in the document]
mariankrotil: No strong opinion.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -85,35 +126,122 @@ async def create_flow(
...
     )
 
+    LOG.info(f'Created legacy flow "{name}" with configuration ID "{api_config.id}" (type: {flow_type})')
+    return tool_response
+
+
+@tool_errors()
+@with_session_state()
+async def create_conditional_flow(
+    ctx: Context,
```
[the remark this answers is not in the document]
ncapek: David already played around a little with the descriptions and got the agent to more consistently fetch the schema . See `create_conditional_flow` description. I also copy-pasted that part of the description into the `create_flow` tool.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -85,35 +126,122 @@ async def create_flow(
...
+    - **Structured conditions**: Type-safe condition objects with proper validation
+    - **Structured retry configurations**: Properly typed retry parameters and conditions
+    - **Structured task configurations**: Type-safe job, notification, and variable task configurations
+    - **Better validation**: Compile-time validation of flow structure and relationships
+    - **Enhanced error messages**: More descriptive errors when validation fails
+
+    Enhanced conditional flows use structured models for:
+    - **Conditions**: TaskCondition, PhaseCondition, OperatorCondition, etc.
+    - **Retry configurations**: RetryConfiguration with RetryStrategyParams and RetryOnCondition
+    - **Task configurations**: JobTaskConfiguration, NotificationTaskConfiguration, VariableTaskConfiguration
```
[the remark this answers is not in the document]
ncapek: Thanks. I'll remove this. We should certainly play around with the descriptions a little more. This PR was more focused on the code.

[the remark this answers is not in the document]
ncapek: Removed

src/keboola_mcp_server/tools/flow/utils.py
```diff
@@ -102,63 +171,46 @@ def ensure_task_ids(tasks: list[dict[str, Any]]) -> list[FlowTask]:
...
+                found_flow_type = flow_type
+                break
+            except Exception:
+                continue
+
+        if raw_flow and found_flow_type:
+            api_flow = APIFlowResponse.model_validate(raw_flow)
+            flows.append(FlowSummary.from_api_response(api_config=api_flow, flow_component_id=found_flow_type))
+        else:
+            LOG.warning(f'Failed to retrieve flow {flow_id}.')
```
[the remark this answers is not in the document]
ncapek: I changed this code a little. Added a helper function to find the flow by trying both flow types. And used it in `_get_flows_by_ids`.

src/keboola_mcp_server/tools/flow/tools.py
ncapek: Fixed.

src/keboola_mcp_server/tools/flow/tools.py
```diff
@@ -187,48 +200,59 @@ async def update_flow(
...
-    ] = tuple(),
+    flow_ids: Annotated[Optional[Sequence[str]], Field(description='IDs of the flows to retrieve.')] = None,
 ) -> ListFlowsOutput:
-    """Retrieves flow configurations from the project."""
+    """Retrieves flow configurations from the project. Optionally filtered by IDs."""
 
     client = KeboolaClient.from_state(ctx.session.state)
     links_manager = await ProjectLinksManager.from_client(client)
 
     if flow_ids:
```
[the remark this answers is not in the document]
ncapek: We could do that. But doesn't the flow listing endpoint have paginations, limits, offsets? If the project has many flows, this could be problematic. The utility functions construct and return structured objects which are already processed from the endpoint calls so the tool should be getting consistent data.

src/keboola_mcp_server/tools/flow/tools.py
ncapek: It's gone.

src/keboola_mcp_server/tools/flow/__init__.py
vita-stejskal, line 1: I have to admit that I still struggle to see meaning behind these imports in `__init__.py`. IMO they only clutter the project and make it harder to see where the imported stuff really comes from.

src/keboola_mcp_server/tools/flow/api_models.py
```diff
@@ -0,0 +1,85 @@
...
+
+from typing import Any, Optional
+
+from pydantic import AliasChoices, BaseModel, Field, model_validator
+
+from keboola_mcp_server.client import ORCHESTRATOR_COMPONENT_ID
+
+
+class APIFlowResponse(BaseModel):
```
vita-stejskal, line 15: The same comment that I had to `CreateConfigurationAPIResponse` applies here. This should be added to `client.py` and returned by its functions, right?

OTOH, I also commented somewhere else that the `flow_*` functions in `AsyncStorageClient` class should be removed, because they no longer give any advantage over `configuration_*` functions. So, this response model could perhaps be renamed and returned from the `configuration_*` functions.

src/keboola_mcp_server/tools/flow/__init__.py
vita-stejskal, line 1: I have to admit that I still struggle to see compelling reasons for adding these imports into `__init__.py` files. IMO they only make it harder to see where the imported stuff actually comes from.

ncapek: I had this discussion with @radcliffkey and @mariankrotil and I believe the main motivation is that it allows us to improt from `keboola_mcp_server.tools.flow` rather than `keboola_mcp_server.tools.flow.tools` which contains double "tools" and is not very pretty. I agree that it's a trivial thing.

src/keboola_mcp_server/tools/components/api_models.py
```diff
@@ -126,3 +126,37 @@ class ConfigurationAPIResponse(BaseModel):
         description='Configuration metadata',
         validation_alias=AliasChoices('metadata', 'configuration_metadata', 'configurationMetadata'),
     )
+
+
+class CreateConfigurationAPIResponse(BaseModel):
```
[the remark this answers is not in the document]
ncapek: I can move it to the client but we don't have the client responses in a separate model file yet and `client.py` is quite large now.

src/keboola_mcp_server/tools/flow/api_models.py
ncapek: I moved the API reponses into the `client.py` module. But it's really huge now and should be further refactored.

src/keboola_mcp_server/client.py
```diff
@@ -800,52 +803,58 @@ async def flow_create(
         name: str,
         description: str,
         flow_configuration: dict[str, Any],
+        flow_type: FLOW_TYPE = CONDITIONAL_FLOW_COMPONENT_ID,
     ) -> JsonDict:
```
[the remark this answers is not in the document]
ncapek: Removed

src/keboola_mcp_server/tools/project.py
```diff
@@ -26,27 +26,13 @@ def add_project_tools(mcp: FastMCP) -> None:
...
-        ...,
-        description='The sql dialect used in the project.'
-    )
-    links: list[Link] = Field(..., description='The links relevant to the project.')
+    project_id: str | int = Field(description='The id of the project.')
+    project_name: str = Field(description='The name of the project.')
+    project_description: str = Field(description='The description of the project.',)
+    organization_id: str | int = Field(description='The ID of the organization this project belongs to.')
+    sql_dialect: str = Field(description='The sql dialect used in the project.')
+    conditional_flows_disabled: bool = Field(description='Whether the project has conditional flows disabled.')
```
vita-stejskal, line 34: I don't think that we should copy the inverted nature of this project feature. The field should be:
```
conditional_flows: bool = Field(description='Whether the project supports conditional flows.')
```
In general, it's easier to understand boolean flags when they are formulated in the direct way rather than in the inverted way. The inverted way usually just works around the lack of the proper means of enforcing the flag's default value. Since we always set this flag, it's better to use the direct way flag.

After all, this flag will be visible to the LLMs.

src/keboola_mcp_server/tools/project.py
vita-stejskal: @odinuv I can't access the link. What is the final name of this project feature? @ncapek please make sure to use the correct name. Check both names if not sure.

tests/tools/flow/test_model.py
vita-stejskal, line 1: AFAICT you've updated the existing tests written for testing the legacy flows (`keboola.orchestrator`), but you did not write any new tests to cover the conditional flows (`keboola.flow`). Would you add them later or are they not necessary?

integtests/tools/flow/test_tools.py
```diff
@@ -1,24 +1,35 @@
...
     get_flow_schema,
     list_flows,
     update_flow,
 )
+from keboola_mcp_server.tools.project import get_project_info
+
+LOG = logging.getLogger(__name__)
 
 
+@pytest.mark.skip
```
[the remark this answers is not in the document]
ncapek: Unskipped

src/keboola_mcp_server/tools/flow/utils.py
```diff
@@ -102,63 +182,70 @@ def ensure_task_ids(tasks: list[dict[str, Any]]) -> list[FlowTask]:
...
-        if task.phase not in phase_ids:
-            raise ValueError(f'Task {task.id} references non-existent phase {task.phase}')
-
-    _check_circular_dependencies(phases)
+class FlowResolutionError(Exception):
+    pass
 
 
-def _check_circular_dependencies(phases: list[FlowPhase]) -> None:
+async def _resolve_flow_by_id(client: KeboolaClient, flow_id: str) -> tuple[APIFlowResponse, FlowType]:
```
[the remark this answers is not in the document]
ncapek: I have mixed feelings about the name. Originally I called the function `find_flow_(by_id)` but it didn't quite feel right because once the migrations happen, flows with the same ID can exist under both flow types. So I was looking for a word to indicate that it's unclear what the flow type of a given flow is and it needs to be figured out during the lookup process.

src/keboola_mcp_server/client.py
```diff
@@ -1274,6 +1196,216 @@ class ComponentSuggestionResponse(BaseModel):
...
+        description='Flow configuration metadata',
+        validation_alias=AliasChoices('metadata', 'configuration_metadata', 'configurationMetadata'),
+    )
+
+    # Timestamps
+    created: Optional[str] = Field(None, description='Creation timestamp')
+    updated: Optional[str] = Field(None, description='Last update timestamp')
+
+
+class ComponentAPIResponse(BaseModel):
```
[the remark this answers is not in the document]
ncapek: I can move it. But I think we had a separate task for client refactoring that encompasses changes related to the client methods returning structured objects.

src/keboola_mcp_server/tools/flow/utils.py
```diff
@@ -102,63 +182,70 @@ def ensure_task_ids(tasks: list[dict[str, Any]]) -> list[FlowTask]:
...
+    errors = []
 
-        dependencies = graph.get(phase_id, [])
+    for flow_type in FLOW_TYPES:
+        try:
+            raw_flow = await client.storage_client.configuration_detail(
+                component_id=flow_type,
+                configuration_id=flow_id
+                )
+            api_flow = APIFlowResponse.model_validate(raw_flow)
```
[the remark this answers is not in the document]
ncapek: No. I just didn't want to overbloat this PR which is already very large. And I think we have separate JIRA tasks related to client refactoring.

src/keboola_mcp_server/client.py
```diff
@@ -1274,6 +1196,216 @@ class ComponentSuggestionResponse(BaseModel):
...
+        validation_alias=AliasChoices('changeDescription', 'change_description', 'change-description'),
+    )
+    metadata: list[dict[str, Any]] = Field(
+        default_factory=list,
+        description='Configuration metadata',
+        validation_alias=AliasChoices('metadata', 'configuration_metadata', 'configurationMetadata'),
+    )
+
+
+class CreateConfigurationAPIResponse(BaseModel):
```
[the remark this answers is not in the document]
ncapek: @odinuv If I fetch the raw API response from an endpoint, am I guaranteed to receive all the fields, i.e. can I model all of the API response fields based on a specific response? Or can two responses (from the same endpoint) return a slightly different set of fields?

src/keboola_mcp_server/client.py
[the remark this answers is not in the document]
ncapek: I identified some strange patterns. For example `ConfigurationAPIResponse` current contains a field for `component_id` which is not actually returned in the config detail endpoint. Changing the pydantic model to perfectly mimic the API response would break some things in configuration tools. Therefore, I suggest that we leave this as is for now and deal with that in a different PR.

tests/tools/flow/test_model.py
vita-stejskal: > all projects can run both flow types and therefore it's okay to run all flow integrations tests on all projects. 

Yes. Plus, the unit tests work in the mocked environment and so should be testing both the legacy and conditional flows regardless of whether CF are deployed/available or not. Since the CF have already been deployed and are generally available on all stacks and in all projects the Integration tests can now also test both the LF and CF.

> The only place that needs to be updated is list_tools, right?

Yes. The `ToolsFilteringMiddleware` needs to be updated to let through the correct variant of the "create flow" tool. Similarly, the integration tests that check the tools available on a server and visible to a connected client will have to be updated in a similar way as I did for the `search` tool. The change needs to be done in [test_mcp_server.py:239](https://github.com/keboola/mcp-server/blob/main/integtests/test_mcp_server.py#L239).

src/keboola_mcp_server/tools/flow/utils.py
```diff
@@ -102,63 +182,70 @@ def ensure_task_ids(tasks: list[dict[str, Any]]) -> list[FlowTask]:
...
-                try:
-                    cycle_start_index = path.index(dep_id)
-                    return path[cycle_start_index:] + [dep_id]
-                except ValueError:
-                    return [phase_id, dep_id]
 
-        path.pop()
-        rec_stack.remove(phase_id)
-        return None
+async def _get_flows_by_ids(
```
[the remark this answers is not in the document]
ncapek: Fixed

src/keboola_mcp_server/tools/flow/utils.py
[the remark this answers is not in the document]
vita-stejskal: Ok, I understand. Let's straighten this up in another PR under [AI-1247](https://keboola.atlassian.net/browse/AI-1247).

Just to make sure we understand each other well. This usage pattern is exactly what I want to get rid of when talking about the `AsyncStorageClient` functions returning the response models rather than the generic `JsonDict` instance.

src/keboola_mcp_server/tools/project.py
ncapek: Fixed

src/keboola_mcp_server/client.py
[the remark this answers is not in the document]
ncapek: > The client function can add the component_id to the API response so that the model can pick it up.
I thought that API responses are designed solely to mimic and structure the API response. Adding additional fields that the API does not return feels like something else and should probably be handled in the domain models, no?

> !! This is not and has never been the goal. !!
> 
> The pydantic models for SAPI response only need to carry enough information to support the MCP tools. 
It seems to me that if a developer is going to put in the work to model the API responses, it's much easier to just use all the fields and filter out the important information downstream in the domain models. Otherwise we might eventually find ourselves having to add the missing fields back into the API models which sounds like very inefficient work.

src/keboola_mcp_server/tools/flow/utils.py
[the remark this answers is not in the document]
ncapek: I have removed the awkward error handling but kept the name for now.

tests/tools/flow/test_model.py
ncapek: Integration tests added.

vita-stejskal (approved):

PR #207, opened 2025-07-15
Ai 1243 mcp update mcp and fastmcp libraries

mariankrotil:
We upgraded to "fastmcp == 2.10.4" and "mcp == 1.10.1". The latest fastmcp version introduced an issue related to validation_aliases and serialization aliases in Pydantic base models. Specifically, the output schema generated for tools only considers the first validation_alias, and if it doesn't match the serialization alias, the validation fails. We've sorted the aliases to work around the issue and have reported it on their repo here: https://github.com/jlowin/fastmcp/issues/1142

Additionally, this PR resolves the security alert. https://github.com/keboola/mcp-server/security/dependabot/1

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,15 +4,15 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.12.0"
+version = "1.12.1"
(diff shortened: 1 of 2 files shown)


radcliffkey (approved):
Don't forget to update `uv.lock`, too

PR #210, opened 2025-07-20
Ai 1205 filter tools

vita-stejskal:


radcliffkey (approved):
great job!

src/keboola_mcp_server/mcp.py
```diff
@@ -77,138 +77,191 @@ def add_tool(self, tool: Tool) -> None:
...
+        MCP functions down the chain. It is called for each tool, prompt, resource, etc. calls.
+
+        :param context: Middleware context containing FastMCP context.
+        :param call_next: Next middleware in the chain to call.
+        :returns: Result from executing the middleware chain.
         """
+        ctx = context.fastmcp_context
+        assert isinstance(ctx, Context), f'Expecting Context, got {type(ctx)}.'
+
+        if not isinstance(ctx.session, MagicMock):
```
mariankrotil, line 118: Is it possible to remove this check, which is there only for the tests? And potentially improve mocking within the tests? I would apply it If it was a simple and quick change.

src/keboola_mcp_server/mcp.py
```diff
@@ -77,138 +77,191 @@ def add_tool(self, tool: Tool) -> None:
...
+
+            # TODO: We could probably get rid of the 'state' attribute set on ctx.session and just
+            #  pass KeboolaClient and WorkspaceManager instances to a tool as extra parameters.
+            state = self._create_session_state(config)
+            ctx.session.state = state
+
+        try:
+            return await call_next(context)
+        finally:
+            ctx.session.state = {}
```
mariankrotil, line 161: Is it necessary to reset the state to an empty dictionary `{}` after each tool call? Would not it be better to persist the state across interactions within the same user context session?

src/keboola_mcp_server/mcp.py
```diff
@@ -77,138 +77,191 @@ def add_tool(self, tool: Tool) -> None:
...
+    async def on_list_tools(
+            self,
+            context: MiddlewareContext[mt.ListToolsRequest],
+            call_next: CallNext[mt.ListToolsRequest, list[Tool]]
+    ) -> list[Tool]:
+        tools = await call_next(context)
+        features = await self.get_project_features(context.fastmcp_context)
+
+        if 'global-search' not in features:
+            tools = [t for t in tools if t.name != 'search']
```
mariankrotil, line 219: Can we reference those features using constant function names instead of hardcoded string literals?

mariankrotil (approved):

integtests/test_mcp_server.py
```diff
@@ -235,11 +234,14 @@ async def _assert_basic_setup(server: FastMCP, client: Client):
     assert len(server_tools) > 0
     assert len(server_prompts) > 0
 
-    assert len(client_tools) == len(server_tools)
+    client_tool_names = sorted(t.name for t in client_tools)
+    # ignore 'search' tool which may or may not be exposed based on the testing project's features
+    server_tool_names = sorted(t for t in server_tools.keys() if t not in ['search'])
```
mariankrotil, line 239: The tests are failing perhaps because the client's list of tools may include `search`, if it is enabled for the current test project. Hence, a server with enabled global-search will return `search` tool within the `list_tools` called from the client, but we filter out the `search` tool from the server tools list, resulting in the error when comparing.

vita-stejskal: Fixed in 96d0316 and e89591b.

src/keboola_mcp_server/mcp.py
mariankrotil: We skip this part if `ctx.session` is of a mock type and do not set the `client` or `workpsace`, as they are already set during the context mocking in the conftest file and mainly because it would fail otherwise. I suppose we could mock the context differently or mock the method that constructs the context session state itself, which would remove the MagicMock check. However, I would give up on that to save time.

src/keboola_mcp_server/mcp.py
vita-stejskal: Fixed in a637d78.

ncapek (approved):

src/keboola_mcp_server/mcp.py
```diff
@@ -77,138 +77,193 @@ def add_tool(self, tool: Tool) -> None:
...
+
+    async def on_list_tools(
+            self,
+            context: MiddlewareContext[mt.ListToolsRequest],
+            call_next: CallNext[mt.ListToolsRequest, list[Tool]]
+    ) -> list[Tool]:
+        tools = await call_next(context)
+        features = await self.get_project_features(context.fastmcp_context)
+
+        from keboola_mcp_server.tools import search
```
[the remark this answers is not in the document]
vita-stejskal: Ideally, but it cannot be due to the circular dependency.

In the long term, it might be better to come up with some tags-based system that the MCP tools could use for expressing what Keboola project features they require (or forbid) for them to work.

PR #212, opened 2025-07-28
feat: add context parameter to all MCP tools

AdamVyborny:
## Summary

Add required `context` parameter to all 33 MCP tool functions for monitoring and debugging tool usage patterns. The parameter provides a brief explanation (8-15 words) of why each tool call is being made.

## Why

In order to be able to display short context of why the tool has been used in the timeline and gain more insights into MCP Server usage patterns. The parameter won't be used in the tool itself - its sole purpose is to keep track of why the tool was called.

This enables:
- **Timeline functionality** - Display reasoning for each tool call in user interfaces
- **Usage insights** - Better understanding of how users interact with MCP tools
- **Debugging support** - Clear context for troubleshooting tool usage

## Changes

- Add `context` parameter to all tool functions with proper type annotations
- Parameter specification: `"Brief explanation of why this tool call is being made (8-15 words)"`
- Update all 287 unit tests and integration tests to include context parameter  
- Create automated enforcement test to ensure future tools include the parameter
- Handle edge case with `_ctx` parameter in `get_flow_schema` function

## Examples

The context parameter provides reasoning for tool calls:
- "Setting up Salesforce extractor for customer analysis project"
- "Checking existing configurations before creating new one" 
- "Analyzing customer data to identify top segments"
- "Executing transformation to calculate customer metrics"

## Test Plan

- [x] All 287 unit tests pass
- [x] All flake8 code style checks pass
- [x] Context parameter enforcement test validates all 33 functions
- [x] Integration tests updated (require credentials to run)

Related: [AI-1310](https://keboola.atlassian.net/browse/AI-1310)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/keboola_mcp_server/tools/components/tools.py
+++ b/src/keboola_mcp_server/tools/components/tools.py
@@ -106,6 +106,7 @@ def add_component_tools(mcp: KeboolaMcpServer) -> None:
 @with_session_state()
 async def list_configs(
     ctx: Context,
+    context: Annotated[str, Field(description='Brief explanation of why this tool call is being made (8-15 words)')],
     component_types: Annotated[
         Sequence[ComponentType],
         Field(description='List of component types to filter by. If none, return all components.'),
@@ -158,6 +159,7 @@ async def list_configs(
 @with_session_state()
 async def list_transformations(
     ctx: Context,
+    context: Annotated[str, Field(description='Brief explanation of why this tool call is being made (8-15 words)')],
     transformation_ids: Annotated[
         Sequence[str],
         Field(description='List of transformation component IDs to retrieve configurations for.'),
@@ -206,6 +208,7 @@ async def list_transformations(
 @with_session_state()
 async def get_component(
     ctx: Context,
+    context: Annotated[str, Field(description='Brief explanation of why this tool call is being made (8-15 words)')],
     component_id: Annotated[str, Field(description='ID of the component/transformation')],
 ) -> Annotated[Component, Field(description='The component.')]:
     """
@@ -230,6 +233,8 @@ async def get_component(
 @tool_errors()
 @with_session_state()
 async def get_config(
+    ctx: Context,
+    context: Annotated[str, Field(description='Brief explanation of why this tool call is being made (8-15 words)')],
     component_id: Annotated[str, Field(description='ID of the component/transformation')],
     configuration_id: Annotated[
         str,
@@ -237,7 +242,6 @@ async def get_config(
             description='ID of the component/transformation configuration',
         ),
     ],
-    ctx: Context,
 ) -> Annotated[Configuration, Field(description='The component/transformation and its configuration.')]:
     """
     Gets information about a specific component/transformation configuration.
@@ -287,6 +291,7 @@ async def get_config(
 @with_session_state()
 async def create_sql_transformation(
     ctx: Context,
+    context: Annotated[str, Field(description='Brief explanation of why this tool call is being made (8-15 words)')],
     name: Annotated[
         str,
         Field(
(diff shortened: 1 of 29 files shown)


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -106,6 +106,7 @@ def add_component_tools(mcp: KeboolaMcpServer) -> None:
 @with_session_state()
 async def list_configs(
     ctx: Context,
+    context: Annotated[str, Field(description='Brief explanation of why this tool call is being made (8-15 words)')],
     component_types: Annotated[
         Sequence[ComponentType],
         Field(description='List of component types to filter by. If none, return all components.'),
```
mariankrotil, line 112: A bit afraid that we have selected a slightly unhappy name for the reasoning tool parameter, called `context`.

It could be confused with the MCP's built-in `Context` type parameter, which enables features like accessing user session and server state, logging, interacting with users agent, elicitation, and more just from the tool. Although, we typically refer to it as `ctx`, it is still a shorthand for `context`.

Thankfully, the agent never sees the `ctx` parameter, as it is filtered out due to its `Context` type and injected only when the tool is called. However, this naming overlap might still be confusing for developers. 

Please, we may consider renaming the reasoning parameter to something clearer, like `reasoning`, `explanation`, `reason`, `purpose`, `justification`, `intent` or other synonyms. @jordanrburger

src/keboola_mcp_server/tools/doc.py
```diff
@@ -35,10 +35,17 @@ class DocsAnswer(BaseModel):
...
     query: Annotated[str, Field(description='Natural language query to search for in the documentation.')],
 ) -> Annotated[DocsAnswer, Field(description='The retrieved documentation.')]:
     """
     Answers a question using the Keboola documentation as a source.
+
+    'context' parameter provides reasoning for why the call is being made. Examples:
+    - 'Understanding OAuth setup process for Google Analytics extractor'
+    - 'Finding documentation on Storage API bucket permissions'
+    - 'Learning about transformation development best practices'
+    - 'Researching configuration options for specific component'
```
mariankrotil, line 48: Are those parameter examples really necessary? I consider the tool parameter description to be self-descriptive enough and that by providing such examples within the tool description we could overshadow the core concept of the tool.

I believe that more understandable would be to create "end2end" examples within the tool description demonstrating how to set up the parameters given the user input.

AdamVyborny: @mariankrotil it is just draft, I did not review the code yet. This is one of the things where I was thinking the same as you :)

AdamVyborny:
@cjayyy Do we want to also return the "context" in tool call response? Or you will parse it from the tool request argument?

src/keboola_mcp_server/tools/doc.py
mariankrotil: Do not worry, I just did a quick check.

AdamVyborny:
Slack discussion related: https://keboolaglobal.slack.com/archives/C094NK1JPJB/p1753778103438069

vita-stejskal:
I am against this change.

If the "justification" parameter is solely for the "timeline" implementation which is "The Client" thing it should be implemented in The Client itself and not in the MCP server. The MCP tools have no use for this parameter, the general purpose AI assistants such as Claude or Cursor have no use for this parameter either. So, all-in-all this parameter seems as a completely artificial addition that makes no sense in the context of the MCP server itself.

davidesner:
> We have two options. We would wrap tools from other MCPs with that parameter or we do it even for our MCP and we lose the posibility to get the insights for usage analytics and debugging. I have no strong opinion on this. Probably @davidesner and @jordanrburger could do the decision.

We want this in our MCP Server now, so we get insights. @vita-stejskal you are correct to say that this is artificial parameter, it is solely to help us understand how the MCP server is being used + it helps with the timeline in our client. It is questionable how much it will affect token spend. If we think it's too much we could just have it on the client side.

If we eventually decide that we would like to have this context for other tools used by our agent in order to get the timeline info, we can then add a generic tool wrapper for tools imported from external MCP servers if possible.

PR #221, opened 2025-08-04
Fix logging from FastMCP and such

tomasfejfar:
Hopefully this should improve logging in datadog (AI-1322).


ncapek:
Can you please explain exactly what you are trying to accomplish? I have no context and it seems to me that the changes in the file are very wide. For example do we really need a dedicated logger only for `dotenv`? That seems like absolute overkill.

ncapek:
> How would you verify it?

If we deploy it to Canary, will that log to Datadog?

vita-stejskal:
Closing this PR in favor of #230.

PR #231, opened 2025-08-12
AI-1342: upgrade libs - mcp==1.17.0 and fastmcp==2.12.4

vita-stejskal:


vita-stejskal:
The `fastmcp~=2.11` library versions seem to have changed the way how HTTP request context is accessed. There is now no way to access the HTTP headers of the initial request to the `POST /sse` endpoint. When the MCP server is started without the OAuth enabled, it uses the initial HTTP request headers to read the SAPI token. This mechanism is now broken which is why the integration test `test_sse_setup[False]` fails. I am trying to find some workaround.

PR #233, opened 2025-08-13
AI-1322: disarm 'FastMCP' logger

vita-stejskal:
Another try to tame the loggers in the MCP server. The change disarms the `FastMCP` logger that's set up by the `fastmcp` library. It's static and initialized from the library's code without any way to customize it. The change keeps the logger, but removes its handlers and set's it to forward messages up to the its parent logger.

### DIFF
--- a/src/keboola_mcp_server/cli.py
+++ b/src/keboola_mcp_server/cli.py
@@ -72,7 +72,11 @@ async def run_server(args: Optional[list[str]] = None) -> None:
         log_config = None
 
     if log_config:
-        logging.config.fileConfig(log_config, disable_existing_loggers=False)
+        logging.config.fileConfig(log_config, disable_existing_loggers=True)
+        # disarm the 'FastMCP' logger; there is no legitimate way to configure this in fastmcp
+        fastmcp_logger = logging.getLogger('FastMCP')
+        fastmcp_logger.handlers.clear()
+        fastmcp_logger.propagate = True
     else:
         logging.basicConfig(
             format='%(asctime)s %(name)s %(levelname)s: %(message)s',
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,4 +1,8 @@
+import json
+import subprocess
+import time
 from dataclasses import asdict
+from pathlib import Path
 from typing import Annotated, Any
 
 import pytest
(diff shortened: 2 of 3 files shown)


tests/test_server.py
```diff
@@ -232,3 +236,34 @@ async def assessed_function(ctx: Context, param: str) -> str:
...
+        ['python', '-m', 'keboola_mcp_server', '--transport', 'sse', '--log-config', log_config_file.absolute()],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    # give the server time to fully start
+    time.sleep(5)
+
+    # kill the server and capture streams
+    p.kill()
```
radcliffkey, line 256: isn't `terminate()` enough?

tests/test_server.py
```diff
@@ -232,3 +236,34 @@ async def assessed_function(ctx: Context, param: str) -> str:
...
         assert isinstance(result.content[0], TextContent)
         assert result.content[0].text == 'value'
+
+
+def test_json_logging(mocker):
+    log_config_file = Path(__file__).parent.parent / 'logging-json.conf'
+    assert log_config_file.is_file(), f'No logging config file found at {log_config_file.absolute()}'
+
+    # start the MCP server process with json logging
+    p = subprocess.Popen(
```
radcliffkey, line 246: Just FYI, there is also an async way for working with subprocesses

https://docs.python.org/3.12/library/asyncio-subprocess.html#asyncio-subprocess

radcliffkey (approved):

src/keboola_mcp_server/cli.py
```diff
@@ -72,7 +72,11 @@ async def run_server(args: Optional[list[str]] = None) -> None:
         log_config = None
 
     if log_config:
-        logging.config.fileConfig(log_config, disable_existing_loggers=False)
+        logging.config.fileConfig(log_config, disable_existing_loggers=True)
```
radcliffkey, line 75: GPT-5 says this doesn't disable potential child loggers from FastMCP, but I don't know if that's a real problem

ncapek (approved):

src/keboola_mcp_server/cli.py
vita-stejskal: I reverted this in d3ccff1.

tests/test_server.py
vita-stejskal: Fixed in 9d443da.

PR #240, opened 2025-08-19
AI-1352: delete 'get_sql_dialect' tool; it's superseded by 'get_project_info' tool

vita-stejskal:


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -418,7 +418,7 @@ async def create_sql_transformation(
 
     # Get the SQL dialect to use the correct transformation ID (Snowflake or BigQuery)
     # This can raise an exception if workspace is not set or different backend than BigQuery or Snowflake is used
-    sql_dialect = await get_sql_dialect(ctx)
+    sql_dialect = await WorkspaceManager.from_state(ctx.session.state).get_sql_dialect()
```
ncapek, line 421: Is there a reason to prefer accessing the sql dialect through this method vs the `get_project_info` tool? Is it just because of the easier setup?

vita-stejskal:
> Same as I wrote to Marian:
> 
> > When adding/removing tools, we should also update the list in README.md (AI can do it very well using TOOLS.md). Or we can agree to drop it and reference TOOLS.md

Fixed in 8a0915b. I personally would replace the `Supported Tools` table with the link to `TOOLS.md`. Originally, the file looked like this, but then it underwent several major overhauls and the tools table was added.

Another approach would be to generate the tools table by `generate_tool_docs.py` which is now part of out `tox`-based checks. That way we could keep the table up-to-date. The tricky part is to make `generate_tool_docs.py` update that part of `README.md` file which is otherwise static.

PR #246, opened 2025-08-25
Make sure app is running with correct user

ujovlado:
Changes:

- Make sure app is running with correct user

---

Application was semi-prepared to run with specific user:
https://github.com/keboola/mcp-server/blob/85ce01d53f75ade2a378577be220ccec34c4e1b8/Dockerfile#L28

This update finalizes the implementation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,14 +22,23 @@ RUN --mount=type=cache,target=/root/.cache/uv     uv pip install ddtrace~=3.0
 
 FROM python:3.12-slim-bookworm
 
+ARG APP_USER_NAME=app
+ARG APP_USER_UID=1000
+ARG APP_USER_GID=1000
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,14 +22,23 @@ RUN --mount=type=cache,target=/root/.cache/uv     uv pip install ddtrace~=3.0
 
 FROM python:3.12-slim-bookworm
 
+ARG APP_USER_NAME=app
+ARG APP_USER_UID=1000
```
ncapek, line 26: Why 1000?

ncapek (approved):

Dockerfile
ujovlado: It's a common practise. Usually the first user added do the system (even if it is via GUI) is with uid/gid 1000.

PR #253, opened 2025-09-03
AI-1484 readme add branch-id support and flow tools info 

mariankrotil:


davidesner (changes requested):
@mariankrotil I suggested few minor changes. I also noticed that there is no guidance on how to setup different remote transports and what are the options. I always need to look it up in code myself. But we can leave that for another PR if you like

PR #255, opened 2025-09-08
Enhanced bucket counting in ListBucketsOutput

jordanrburger:
Added bucket counts by stage in storage.py and corresponding tests in test_storage.py.
Screenshot 2025-09-08 at 2 37 19 PM

### DIFF
--- a/src/keboola_mcp_server/tools/storage.py
+++ b/src/keboola_mcp_server/tools/storage.py
@@ -24,6 +24,10 @@ LOG = logging.getLogger(__name__)
 
 STORAGE_TOOLS_TAG = 'storage'
 
+BUCKET_ID_PARTS = 2
+TABLE_ID_PARTS = 3
+COLUMN_ID_PARTS = 4
+
 
 def add_storage_tools(mcp: KeboolaMcpServer) -> None:
     """Adds tools to the MCP server."""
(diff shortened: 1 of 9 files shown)


src/keboola_mcp_server/tools/storage.py
```diff
@@ -387,7 +394,22 @@ async def list_buckets(ctx: Context) -> ListBucketsOutput:
             bucket = await _combine_buckets(client, links_manager, prod_bucket, next(iter(dev_buckets), None))
             buckets.append(bucket.model_copy(update={'links': None}))  # no links when listing buckets
 
-    return ListBucketsOutput(buckets=buckets, links=[links_manager.get_bucket_dashboard_link()])
+    # Count buckets by stage
+    input_count = sum(1 for bucket in buckets if bucket.stage == 'in')
+    output_count = sum(1 for bucket in buckets if bucket.stage == 'out')
```
mariankrotil, line 399: Do we have only `in` and `out` stage buckets? If so, then we can count only the one group and derive the other by: `len(buckets) - input_count`

mariankrotil (approved):
Integration Tests are failing because `BucketDef` only stores IDs and names for the test buckets. If you want to test additional bucket information, you have to fetch the bucket details from SAPI or create a mocked version in tests to test your functionality.

PR #257, opened 2025-09-09
Ai 1458 optimize change description

ncapek:
This PR changes the `update_description` tool to handle batch updates in a single tool call in an efficient manner.

Update of bucket, table and column in a single call example:
image

Invalid path handling:
image

Large batch update of many columns:
image

### DIFF
--- a/src/keboola_mcp_server/tools/storage.py
+++ b/src/keboola_mcp_server/tools/storage.py
@@ -557,7 +557,6 @@ def _group_updates_by_type(updates: dict[str, str]) -> DescriptionUpdateGroups:
     column_updates_by_table: dict[str, dict[str, str]] = {}
 
     for path, description in updates.items():
-
         parsed = _parse_path(path)
 
         if parsed.item_type == 'bucket':
@@ -660,11 +659,19 @@ async def update_descriptions(
 ]:
     """Updates descriptions for Keboola storage items (buckets, tables, columns)."""
     client = KeboolaClient.from_state(ctx.session.state)
-
-    grouped_updates = _group_updates_by_type(updates)
-
     results: list[UpdateItemResult] = []
+    valid_updates: dict[str, str] = {}
 
+    # Handle invalid paths first and filter valid ones
+    for path, description in updates.items():
+        try:
+            _parse_path(path)
+            valid_updates[path] = description
+        except ValueError as e:
+            results.append(UpdateItemResult(path=path, success=False, error=f'Invalid path format: {e}'))
+
+    # Process valid updates
+    grouped_updates = _group_updates_by_type(valid_updates)
     for bucket_id, description in grouped_updates.bucket_updates.items():
         result = await _update_bucket_description(client, bucket_id, description)
         results.append(result)
--- a/src/keboola_mcp_server/tools/storage.py
+++ b/src/keboola_mcp_server/tools/storage.py
@@ -24,6 +24,10 @@ LOG = logging.getLogger(__name__)
 
 STORAGE_TOOLS_TAG = 'storage'
 
+BUCKET_PATH_PARTS = 2
+TABLE_PATH_PARTS = 3
+COLUMN_PATH_PARTS = 4
+
 
 def add_storage_tools(mcp: KeboolaMcpServer) -> None:
     """Adds tools to the MCP server."""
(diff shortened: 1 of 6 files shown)


src/keboola_mcp_server/tools/storage.py
```diff
@@ -260,11 +264,35 @@ class ListTablesOutput(BaseModel):
...
+
+
+class UpdateDescriptionsOutput(BaseModel):
+    results: list[UpdateItemResult] = Field(description='Results for each update attempt.')
+    total_processed: int = Field(description='Total number of items processed.')
+    successful: int = Field(description='Number of successful updates.')
+    failed: int = Field(description='Number of failed updates.')
+
+
+class ParsedPath(BaseModel):
```
radcliffkey, line 281: this is more of a structured ID than a path

src/keboola_mcp_server/tools/storage.py
```diff
@@ -497,85 +525,166 @@ async def list_tables(
...
+        result = await _update_bucket_description(client, bucket_id, description)
+        results.append(result)
+
+    for table_id, description in grouped_updates.table_updates.items():
+        result = await _update_table_description(client, table_id, description)
+        results.append(result)
+
+    for table_id, column_updates in grouped_updates.column_updates_by_table.items():
+        table_results = await _update_column_descriptions(client, table_id, column_updates)
+        results.extend(table_results)
```
mariankrotil, line 685: We could parallelize the for-loops by asynchronously updating the list in batches. For example, with a batch size of 10, group_bucket would be updated in chunks of 10 elements at a time.

ncapek: I don't think we need this now. As I understand it, the issue was that the agent had to call the tool for each individual change. So it was more of a context/tool calling inefficiency, rather than speed.

src/keboola_mcp_server/tools/storage.py
```diff
@@ -497,85 +525,166 @@ async def list_tables(
...
+            UpdateItemResult(path=f'{table_id}.{column_name}', success=False, error=str(e))
+            for column_name in column_updates.keys()
+        ]
+
+
+@tool_errors()
+async def update_descriptions(
+    ctx: Context,
+    updates: Annotated[
+        dict[str, str],
```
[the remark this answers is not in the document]
ncapek: I can try it out and see how the agent reacts to this.

src/keboola_mcp_server/tools/storage.py
mariankrotil: Check how it is named in the bucket / table / column instances that already include the required param, and use the same naming for consistency.

src/keboola_mcp_server/tools/storage.py
ncapek: The main use case for bulk updates is when a large table with many columns is being updated at once. And columns are updated with a single API call.

src/keboola_mcp_server/tools/storage.py
```diff
@@ -497,85 +525,166 @@ async def list_tables(
...
+    for path, description in updates.items():
+        parsed = _parse_path(path)
+
+        if parsed.item_type == 'bucket':
+            bucket_updates[parsed.bucket_id] = description
+        elif parsed.item_type == 'table':
+            table_updates[parsed.table_id] = description
+        elif parsed.item_type == 'column':
+            if parsed.table_id not in column_updates_by_table:
+                column_updates_by_table[parsed.table_id] = {}
```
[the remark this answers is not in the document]
ncapek: Fixed

src/keboola_mcp_server/tools/storage.py
```diff
@@ -497,85 +525,166 @@ async def list_tables(
...
+    :param path: Path string (e.g., "in.c-bucket", "in.c-bucket.table", "in.c-bucket.table.column")
+    :return: ParsedPath object with structured data
+    """
+    if not path.startswith(('in.', 'out.')):
+        raise ValueError(f'Invalid path format: {path} - must start with in. or out.')
+
+    parts = path.split('.')
+
+    if len(parts) == BUCKET_PATH_PARTS:
+        return ParsedPath(item_type='bucket', bucket_id=path, table_id='', column_name='')
```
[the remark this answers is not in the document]
ncapek: Done

PR #273, opened 2025-09-30
AI-1670 Enable partial config updates

radcliffkey:


ncapek (approved):

radcliffkey:
I have also improved tool and parameter descriptions for `update_config`

src/keboola_mcp_server/tools/components/model.py
```diff
@@ -180,6 +180,46 @@ def from_api_response(cls, api_response: ComponentAPIResponse) -> 'Component':
...
+    """
+    Set or create a parameter value at the specified path.
+
+    Use this operation to:
+    - Update an existing parameter value
+    - Create a new parameter key
+    - Replace a nested parameter value
+    """
+
+    op: Literal['set']  # name 'op' inspired by JSON Patch (https://datatracker.ietf.org/doc/html/rfc6902)
```
vita-stejskal, line 198: Shouldn't the `op` field be set by default to `'set'` which is the only value that's allowed? The same applies for the other `ConfigParam*` models.

src/keboola_mcp_server/tools/components/model.py
```diff
@@ -180,6 +180,46 @@ def from_api_response(cls, api_response: ComponentAPIResponse) -> 'Component':
         )
 
 
+# ============================================================================
+# CONFIGURATION PARAMETER UPDATE MODELS
+# ============================================================================
+
+
+class ConfigParamSet(BaseModel):
```
vita-stejskal, line 188: Can this operation be used for appending an item to a list? Similarly, can `ConfigParamRemove` operation be used for removing an item from a list? If the answers are no, should we add a special operation for partial list updates?

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -824,61 +826,101 @@ async def update_config(
...
+                'Only provide if updating storage mappings - this replaces the ENTIRE storage configuration. '
+                '\n\n'
+                'When to use:\n'
+                '- Adding/removing input or output tables\n'
+                '- Modifying table/file mappings\n'
+                '- Updating table destinations or sources\n'
+                '\n'
+                'Important:\n'
+                '- Not applicable for row-based components (they use row-level storage)\n'
+                "- Must conform to the component's storage schema\n"
```
vita-stejskal, line 884: Can the storage schema be different for different components? I thought that the storage schema is still the same.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -901,11 +943,14 @@ async def update_config(
...
+    if parameter_updates:
+        current_params = configuration_payload.get('parameters', {})
+        updated_params = update_params(current_params, parameter_updates)
+
         parameters_cfg = validate_root_parameters_configuration(
             component=component,
-            parameters=parameters,
-            initial_message='The "parameters" field is not valid.',
+            parameters=updated_params,
+            initial_message='The updated "parameters" field is not valid.',
```
vita-stejskal, line 953: This error message is now misleading/inaccurate. There is no `parameters` field anymore. Maybe you could say something like:

- `The updated configuration parameters are not valid.`
- or `Applying the "parameter_updates" field operations failed.`
- or `Applying the "parameter_updates" field operations resulted in an invalid configuration.`

src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -418,3 +421,101 @@ async def set_cfg_update_metadata(
...
+
+        if not matches:
+            raise ValueError(f'Path "{update.path}" does not exist')
+
+        return jsonpath_expr.filter(lambda x: True, params)
+
+
+def update_params(params: dict[str, Any], updates: Sequence[ConfigParamUpdate]) -> dict[str, Any]:
+    """
+    Apply a list of parameter updates to the given parameters dictionary.
```
vita-stejskal, line 510: Nitpick: `Apply` -> `Applies`

src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -418,3 +421,101 @@ async def set_cfg_update_metadata(
...
+
+        if not matches:
+            raise ValueError(f'Path "{update.path}" does not exist')
+
+        for match in matches:
+            current_value = match.value
+            if not isinstance(current_value, str):
+                raise ValueError(f'Path "{match.full_path}" is not a string')
+
+            new_value = current_value.replace(update.search_for, update.replace_with)
```
vita-stejskal, line 494: Should you check and fail when the `update.search_for` string is not found and the "replace" is effectively a no-op?

src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -418,3 +421,101 @@ async def set_cfg_update_metadata(
...
+            params = match.full_path.update(params, new_value)
+
+        return params
+
+    elif update.op == 'remove':
+        matches = jsonpath_expr.find(params)
+
+        if not matches:
+            raise ValueError(f'Path "{update.path}" does not exist')
```
vita-stejskal, line 503: It's probably safer to raise an exception when the path does not exist, but this makes the `remove` operation non-idempotent, while the other two operations are idempotent AFAICT.

vita-stejskal (approved):
This looks neat!

src/keboola_mcp_server/tools/components/tools.py
radcliffkey: Good catch! Fixed in
https://github.com/keboola/mcp-server/pull/273/commits/ffb4ab6652562e2b3ae7be684abb6d5bb7157063

src/keboola_mcp_server/tools/components/utils.py
radcliffkey: I think it's ok - it's a property of remove operation that you cannot remove the same thing twice and I agree that it's safer to throw an error because removing the same thing twice might indicate that the LLM has a wrong context.

src/keboola_mcp_server/tools/components/model.py
radcliffkey: No, but I think we can survive without it for now. It will probably be more important for transformations which use code block lists.

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -1076,11 +1115,14 @@ async def update_config_row(
...
+    if parameter_updates:
+        current_params = configuration_payload.get('parameters', {})
+        updated_params = update_params(current_params, parameter_updates)
+
         parameters_cfg = validate_row_parameters_configuration(
             component=component,
-            parameters=parameters,
-            initial_message='Field "parameters" is not valid.\n',
+            parameters=updated_params,
+            initial_message='Applying the "parameter_updates" resulted in an invalid configuration.',
```
vita-stejskal, line 1125: Maybe `... an invalid >>row<< configuration`.

vita-stejskal (approved):

src/keboola_mcp_server/tools/components/tools.py
radcliffkey: done in https://github.com/keboola/mcp-server/pull/273/commits/15e808979491a1322d63e28e20821c11de69b002

PR #286, opened 2025-10-16
AI-1794: always create the top level Starlette app with ForwardSlashMiddleware

vita-stejskal:


integtests/test_mcp_server.py
```diff
@@ -24,185 +36,164 @@ async def test_stdio_setup(
...
+        'run_job',
+        # 'search',
+        'update_config',
+        'update_config_row',
+        'update_descriptions',
+        'update_flow',
+        'update_sql_transformation',
+    }
+    actual_tools = {tool.name for tool in tools}
+    missing_tools = expected_tools - actual_tools
```
mariankrotil: We may consider calculating the set difference also in the opposite direction (`actual − expected`) so if we add a new tool, we can detect it.

integtests/test_mcp_server.py
```diff
@@ -1,16 +1,28 @@
...
+    DEV_STORAGE_TOKEN_ENV_VAR,
+    DEV_WORKSPACE_SCHEMA_ENV_VAR,
+    ConfigDef,
+)
 from keboola_mcp_server.tools.components.model import Configuration
-from keboola_mcp_server.workspace import WorkspaceManager
+from keboola_mcp_server.tools.project import ProjectInfo
+
+LOG = logging.getLogger(__name__)
+HttpTransportStr = Literal['sse', 'streamable-http']
```
mariankrotil: Do we want to also support `http-compat` here and test it also as we are using it, or are we planning to deprecate it soon and use only the `streamable-http`?

mariankrotil (approved):

integtests/conftest.py
```diff
@@ -25,18 +20,13 @@
...
-    [Literal['sse', 'streamable-http'], str, dict[str, str] | None], _AsyncGeneratorContextManager[Client]
-]
-
 LOG = logging.getLogger(__name__)
 
-STORAGE_API_TOKEN_ENV_VAR = 'INTEGTEST_STORAGE_TOKEN'
 STORAGE_API_URL_ENV_VAR = 'INTEGTEST_STORAGE_API_URL'
+STORAGE_API_TOKEN_ENV_VAR = 'INTEGTEST_STORAGE_TOKEN'
 WORKSPACE_SCHEMA_ENV_VAR = 'INTEGTEST_WORKSPACE_SCHEMA'
+STORAGE_API_TOKEN_ENV_VAR_2 = 'INTEGTEST_STORAGE_TOKEN_PRJ2'
```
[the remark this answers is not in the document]
vita-stejskal: Fixed in #e8ec959

PR #290, opened 2025-10-21
AI-1618: Add automatic SQL query formatting with sqlglot

davidesner:
# AI-1618: Add automatic SQL query formatting with sqlglot

## Summary
Adds automatic SQL formatting using sqlglot library to improve readability of LLM-generated SQL queries in transformations. Formatting is applied during creation and updates of SQL transformations, with graceful fallback to original SQL if formatting fails. The agent is informed if the code it has generated was reformatted

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.31.0"
+version = "1.32.0"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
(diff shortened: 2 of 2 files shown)


src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -539,7 +549,32 @@ async def update_sql_transformation(
     updated_configuration = current_config.get('configuration', {})
 
     if parameters is not None:
-        updated_configuration['parameters'] = parameters.model_dump(by_alias=True)
+        formatted_blocks = []
```
mariankrotil: We may encapsulate this logic into a single method that could be used in both `create_sql_transformation` and `update_sql_transformation` tools

radcliffkey:
Local testing with Cursor went fine - I was able to create and update transformations and their jobs were successful at the first try

mariankrotil (approved):

vita-stejskal (approved):

PR #292, opened 2025-10-22
AI-1767: set component ID in the emited SAPI events based on the connected MCP client

vita-stejskal:


src/keboola_mcp_server/errors.py
```diff
@@ -15,6 +15,11 @@
 LOG = logging.getLogger(__name__)
 F = TypeVar('F', bound=Callable[..., Any])
 
+_UA_TO_CID: Mapping[str, str] = {
```
radcliffkey: pls add a comment or make the name a bit more descriptive

vita-stejskal: Fixed in c9471a3.

PR #297, opened 2025-10-28
AI-1803: add QueryServiceClient; use Query Service in _SnowflakeWorkspace

vita-stejskal:


src/keboola_mcp_server/workspace.py
```diff
@@ -188,8 +192,59 @@ async def get_table_info(self, table: Mapping[str, Any]) -> DbTableInfo | None:
...
+                token=self._client.token,
+                headers=self._client.headers,
+            )
+
+        ts_start = time.perf_counter()
+        job_id = await self._qsclient.submit_job(statements=[sql_query], workspace_id=str(self.id))
+        while (job_status := await self._qsclient.get_job_status(job_id)) and job_status['status'] not in [
+            'completed',
+            'failed',
+        ]:
```
mariankrotil: We should also handle the status `canceled` so that we don't end up waiting.

src/keboola_mcp_server/clients/query.py
```diff
@@ -0,0 +1,93 @@
...
+
+from keboola_mcp_server.clients import KeboolaServiceClient, RawKeboolaClient
+from keboola_mcp_server.clients.base import JsonDict
+
+
+class QueryServiceClient(KeboolaServiceClient):
+
+    def __init__(self, raw_client: RawKeboolaClient, branch_id: str) -> None:
+        """
+        Creates a QueryServiceClient from a RawKeboolaClient and a branch id and a workspace id.
```
radcliffkey: the workspace ID is not really used

src/keboola_mcp_server/clients/query.py
radcliffkey: Would it be a lot of work to return pydantic objects instead of `JsonDict`? IMHO for new clients it's worth it.

src/keboola_mcp_server/workspace.py
```diff
@@ -188,8 +192,59 @@ async def get_table_info(self, table: Mapping[str, Any]) -> DbTableInfo | None:
...
+                        break
+            if not real_branch_id:
+                raise RuntimeError('Cannot determine the default branch ID')
+
+            self._qsclient = QueryServiceClient.create(
+                root_url=urlunparse(('https', f'query.{self._client.hostname_suffix}', '', '', '', '')),
+                branch_id=real_branch_id,
+                token=self._client.token,
+                headers=self._client.headers,
+            )
```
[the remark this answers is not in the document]
vita-stejskal: I was tempted too at first, but then reconsidered for two reasons:

* The `QueryServiceClient` handles the branch ID differently than the other agents. The Query Service API does not support the symbolic names for the main/default branch.
* The client needs the workspace ID, which is hidden in the `WorkspaceManager`.
* And that takes us to the most important points -- it is the `WorkspaceManager` that is our abstraction for accessing the project's database backend. An instance of `WorkspaceManager` is available to the tools via their `Context` parameter.

So, it seems like a correct approach to encapsulate the use of the Query Service in the `WorkspaceManager`. The `QueryServiceClient`'s implementation is placed next to the other clients, but it's not meant to be used directly outside of the `WorkspaceManager`. I can move the `QueryServiceClient` implementation, if you think it'd help preventing future misunderstanding.

src/keboola_mcp_server/workspace.py
```diff
@@ -188,8 +192,59 @@ async def get_table_info(self, table: Mapping[str, Any]) -> DbTableInfo | None:
...
+        ]:
+            await asyncio.sleep(1)
+            elapsed_time = time.perf_counter() - ts_start
+            if elapsed_time > 300:  # 5 minutes
+                raise RuntimeError(
+                    f'Query execution timed out after {elapsed_time:.2f} seconds: '
+                    f'job_id={job_id}, status={job_status["status"]}'
+                )
+
+        statement_id = cast(list[JsonDict], job_status['statements'])[0]['id']
```
radcliffkey: is it not possible to have multiple statements?

vita-stejskal: Yes, in general, it is possible to submit a job with multiple SQL statements in which case the job's status will report the status of each of the statements. However, the jobs submitted by `_SnowflakeWorkspace` implementation always contain only one statement AFAIK.

src/keboola_mcp_server/workspace.py
```diff
@@ -188,8 +192,59 @@ async def get_table_info(self, table: Mapping[str, Any]) -> DbTableInfo | None:
...
+
+        ts_start = time.perf_counter()
+        job_id = await self._qsclient.submit_job(statements=[sql_query], workspace_id=str(self.id))
+        while (job_status := await self._qsclient.get_job_status(job_id)) and job_status['status'] not in [
+            'completed',
+            'failed',
+        ]:
+            await asyncio.sleep(1)
+            elapsed_time = time.perf_counter() - ts_start
+            if elapsed_time > 300:  # 5 minutes
```
radcliffkey: the time out could be extracted to a constant

src/keboola_mcp_server/workspace.py
vita-stejskal: Good catch, fixed in 654f2b2.

For the context -- the MCP server never cancels jobs that it submitted, but I assume that someone else with proper permissions could cancel the server's jobs and so the server should handle this status correctly.

mariankrotil: Yes, it would need to be done from somewhere else. Maybe also from the UI by the user stopping the job if possible.

vita-stejskal:
> We should discuss how the new endpoint will be reflected in the injected code in the data apps because they also use the query endpoint.

Here is a task to cover this -- [AI-1873](https://linear.app/keboola/issue/AI-1873/mcp-use-queryservice-in-the-code-generated-for-data-apps).

src/keboola_mcp_server/clients/query.py
vita-stejskal: Well, it would not be that much work of course, but I am afraid that I'll just keep it as it is for now. This is a recurring debate -- these clients should not live in the MCP Server, they should live in its own project and be maintained by the API owners.

PR #299, opened 2025-10-29
AI-1838: update 'search' tool to use SAPI endpoints for listing items to search through

vita-stejskal:


TOOLS.md
```diff
@@ -1865,6 +1866,111 @@ EXAMPLES:
...
+- user_input: "Find items named 'daily report' or 'weekly summary'"
+  → name_prefixes=["daily.*report", "weekly.*summary"], item_types=[]
+  → Returns all items matching any of these patterns
+
+- user_input: "Show me all configurations related to Google Analytics"
+  → name_prefixes=["google.*analytics"], item_types=["configuration"]
+  → Returns configurations with matching patterns
+
+CONSIDERATIONS:
+- Searches across id, name, displayName, and description fields
```
mariankrotil: Do we need to search across IDs?

Like the idea to search across descriptions however isn't it slow?

TOOLS.md
```diff
@@ -1865,6 +1866,111 @@ EXAMPLES:
...
+by matching patterns against item ID, name, display name, or description. Returns matching items grouped by type
+with their IDs and metadata.
+
+WHEN TO USE:
+- User asks to "find", "locate", or "search for" something by name
+- User mentions a partial name and you need to find the full item (e.g., "find the customer table")
+- User asks "what tables/configs/flows do I have with X in the name?"
+- You need to discover items before performing operations on them
+- User asks to "list all items with [name] in it"
+- DO NOT use for listing all items of a specific type. Use list_configs, list_tables, list_flows, etc instead.
```
mariankrotil: As a future improvement, we could consider allowing this search function to replace all listing tools when no pattern filters are provided?

TOOLS.md
vita-stejskal: > Do we need to search across IDs?

Yes, I think we do. Obviously, the hash-like generated IDs of component configs are useless for searching, but IDs of tables or buckets often indicate what they contain.

> Like the idea to search across descriptions however isn't it slow?

I don't think so. Using regex patterns should still be pretty fast even on a large text. It could be a bit slower to pull the descriptions from SAPI (vs. not to do so), but it's probably nothing terribly slow either.

src/keboola_mcp_server/tools/search.py
```diff
@@ -31,98 +33,208 @@ def add_search_tools(mcp: FastMCP) -> None:
...
+            _matches_pattern(bucket_id, patterns)
+            or _matches_pattern(bucket_name, patterns)
+            or _matches_pattern(bucket_display_name, patterns)
+            or _matches_pattern(bucket_description, patterns)
+        ):
+            hits.append(
+                SearchHit(
+                    bucket_id=bucket_id,
+                    item_type='bucket',
+                    created=bucket.get('created', datetime.now().isoformat()),
```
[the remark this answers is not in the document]
vita-stejskal: The `SearchHit.created` is used only for sorting the hits and placing the most recently created items at the top of the list. You are right that those without `created` field would constantly be prioritized to the top of the list which could be unexpected. I'll make those appear at the bottom of the list.

OTOH the reality is that all items seem to always have the `created` field filled in. They also seem to have the `updated` field filled in consistently and so it's probably better to use `updated` and fall back to `created` and fall back to `None` (or something very old) for the sorting.

src/keboola_mcp_server/tools/search.py
```diff
@@ -31,98 +33,208 @@ def add_search_tools(mcp: FastMCP) -> None:
...
+        for component in components:
+            if not (component_id := component.get('id')):
+                continue
+
+            item_type: ItemType
+            if component_id in ['keboola.orchestrator', 'keboola.flow']:
+                item_type = 'flow'
+            elif component_type == 'transformation':
+                item_type = 'transformation'
+            elif component_type == 'keboola.sandboxes':
```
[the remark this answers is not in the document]
vita-stejskal: Fixed in a63880f.

src/keboola_mcp_server/tools/search.py
```diff
@@ -203,13 +313,52 @@ async def search(
...
+    if types_to_fetch:
+        all_hits = [item for item in all_hits if item.item_type in types_to_fetch]
+
+    # TODO: Should we sort by the item type too?
+    all_hits.sort(
+        key=lambda x: (x.created, x.bucket_id, x.table_id, x.component_id, x.configuration_id, x.configuration_row_id),
+        reverse=True,
     )
-    return GlobalSearchOutput.from_api_responses(response)
+    paginated_hits = all_hits[offset : offset + limit]
```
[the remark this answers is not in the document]
vita-stejskal: I am afraid that this is not possible. The SAPI endpoint we use (e.g for listing components, buckets, tables, etc) do not support paging at all.

src/keboola_mcp_server/tools/components/utils.py
```diff
@@ -416,6 +418,18 @@ async def set_cfg_update_metadata(
 # ============================================================================
 
 
+def get_nested(obj: Mapping[str, Any] | None, key: str, *, default: T = None) -> T | None:
```
radcliffkey: nitpick: maybe it should be `default: T | None = None`. Also, we can't guarantee that the return type will be `T` but that's fine

src/keboola_mcp_server/tools/search.py
```diff
@@ -31,98 +42,230 @@ def add_search_tools(mcp: FastMCP) -> None:
...
         Field(
-            description='One or more name prefixes to search for. An item matches if its name (or any word in the '
-            'name) starts with any of these prefixes. Case-insensitive. Examples: ["customer"], ["sales", "revenue"], '
-            '["test"]. Do not use empty strings or empty lists.'
+            description='One or more search patterns to match against item ID, name, display name, or description. '
+            'Supports regex patterns. Case-insensitive. Examples: ["customer"], ["sales", "revenue"], '
+            '["test.*table"]. Do not use empty strings or empty lists.'
         ),
     ],
     item_types: Annotated[
```
radcliffkey: the item types contain `workspace`. Are we actually able to search workspaces?

vita-stejskal: Yes, I think so. A workspace is a configuration of the `keboola.sandboxes` component. I haven't tried that. but I think the `search` tool is able to find it and the components tools should be able to fetch/update its config.

vita-stejskal: I confirm that the workspaces are accessible by the search and components tools.

PR #306, opened 2025-11-04
[AI-1693] docs: clarify flow conditions not needed for failures

AdamVyborny:
## Why

The AI agent was over-creating flow conditions, particularly failure conditions and unnecessary success conditions. This added complexity to generated flows without providing value, as phase/task failures already automatically end the flow execution.

## Context

Relates to https://linear.app/keboola/issue/AI-1693

The create_conditional_flow tool description needed clearer guidance about when conditions are actually needed. The previous wording was too generic, leading agents to defensively add conditions for every scenario.

## Approach

Updated the tool description to explicitly state:
- Phase/task failures automatically end the flow - no need for explicit failure conditions
- Success conditions should only be added when users explicitly request branching or conditional logic

## Impact

- Reduces unnecessary complexity in generated flows
- Makes default flow configurations simpler and easier to understand
- Agents will only add conditions when truly needed for branching logic
- No breaking changes - existing flows remain unaffected

---

### DIFF
--- a/integtests/tools/test_data_apps.py
+++ b/integtests/tools/test_data_apps.py
@@ -11,12 +11,13 @@ from keboola_mcp_server.config import Config, MetadataField, ServerRuntimeInfo
 from keboola_mcp_server.server import create_server
 from keboola_mcp_server.tools.data_apps import (
     _DEFAULT_PACKAGES,
-    _QUERY_DATA_FUNCTION_CODE,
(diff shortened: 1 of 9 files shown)


AdamVyborny:
@vita-stejskal dotáhnete to nějak do release, prosím. Nebo musím něco udělat aby procházeli ty integ testy?

PR #311, opened 2025-11-07
Ai 1839 fix registration to anthropic registry

vita-stejskal:


vita-stejskal:
I'll re-do the fix in another PR.

PR #314, opened 2025-11-10
feat: Add PR template with deployment testing part before deployment

Matovidlo:
* Simple PR that contains what has to be tested before each release

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release_notes_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release_notes_template.md
@@ -1,0 +1,75 @@
+# Release Notes
+
+List the key changes in this release using bullet points:
+- New features added (with brief description of what they enable)
+- Enhancements to existing functionality
+- Bug fixes (reference PR/issue numbers)
+- Breaking changes (if any - mark clearly with ⚠️)
+- Dependency updates (library versions)
+
+# Plans for Customer Communication
+
+Describe how and when customers will be notified about this release:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,18 @@
+## Description
+
+## Changes
+
+## Deployment Testing
+
+- [ ] Tested with Cursor
+- [ ] Tested with Windsurf
+- [ ] Tested with Claude
```
vita-stejskal: Is this "Claude Desktop" or claude.ai (the web?

.github/pull_request_template.md
```diff
@@ -0,0 +1,18 @@
...
+- [ ] Tested with In Platform Agent
+- [ ] Tested with RO chat
```
vita-stejskal: How are the chats supposed to be tested prior the deployment?

.github/pull_request_template.md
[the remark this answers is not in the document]
mariankrotil: Thanks for clarifying. I just meant that testing all client integrations (Cursor, Claude.ai, Windsurf) could take quite a while and replacing with testing only one integration would be enough because the mcp server implements the defined protocol which all other clients should also follow. However, if our clients rely on those client integrations, we should test them all as you say, either manually or ideally automatically. Maybe we could add tests to the CI pipeline tests, similar to how we test the built image, though I am not sure yet how to set it up.

mariankrotil: I would write it there explicitly if both

PR #318, opened 2025-11-12
AI-1749: impose limits on the size of data returned from the 'query_data' tool

vita-stejskal:


src/keboola_mcp_server/tools/sql.py
```diff
@@ -14,13 +14,17 @@
...
+MAX_CHARS = 50_000
 
 
 class QueryDataOutput(BaseModel):
     """Output model for SQL query results."""
 
     query_name: str = Field(description='The name of the executed query')
     csv_data: str = Field(description='The retrieved data in CSV format')
+    csv_data_rows: int | None = Field(default=None, description='Number of rows fetched in the "csv_data" field.')
+    total_query_rows: int | None = Field(default=None, description='Total number of rows selected by the SQL query.')
```
radcliffkey: I am still not sure from this description what exactly is the difference between the two fields. I thought that `csv_data_rows` could be lower than `total_query_rows` if there was truncation, but currently it looks like we don't do truncation.

src/keboola_mcp_server/workspace.py
```diff
@@ -67,12 +67,24 @@ class SqlSelectData:
         description='Selected rows, each row is a dictionary of column: value pairs.'
     )
 
+    def value_chars(self) -> int:
+        chars = 0
+        for r in self.rows:
+            for v in r.values():
+                if v:
```
radcliffkey: could the values be other than string? If so, then int `0` would not be counted

tests/tools/test_sql.py
radcliffkey: Can you add a test for the overflow case?

src/keboola_mcp_server/workspace.py
```diff
@@ -198,7 +210,10 @@ async def get_table_info(self, table: Mapping[str, Any]) -> DbTableInfo | None:
 
         return None
 
-    async def execute_query(self, sql_query: str) -> QueryResult:
+    async def execute_query(self, sql_query: str, *, max_rows: int | None = None) -> QueryResult:
+        if max_rows is not None and max_rows < 0:
```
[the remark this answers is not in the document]
vita-stejskal: @mariankrotil What do you mean by "a clamp"?

src/keboola_mcp_server/workspace.py
```diff
@@ -218,25 +233,68 @@ async def execute_query(self, sql_query: str) -> QueryResult:
...
+                rows_to_fetch = min(page_size, remaining)
+            else:
+                rows_to_fetch = page_size
+
+            results = await self._qsclient.get_job_results(
+                job_id,
+                statement_id,
+                offset=offset,
+                limit=max(rows_to_fetch, 100),  # QueryService requires 100 - 10_000
             )
```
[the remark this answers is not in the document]
vita-stejskal: > And In case it exceeds the max limit of QS (10000) we could fetch them in batches using for cycle instead of a while cycle which I am not a fan of.

I don't think that the `for`-loop would be that much different from the current `while`-loop.

> Or, so far we fetched everything without failing, we could preserve such approach though inefficient, however, apply pagination afterwards for the agent which is the main reason i think

Unfortunately, this is not true. The current implementation (prior this change) is incorrect. It only fetches at max 500 rows which is the default value of the page size in the QueryService's API. I did not notice that default value when re-implementing `_SnowflakeWorkspace` to use the QueryService.

tests/tools/test_sql.py
radcliffkey: yes, and the truncation

src/keboola_mcp_server/workspace.py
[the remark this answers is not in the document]
mariankrotil: Sorry for not being more specific. By clamp I meant that it forces a value to stay within a given minimum and maximum range. If the value is below the minimum it returns the minimum, and if it's above the maximum it returns the maximum.
Like this `max(min_value==1, min(max_rows, max_value==1000))`

src/keboola_mcp_server/workspace.py
vita-stejskal: Fixed in 175470ac.

src/keboola_mcp_server/workspace.py
```diff
@@ -67,12 +67,24 @@ class SqlSelectData:
...
 @dataclass(frozen=True)
 class QueryResult:
     status: QueryStatus = Field(description='Status of running the SQL query.')
-    data: SqlSelectData | None = Field(None, description='Data selected by the SQL SELECT query.')
-    message: str | None = Field(None, description='Either an error message or the information from non-SELECT queries.')
+    data: SqlSelectData | None = Field(default=None, description='Data selected by the SQL SELECT query.')
+    message: str | None = Field(
+        default=None, description='Either an error message or the information from non-SELECT queries.'
+    )
+    data_rows: int | None = Field(default=None, description='Number of rows fetched in the "data" field.')
```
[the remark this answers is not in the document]
vita-stejskal: Fixed in ddb18d6c.

PR #319, opened 2025-11-12
AI-1616: add instructions to test the SQL before using it in a transformation

vita-stejskal:


mariankrotil (approved):
Agree with rado's comment

vita-stejskal:
I am closing this PR due to the controversy about its actual effects on the KAI Agent mentioned in the comments and subsequent offline discussions. We should first measure how the portion of transformations that fail due to KAI agent's mistakes in SQL. The feeling is that this portion is actually not that big to justify the extra step of testing the SQL before using it in a transformation.

PR #331, opened 2025-12-05
AI-2165: upgrade mcp==1.22.0, fastmcp==2.13.1 and other libs

vita-stejskal:
These upgrades unfortunately do not fix the security warning -- [Model Context Protocol (MCP) Python SDK does not enable DNS rebinding protection by default](https://github.com/keboola/mcp-server/security/dependabot/5). The fix is present in `mcp==1.23.0` which is incompatible with `fastmcp==2.13.*` versions. We will have to wait for `fastmcp==2.14` or newer.

Plus there is a regression (2524)[https://github.com/jlowin/fastmcp/issues/2524] in `fastmcp==2.13.2` which breaks the tools that use `Context ctx` parameter (i.e. all our tools).

See also:
* https://github.com/jlowin/fastmcp/releases/tag/v2.13.3
* https://github.com/jlowin/fastmcp/issues/2524

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -1778,7 +1778,7 @@ REQUIREMENTS:
 - Only add conditions/retry logic when user explicitly requests branching or error handling
 - All phases must be connected: no dangling phases are allowed
 - The flow must have exactly one entry point (one phase with no incoming transitions)
-- Every phase must either transition to another phase or end the flow
+- Every phase must either transition to another phase or end the flow by having empty next array field.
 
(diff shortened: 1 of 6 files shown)


mariankrotil (approved):
Approving and can we upgrade to `fastmcp 2.13.3` now when the regression is fixed? It seems that the `mcp 1.23.1` version is now used in the [fastmcp/pyproject](https://github.com/jlowin/fastmcp/blob/main/pyproject.toml) dependencies.

PR #338, opened 2025-12-17
AI-2219: extend search to include table column names and descriptions

vita-stejskal:
Enhance the search tool to find tables by matching patterns against column names and column descriptions, making it easier to discover tables based on their schema structure.

- Add _check_column_match() helper to search column names and descriptions
- Update _fetch_tables() to include columns and columnMetadata in API call
- Extend matching logic to check columns alongside existing table fields
- Add parameterized test covering 4 column search scenarios
- Update existing test to match new API call signature
- Ensure tables appear only once even when multiple fields match

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.36.2"
+version = "1.36.3"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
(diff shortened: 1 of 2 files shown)


src/keboola_mcp_server/tools/search.py
```diff
@@ -161,6 +178,7 @@ async def _fetch_tables(client: KeboolaClient, patterns: list[re.Pattern]) -> li
                 or _matches_pattern(table_name, patterns)
                 or _matches_pattern(table_display_name, patterns)
                 or _matches_pattern(table_description, patterns)
+                or _check_column_match(table, patterns)
```
mariankrotil: Just an idea whether we really want to do this every time for the table `item_type` and whether we shouldn't consider another `item_type` like table_column that would run `_check_column_match` if present? In the normal case when we search for a table, isn't this unnecessary?

PR #344, opened 2025-12-23
Ai 981 add flow scheduling tool

mariankrotil:
Wire the schedule tooling to flow tools.
- add, update, remove schedule is handled in the `update_flow` tool
- `get_flows` tool returns number of schedules for each returned flow in the List view.
- `get_flows` tool returns additional field Schedule Output in the Flow Detail view. Each Schedule output contains list of schedule settings and links to the flow schedules page in the ui.

Add integration tests for scheduler workflow

### DIFF
--- a/src/keboola_mcp_server/tools/flow/scheduler_model.py
+++ b/src/keboola_mcp_server/tools/flow/scheduler_model.py
@@ -46,7 +46,7 @@ class ScheduleDetail(BaseModel):
 
     schedule_id: str = Field(
         description='Schedule configuration ID',
-        serialization_alias='schedule_id',
+        serialization_alias='scheduleId',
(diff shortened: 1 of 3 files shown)


src/keboola_mcp_server/tools/flow/scheduler_model.py
```diff
@@ -46,8 +46,8 @@ class ScheduleDetail(BaseModel):
 
     schedule_id: str = Field(
         description='Schedule configuration ID',
-        serialization_alias='scheduleId',
-        validation_alias=AliasChoices('id', 'schedule_id'),
+        serialization_alias='schedule_id',
```
vita-stejskal: This change seems wrong. It should be reverted, I think.

vita-stejskal (approved):

PR #348, opened 2026-01-12
Ai 2161 update search tool to search in configurations

mariankrotil:
## Description
Extended the search tool beyond table/bucket usage references to robust configuration-aware discovery across items (configuration, transformation, flow, data-app, etc.).

Key updates in this branch:

- Added config-based search path reporting via match_scopes in SearchHit.
- Reworked config matching to use JSONPath traversal (including scoped search) and return exact matched paths.
config-based search now always collects all matched paths per item (simpler agent behavior, more complete results).
- Removed return_all_matches and case_sensitive from the search tool API to reduce complexity and ambiguity.
- Improved scope handling for scalar paths (e.g. direct scope like parameters.api.baseUrl) while preserving descendant matching.
- Kept output practical by exposing only most-specific paths in match_scopes (while internal matches keep full detail).
- Updated search tool docs/examples and project system prompt guidance to explicitly cover config-based search usage.

**Linear**: AI-2161-buckets-tables-references

### Change Type  

### Summary

- Update search tool

## Testing

- [x ] Tested with Cursor AI desktop (`Streamable-HTTP` transports)

### Optional testing
- [x] Tested with Cursor AI desktop (all transports)
- [ ] Tested with claude.ai web and `canary-orion` MCP (`SSE` and `Streamable-HTTP`)
- [x] Tested with In Platform Agent on `canary-orion`

## Checklist

- [x] Self-review completed
- [x] Integration tests added/updated (if applicable)

### DIFF
--- a/src/keboola_mcp_server/tools/search.py
+++ b/src/keboola_mcp_server/tools/search.py
@@ -119,7 +119,7 @@ class SearchHit(BaseModel):
     description: str | None = Field(default=None, description='Description of the item.')
     matches: list[PatternMatch] = Field(
         default_factory=list,
-        description='Most specific JSONPath scopes with grouped matched patterns ' '(config-based search only).',
+        description='Most specific JSONPath scopes with grouped matched patterns (config-based search only).',
     )
     links: list[Link] = Field(default_factory=list, description='Links to the item.')
 


src/keboola_mcp_server/tools/search.py
```diff
@@ -474,59 +516,89 @@ async def search(
...
+
+    - user_input: "Show me all configurations/components related to Google Analytics"
+      → `patterns=["google.*analytics"], search_type="textual", mode="regex", item_types=["component"]`
 
-    - user_input: "Show me all configurations related to Google Analytics"
-      → patterns=["google.*analytics"], item_types=["configuration"]
-      → Returns configurations with matching patterns
+    - user_input: "Find storage input mappings referencing specific tables:"
+      → `patterns=["\\"storage\\".*\\"input\\".*:\\s*\\"in\\..*\\.customers\\""], search_type="config-based",
+      mode="regex", item_types=["transformation", "component"]`
```
vita-stejskal: I am skeptical that patterns like this one would work reliably. The configurations can be quite complex and patterns like this one are likely to match many false positives.

Plus the example itself seems weird. It says "... referencing specific tables" and than uses pattern of `\\"in\\..*\\.customers\\"` which would match all `customers` tables in all buckets.

mariankrotil: Yes, this is a bit broad. My intent was to let the agent define patterns that match certain parts of the configuration, but it would probably be better to require those patterns to be wrapped in a `{}` JSON structure to limit their reach.

Do you suggest we should remove pattern examples altogether and rely on scope to navigate within the config instead? Or should we keep them, but provide regex examples that are explicitly JSON-structured?

src/keboola_mcp_server/tools/search.py
```diff
@@ -198,8 +216,21 @@ def _validate_item_types(self) -> 'SearchSpec':
             self.item_types = list({*self.item_types, 'configuration', 'configuration-row'})
         return self
 
+    @model_validator(mode='after')
+    def _compile_jsonpath_exprs(self) -> 'SearchSpec':
+        # Compile commonly used expressions once per SearchSpec instance.
+        self._all_nodes_expr = jsonpath_ng.parse('$..*')
+        self._scope_exprs = []
+        for scope in self.search_scopes:
+            normalized = scope if scope.startswith('$') else f'$.{scope}'
```
vita-stejskal: I think that you should use the new `_normalize_jsonpath()` function from [PR#394](https://github.com/keboola/mcp-server/pull/394/changes#diff-2d4ce44a45364b40cc88a01e5ebf88709110388a02b3f94a8633a5e95a185063R432) to handle the `#`-prefixed keys properly. Without that the `jsonpath_ng.parse()` is likely to fail for paths such as `authorization.#apiKey`.

src/keboola_mcp_server/tools/search.py
```diff
@@ -227,25 +258,76 @@ def match_patterns(self, value: str | JsonDict | None) -> list[str]:
...
+                    PatternMatch(
+                        scope=re.sub(r'\.\[', '[', str(jpath_match.full_path)),
+                        patterns=matched,
+                    )
+                )
+                if not self.return_all_matched_patterns:
+                    return matches
+        return matches
+
+    def _find_scalar_matches_for_expr(self, configuration: JsonDict, parsed_expr: JSONPath) -> list[PatternMatch]:
```
vita-stejskal: This function is pretty much the same as `_find_matches_for_expr()` function. They could easily be collapsed to a single function with an extra "search_subtree" parameter (or "scalars_only" parameter or something like that).

vita-stejskal:
This looks much better now. I think that we are nearly done. Thanks!

src/keboola_mcp_server/tools/search.py
```diff
@@ -163,6 +179,8 @@ class SearchSpec(BaseModel):
     _component_types: Sequence[str] = PrivateAttr(default_factory=tuple)
     _compiled_patterns: list[re.Pattern] = PrivateAttr(default_factory=list)
     _clean_patterns: list[str] = PrivateAttr(default_factory=list)
+    _all_nodes_expr: JSONPath | None = PrivateAttr(default=None)
+    _scope_exprs: list[tuple[str, JSONPath, JSONPath]] = PrivateAttr(default_factory=list)
```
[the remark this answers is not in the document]
mariankrotil: expained in 241fbb20e2538f878cab233555367ad3c7594d3f

PR #385, opened 2026-02-10
Ai 2348 fix non existing branchid

mariankrotil:
## Description

**Linear**: AI-2348

### Change Type

### Summary

Improved error handling when a non-existent branch ID is configured. Previously, `with_branch_id()` raised a generic `httpx.HTTPStatusError` on 404, making it hard for users to understand what went wrong. Additionally, clients could not discover available tools at all if the branch ID was invalid.

**Changes:**
- `KeboolaClient.with_branch_id()` now raises a clear `ValueError` with a user-friendly message on 404, explicitly stating the branch was not found and prompting the user to verify it
- Added `validate: bool = True` parameter to `with_branch_id()` — when `False`, skips the API call and returns a client with the branch ID as-is (no network round-trip, no normalization)
- `SessionStateMiddleware` now only validates the branch ID for `tools/call` requests; all other methods (`tools/list`, `resources/list`, etc.) skip validation so clients can still discover available tools even when the configured branch ID doesn't exist
- Simplified `create_session_state` — single code path using `with_branch_id(validate=)`; no duplicate `KeboolaClient` construction
- Parametrized the branch validation test to cover both `validate=True` (raises on 404) and `validate=False` (skips API call)

## Testing

### Optional testing
- [ ] Tested with claude.ai web and `canary-orion` MCP (`SSE` and `Streamable-HTTP`)

## Checklist

- [x] Self-review completed
- [ ] Project version bumped according to the change type (if applicable)

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -724,18 +724,6 @@ WORKFLOW:
         "type": "object"
       },
       "type": "array"
-    },
-    "is_disabled": {
-      "anyOf": [
-        {
-          "type": "boolean"
-        },
-        {
-          "type": "null"
-        }
-      ],
-      "default": null,
-      "description": "Enable or disable the configuration. Set to True to disable execution (configuration won't run), False to enable execution (configuration will run). Only provide if changing the status, leave as null to preserve current state."
     }
   },
   "required": [
@@ -1012,7 +1000,6 @@ WHEN TO USE:
 - Updating transformation block or code block names
 - Changing input/output table mappings for the transformation
 - Updating the transformation name or description
-- Enabling or disabling the transformation
 - Any combination of the above
 
 PREREQUISITES:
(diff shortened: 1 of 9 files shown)


src/keboola_mcp_server/clients/client.py
```diff
@@ -81,9 +81,24 @@ def from_state(cls, state: Mapping[str, Any]) -> 'KeboolaClient':
...
-            return self
+    async def with_branch_id(self, branch_id: str | None, validate: bool = True) -> 'KeboolaClient':
+        """Return a new client configured for the given branch.
+
+        :param branch_id: Branch ID to use. None means the main/production branch.
+        :param validate: If True (default), verifies the branch exists via an API call and normalizes
+            the default-branch ID to None. Set to False to skip the network call and use the branch ID
+            as-is (e.g. for tools/list requests where tools should be discoverable even with an invalid branch).
+        """
+        if not validate:
```
vita-stejskal: I think that this implementation is wrong. Regardless of the `validate` flag we should always correctly normalize the `branch_id` to `None` for default branches.

IMO the previous version of this function was correct. It just did not treat well the `404 Not Found` errors. That's the only thing I'd improve here.

PR #464, opened 2026-04-10
fix: return real bucket/table IDs on dev branches (prefix-based model)

padak:
## Summary

On projects using the legacy prefix-based branch model (non-branched-storage), `get_buckets`, `get_table`, and `get_tables` were normalizing bucket/table IDs by stripping the branch prefix (e.g. `out.c-15865-test` became `out.c-test`). These normalized IDs don't physically exist in the Storage API and break all downstream tool calls that reference them.

**Root cause:** `_combine_buckets`, `_get_table`, and `_list_tables` in `tools/storage/tools.py` were replacing the real `id` field with `prod_id` (the stripped version) when returning results for dev-branch-only buckets/tables.

**Fix:** Stop overwriting `id` with `prod_id` in the output. The `prod_id` field is preserved for internal deduplication logic (dict key in `_list_tables` for merging prod+dev tables of the same logical table), but the `id` field returned to the caller is now always the real, physical Storage API ID.

### Changes

- `_combine_buckets` (dev-only bucket case): removed `'id': dev_bucket.prod_id` from `model_copy` update
- `_get_table`: removed `'id': table.prod_id` from `model_copy` update
- `_list_tables` (dev bucket tables): removed `'id': table.prod_id` from `model_copy` update
- Updated 7 test expected values to use real (prefixed) IDs instead of stripped ones

Fixes keboola/keboola_agent_cli#113

## Test plan

- [x] All 51 existing tests in `tests/tools/storage/test_tools.py` pass
- [x] `test_get_bucket` dev-only case now expects real prefixed ID
- [x] `test_get_buckets` dev branch case now expects real prefixed ID
- [x] `test_get_table` dev branch cases (3 parametrizations) now expect real prefixed IDs
- [x] `test_get_tables` dev branch case now expects real prefixed IDs for both emails and assets


padak:
## Update: Branched Storage Support Added

Added a second commit that extends the fix to also handle the **branched storage model** (projects where dev branches don't use prefixed bucket/table IDs).

### What changed

**1. `_get_table` now handles branched storage correctly**

The old logic had a bug for branched storage:
1. `_get_table('out.c-results.output')` on a dev branch would find the table, see it has `KBC.createdBy.branch.id` metadata, and **discard it** ("not a prod table")
2. Then try the prefixed fallback `out.c-9999999-results.output` — which doesn't exist in branched storage
3. Result: **table not found** for a table that clearly exists

The fix mirrors how `_find_buckets` already works — when the first lookup returns a table whose `branch_id` matches `client.branch_id`, treat it as the dev table directly:

```python
# Before (discards matching dev table):
if branch_id:
    prod_table = None  # gone forever

# After (saves matching dev table):
if branch_id:
    if branch_id == client.branch_id:
        dev_table = first_result  # branched storage: keep it
```

**Bonus**: this also eliminates a redundant API call when querying with a prefixed dev table ID (e.g., `in.c-1246948-foo.emails`) — previously it would call `table_detail` twice with the same ID.

**2. Six new branched storage tests**

| Test | What it verifies |
|---|---|
| `test_get_bucket_branched_storage` (x2) | Dev-only bucket + prod bucket detail on branched storage branch |
| `test_get_buckets_branched_storage` | Listing all buckets, correct counts |
| `test_get_table_branched_storage` (x2) | **Dev table with metadata** (fails on main!) + prod table on branch |
| `test_get_tables_branched_storage` | Listing tables in dev-only bucket |

### Verified: fails on main, passes on fix

```
$ pytest -k "branched_storage" (on main branch)
PASSED  test_get_bucket_branched_storage[out.c-results]
PASSED  test_get_bucket_branched_storage[in.c-data]
PASSED  test_get_buckets_branched_storage
FAILED  test_get_table_branched_storage[out.c-results.output]  ← the key one
PASSED  test_get_table_branched_storage[in.c-data.customers]
PASSED  test_get_tables_branched_storage

$ pytest (on fix branch)
57 passed ✓
1139 passed (full suite) ✓
```

tests/tools/storage/test_tools.py
```diff
@@ -1758,3 +1759,490 @@ async def test_get_buckets_use_serializer(mocker):
...
+# have the SAME IDs as production — no prefix like "c-{branch_id}-". They are
+# distinguished only by KBC.createdBy.branch.id metadata.
+# =====================================================================================
+
+BRANCHED_STORAGE_BRANCH_ID = '9999999'
+
+
+def _get_sapi_buckets_branched_storage() -> list[dict[str, Any]]:
+    """Mock SAPI buckets for the branched storage model (no prefix-based IDs)."""
+    return [
```
Matovidlo, line 1777: This is all mocked, can we create test on top of actual project with storage branches feature?

padak: Added integration tests in `tests/tools/storage/test_integration_branches.py` that run against a real Keboola project with branched storage OFF (prefix-based model).

**5 tests:**
- `test_get_bucket_by_id_returns_same_id` — fetches bucket by real ID, verifies it's returned unchanged
- `test_list_all_buckets_returns_real_ids` — lists all buckets, verifies dev-branch bucket has real ID
- `test_returned_id_is_api_callable` — uses returned ID in `bucket_detail()` call to prove it works
- `test_get_tables_returns_real_id` — same for tables
- `test_returned_table_id_is_api_callable` — same for tables

**How to run:**
```bash
TEST_STORAGE_TOKEN=xxx TEST_STORAGE_API_URL=https://connection.keboola.com \
    pytest tests/tools/storage/test_integration_branches.py -v
```

Tests are skipped automatically when env vars are not set (safe for CI without credentials). Each test creates its own branch + bucket + table and cleans up after itself.

All 5 pass against `connection.keboola.com` with a project that has branched storage disabled.

Matovidlo: > All 5 pass against connection.keboola.com with a project that has branched storage disabled.

I am sorry but should not it work for both due to gradual rollout that is currently happening? So with branched storage enabled it should work as well as without branched storage featured enabled (the tests that you have created should be working only on disabled based on  the comment)

tests/tools/storage/test_integration_branches.py
```diff
@@ -0,0 +1,237 @@
...
+async def prod_client():
+    """Create a real KeboolaClient pointing at production."""
+    return KeboolaClient(
+        storage_api_url=STORAGE_API_URL,
+        storage_api_token=STORAGE_TOKEN,
+    )
+
+
+@pytest_asyncio.fixture
+async def dev_branch(prod_client):
```
Matovidlo: I miss some check of token verify output of feature that dev_branches on top of `storage-branches` feature works differently than these one + both features are covered + filtration of projects is covered.

tests/tools/storage/test_tools.py
padak: You're right — both modes need to be tested. The current integration tests cover the prefix-based model (branched storage OFF) because that's where the bug is.

For branched storage ON, the `prod_id == id` (no prefix to strip), so the bug doesn't manifest — but we should still test that the code works correctly there too.

I'll extend the tests to support both modes. The test can detect the mode by checking if bucket IDs on the dev branch have the `c-{branch_id}-` prefix pattern. Or we can parametrize with two different projects/tokens.

Regarding `token verify` feature detection — do you mean we should check `owner.features` for a specific feature flag that indicates branched storage? What's the feature name? I see `ProjectFeature = Literal['global-search']` in the codebase but no `storage-branches` feature. Could you point me to the right feature flag name?

padak:
Updated integration tests per @Matovidlo's feedback:

**Now tests both branch storage models:**
- `[prefix-based]` — project with `storage-branches` feature OFF (bucket IDs have `c-{branch_id}-` prefix)
- `[branched-storage]` — project with `storage-branches` feature ON (bucket IDs identical to production)

**Feature detection:** Each test run calls `tokens/verify` and checks `owner.features` for `storage-branches`. If the project doesn't match the expected mode, the test is skipped with a clear message.

**10 tests total** (5 per mode), all passing:
```
test_get_bucket_by_id_returns_same_id[prefix-based]       PASSED
test_get_bucket_by_id_returns_same_id[branched-storage]    PASSED
test_list_all_buckets_includes_dev_bucket[prefix-based]    PASSED
test_list_all_buckets_includes_dev_bucket[branched-storage] PASSED
test_returned_id_is_api_callable[prefix-based]             PASSED
test_returned_id_is_api_callable[branched-storage]          PASSED
test_get_tables_returns_real_id[prefix-based]               PASSED
test_get_tables_returns_real_id[branched-storage]            PASSED
test_returned_table_id_is_api_callable[prefix-based]        PASSED
test_returned_table_id_is_api_callable[branched-storage]    PASSED
```

**Env vars:**
```bash
TEST_KBC_TOKEN_PREFIX=xxx    # project with branched storage OFF
TEST_KBC_TOKEN_BRANCHED=yyy  # project with branched storage ON
TEST_KBC_URL=https://connection.keboola.com  # optional, defaults to this
```

Tests are skipped automatically when env vars are not set.

tests/tools/storage/test_integration_branches.py
```diff
@@ -0,0 +1,250 @@
...
+
+from keboola_mcp_server.clients.client import KeboolaClient
+from keboola_mcp_server.config import Config, ServerRuntimeInfo
+from keboola_mcp_server.mcp import CONVERSATION_ID, ServerState
+from keboola_mcp_server.tools.storage.tools import get_buckets, get_tables
+from keboola_mcp_server.workspace import WorkspaceManager
+
+TOKEN_PREFIX = os.environ.get('TEST_KBC_TOKEN_PREFIX')
+TOKEN_BRANCHED = os.environ.get('TEST_KBC_TOKEN_BRANCHED')
+STORAGE_URL = os.environ.get('TEST_KBC_URL', 'https://connection.keboola.com')
```
Matovidlo: This should not be needed there should be INTEGTEST URL, check documentation of integtest

tests/tools/storage/test_integration_branches.py
```diff
@@ -0,0 +1,250 @@
...
+
+async def _verify_feature(client: KeboolaClient, expect_branched: bool) -> None:
+    """Assert the project has (or lacks) the storage-branches feature."""
+    token_info = await client.storage_client.get(endpoint='tokens/verify')
+    features = token_info.get('owner', {}).get('features', [])
+    has_branched = 'storage-branches' in features
+    if expect_branched and not has_branched:
+        pytest.skip('Project does not have storage-branches feature (expected ON)')
+    if not expect_branched and has_branched:
+        pytest.skip('Project has storage-branches feature (expected OFF)')
```
Matovidlo: I am not sure this is properly handled. If I have project A with storage branches off and expected is `on` then I would skip the test, then if I select proejct B with storage branches on and expected is `off` I would skip test, in case i pick up the projects in proper order it works. I need to run each of these tests on project with enrolled feature and non enrolled feature. Skip essentially skips the workflow and never test it undeer some circumstances.

Typicall solution is to do project filtering upfront, e.g you have `INTEGTEST_STORAGE_TOKENS` which means that some of the tokens could contain the feature and some other not. In case the token contains one feature or another it does particular test. There should be no skipping because it would not run the test

Matovidlo (changes requested):

tests/tools/storage/test_integration_branches.py
padak: Fixed — moved tests to `integtests/tools/test_storage_branch_ids.py` and refactored to use the standard `INTEGTEST_STORAGE_TOKENS` pool:

**Changes:**
- Removed standalone `test_integration_branches.py` from `tests/` (no custom env vars)
- New file: `integtests/tools/test_storage_branch_ids.py`
- Uses `storage_api_url` and `storage_api_token` fixtures from the integtest pool
- Detects `storage-branches` feature via `tokens/verify` → `owner.features` to classify the project model
- **No skipping**: tests run against whatever project the pool provides, and the test output includes `[prefix-based]` or `[branched-storage]` in the assertion messages so you know which model was tested
- For full coverage, the pool should contain projects with both models

The test flow per run:
1. Detect branch model from token features
2. Create dev branch (poll async job)
3. Create bucket + table on branch (poll async jobs)
4. Call `get_buckets`/`get_tables` MCP tools
5. Assert returned IDs are real physical IDs (not stripped)
6. Verify returned IDs work in subsequent `bucket_detail`/`table_detail` API calls
7. Cleanup branch

padak:
I've addressed the feedback — moved tests to `integtests/`, adopted the `INTEGTEST_STORAGE_TOKENS` pool pattern, added feature detection via `tokens/verify`, and removed all skipping logic.

I just noticed the `CLAUDE.md` in the repo — should have read it earlier, apologies. I see the conventions now (Linear issue ID in branch/commits, version bump in every PR, tox for testing, PR template). Since I'm an external contributor without a Linear issue, some of these don't directly apply, but I understand the structure better now.

A few things I'm not sure how to handle from outside your team:
- **Version bump**: should I bump `pyproject.toml` or do you prefer to do that on merge?
- **Linear ID**: I don't have one for this fix — is there an existing ticket I should reference?
- **tox**: I tested locally with pytest against two real projects (prefix-based + branched-storage, 10/10 pass). Happy to run tox if you point me to the right venv setup.

The core fix is 3 one-line changes in `tools.py` (stop overwriting `id` with `prod_id`). Verified against real projects in both branch models. Let me know what else you need to get this across the line.

Matovidlo (approved):
LGTM

davidesner:
@Matovidlo this seems to remove the intentional functionality of bucket deference https://keboola.atlassian.net/browse/AI-1386 Which will cause the agent to put branched table ids into input output mappings and break things. We were intentionally keeping the bucket id pointing to the production and swapping only the FQN so it can be still queried but from the point of agent it looks like the normal table. There may be a better solution, by keeping that concept of branched tables and instruct agent to never IM / OM the branched versions. What has been impleneted in this PR seems now only a regression.

Also the main change was in the API where table/bucket operations on main don't include branched objects anymore and branched version of the api now works. e.g. branch/123/buckets will return what's in the branch. It also seems that the branched call won't return objects that were not touched yet so there will need to be some logic that merges main and branch storage call. Since this PR doesn't touch the API part it won't fix anything and all branched tables will not be available in list commands

Matovidlo:
@padak please check David comment and try to address it. @davidesner please do regular review, do not tag me, Padak opened the PR so he is responsible to address it.

davidesner:
@Matovidlo ok, I just assumed that you own this since you are owner of the task... consider it regular review

davidesner:
FYI I am implementing it from scratch and will make a new PR to merge into this one

davidesner:
The new PR is here https://github.com/keboola/mcp-server/pull/477 I don't think it makes sense to merge it into this one since it is completely reverting all changes made in this one. I run ai reviews now on the new pr

tests/tools/storage/test_tools.py
davidesner: @padak this is not how the api behaves no? Branched storage api returns ONLY buckets that were touched in the branch. Thats why even UI needs to call both main and branched versions and merge the results.

davidesner:
I am closing this, as it's resolved by PR #477

PR #482, opened 2026-04-17
fix: docker smoke test project_id extraction and missing CI variables

Matovidlo:
## Description

**Linear**: AI-XXX

### Change Type  

### Summary

The release CI `docker-smoke-test` job was failing with `Error: Process completed with exit code 1` immediately after printing `✓ Tool call succeeded`. Two bugs:

**1. Wrong grep pattern in `tests/docker/ci.sh`**

`get_project_info` returns a Pydantic model serialised as a JSON string inside the SSE `text` field, so `project_id` appears as `\"project_id\": \"2729\"` (backslash-escaped, quoted string value). The old pattern `grep -oP "\"project_id\":\s*\d+"` never matched because:
- It expected literal `"` but the body contains `\"`
- It expected a bare integer but the value is a quoted string

With `set -Eeuo pipefail` at the top of the script, the failing grep silently killed the process before the error branch could print anything.

**2. Stale variable names in `release.yml`**

The `Run simple e2e test` step referenced `INTEGTEST_STORAGE_TOKEN`, `INTEGTEST_STORAGE_API_URL`, and `INTEGTEST_WORKSPACE_SCHEMA` (all singular/old names). These are no longer configured; the repo now uses the plural pool variables. This caused `storage_token=None` in the container, making every tool call return an auth error while still returning HTTP 200.

**Fixes:**
- Replace grep with `jq` to correctly parse the nested/escaped JSON response; add `|| true` to survive parse failures within the `set -e` shell
- Add `:?` early-exit guards in `ci.sh` so a missing token/URL fails fast with a clear message instead of silently looping 30 times
- Update `release.yml` to use `INTEGTEST_STORAGE_TOKENS`, `INTEGTEST_POOL_STORAGE_API_URL`, and `INTEGTEST_WORKSPACE_SCHEMAS` (extracting the first entry from each space-separated list); pass secrets via `env:` rather than inline expressions

## Testing

### Optional testing

## Checklist

- [x] Self-review completed

### DIFF
--- a/tests/docker/ci.sh
+++ b/tests/docker/ci.sh
@@ -66,10 +66,11 @@ main() {
                 if [ "$http_code" = "200" ] && [ -n "$body" ]; then
                     echo "✓ Tool call succeeded"
 
-                    project_id=$(echo "$body" | grep "^data: " | grep -oP "\"project_id\":\s*\d+" | cut -d: -f2)
+                    project_id=$(echo "$body" | grep "^data: " | head -1 | cut -c7- \
+                        | jq -r '.result.content[0].text | fromjson | .project_id' 2>/dev/null || true)
                     token_project_id=$(echo "$STORAGE_API_TOKEN" | cut -d- -f1)
 
-                    if [ "$project_id" = "$token_project_id" ]; then
+                    if [ -n "$project_id" ] && [ "$project_id" = "$token_project_id" ]; then
                         echo "✓ Project ID test passed"
                         exit 0
                     else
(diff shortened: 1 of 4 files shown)


Matovidlo:
@claude review

tests/docker/ci.sh
```diff
@@ -11,6 +11,10 @@ cleanup() {
 trap cleanup EXIT
```
tomasfejfar, line 11: I think it would be better to not use bash for this 🙈

tomasfejfar (approved):
LGTM

tests/docker/ci.sh
Matovidlo: huge kudos but for now fix it :smile:

PR #492, opened 2026-04-23
AI-3064: restore SQL formatting for readable diff previews

Matovidlo:
## Description

**Linear**: [AI-3064](https://linear.app/keboola/issue/AI-3064/updating-a-transformation-does-not-show-diff-preview)

### Change Type

### Summary

**Root cause**: PR #433 (AI-2773) removed all SQL formatting from transformation write operations to fix a semantic mutation bug where `sqlglot.transpile` was changing `MM→mm`, `!=→<>`, and `TO_VARCHAR→TO_CHAR`. After that removal, LLM-generated SQL was stored as a single long line, making the `POST /preview/config-diff` platform diff preview unreadable.

**Fix**: The semantic issue was caused by `sqlglot.transpile` using `read=dialect` but no `write=dialect`, causing output in sqlglot's generic SQL dialect. Adding `write=dialect.lower()` makes it output back into the same input dialect, preserving all Snowflake/BigQuery-specific syntax.

Changes:
- `sql_utils.py`: Added `write=dialect.lower()` to `sqlglot.transpile` in `format_sql` — this is the key fix preventing semantic drift
- `tf_update.py`: Restored `format_simplified_tf_block`, `format_simplified_tf_code`, and `format_sql` calls in `add_block`, `add_code`, `set_code` operations; also fixed `add_script` to use `\n` separator instead of space
- `utils.py`: Restored `format_simplified_tf_block` call in `create_transformation_configuration`
- Tests updated to reflect the now-formatted SQL in expected values
- Version bumped `1.57.1 → 1.57.2`

## Testing

## Checklist

- [x] Self-review completed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.57.1"
+version = "1.57.2"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/src/keboola_mcp_server/tools/components/sql_utils.py
+++ b/src/keboola_mcp_server/tools/components/sql_utils.py
@@ -129,7 +129,8 @@ def format_sql(sql: str, dialect: str) -> str:
     """
     try:
         # transpile returns a list - one item per statement/comment
-        formatted_items = sqlglot.transpile(sql, read=dialect.lower(), pretty=True)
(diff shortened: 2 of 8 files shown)


davidesner:
@Matovidlo can we close this? I don't think it's necessary or related to the (now closed) issue. If agent generates single line code still, it should be just prompted to format it properly rather than forcing formatting. Every user may have different formatting requirements and one of the reasons we removed it was that users were trying to convince Kai to change formatting and wasn't able to do so. Impact on input tokens is minimal and imo worth the flexibility.

PR #497, opened 2026-04-28
fix(workspace): extend BigQuery query timeout to 5 minutes, update sql transformation folder error handling

Matovidlo:
## Description

**Linear**: AI-3076

### Change Type

### Summary

BigQuery queries run via `query_data` were timing out after ~60 seconds with "Upstream request timed out, please retry". The root cause: BigQuery workspace execution uses the legacy Storage API (`POST /branch/{id}/workspaces/{id}/query`) as a single synchronous HTTP call, subject to the default 60-second `httpx` read timeout. Snowflake is unaffected — it uses the async Query Service poll loop.

**Fix:** Threads an optional per-call `timeout` override down through three layers so only the BigQuery workspace query gets a 5-minute read timeout; all other storage API calls keep the 60-second default.

Changes:
- `clients/base.py` — `RawKeboolaClient.post()` and `KeboolaServiceClient.post()` both accept an optional `timeout: httpx.Timeout | None` that overrides the client default for that single call; `KeboolaServiceClient.post()` forwards it to `raw_client.post()`
- `clients/storage.py` — `workspace_query()` threads the `timeout` through to `post()`
- `workspace.py` — `_BigQueryWorkspace.execute_query()` passes `httpx.Timeout(connect=5.0, read=_QUERY_TIMEOUT, write=10.0, pool=5.0)` (reusing the existing `_QUERY_TIMEOUT = 300.0` constant)
- `tests/tools/test_sql.py` — asserts `workspace_query` is called with the correct timeout, referencing `_BigQueryWorkspace._QUERY_TIMEOUT` rather than a hard-coded value

## Testing

### Optional testing
- [x] Tested with claude.ai web and `canary-orion` MCP (`Streamable-HTTP`)

## Checklist

- [x] Self-review completed

### DIFF
--- a/src/keboola_mcp_server/clients/storage.py
+++ b/src/keboola_mcp_server/clients/storage.py
@@ -3,6 +3,7 @@ import math
 from datetime import datetime
 from typing import Any, Iterable, Literal, Mapping, Optional, Sequence, cast
 
+import httpx
 from pydantic import AliasChoices, BaseModel, Field, field_validator
 
 from keboola_mcp_server.clients.base import JsonDict, KeboolaServiceClient, RawKeboolaClient
@@ -988,12 +989,13 @@ class AsyncStorageClient(KeboolaServiceClient):
     #  and replaced by QueryService.
     #  Unfortunately the QueryService supports only Snowflake backends. We use it in _SnowflakeWorkspace implementation,
     #  but not in _BigQueryWorkspace implementation, which still uses this function.
-    async def workspace_query(self, workspace_id: int, query: str) -> JsonDict:
+    async def workspace_query(self, workspace_id: int, query: str, timeout: httpx.Timeout | None = None) -> JsonDict:
         """
         Executes a query in a given workspace.
 
         :param workspace_id: The id of the workspace
         :param query: The query to execute
+        :param timeout: Optional per-call timeout override; falls back to the client default when None
         :return: The SAPI call response - query result or raise an error.
         """
         return cast(
@@ -1001,6 +1003,7 @@ class AsyncStorageClient(KeboolaServiceClient):
             await self.post(
                 endpoint=f'branch/{self._branch_id}/workspaces/{workspace_id}/query',
                 data={'query': query},
+                timeout=timeout,
             ),
         )
 
--- a/src/keboola_mcp_server/tools/components/tools.py
+++ b/src/keboola_mcp_server/tools/components/tools.py
@@ -871,9 +871,25 @@ async def update_sql_transformation(
     else:
         folder_stripped = folder.strip()
         if folder_stripped:
-            await set_configuration_folder_metadata(client, sql_transformation_id, configuration_id, folder_stripped)
+            try:
+                await set_configuration_folder_metadata(client, sql_transformation_id, configuration_id, folder_stripped)
+            except Exception as exc:
+                LOG.warning(
+                    'Unable to set folder metadata for component "%s", configuration "%s".',
+                    sql_transformation_id,
+                    configuration_id,
+                    exc_info=exc,
+                )
         else:
-            await clear_configuration_folder_metadata(client, sql_transformation_id, configuration_id)
+            try:
+                await clear_configuration_folder_metadata(client, sql_transformation_id, configuration_id)
+            except Exception as exc:
+                LOG.warning(
+                    'Unable to clear folder metadata for component "%s", configuration "%s".',
+                    sql_transformation_id,
+                    configuration_id,
+                    exc_info=exc,
+                )
 
     change_summary = ' '.join(filter(None, [msg, folder_hint])) or None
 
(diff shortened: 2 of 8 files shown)


Matovidlo:
@claude review

src/keboola_mcp_server/clients/storage.py
```diff
@@ -988,19 +989,21 @@ async def workspace_detail(self, workspace_id: int) -> JsonDict:
...
+        :param timeout: Optional per-call timeout override; falls back to the client default when None
         :return: The SAPI call response - query result or raise an error.
         """
         return cast(
             JsonDict,
             await self.post(
                 endpoint=f'branch/{self._branch_id}/workspaces/{workspace_id}/query',
                 data={'query': query},
+                timeout=timeout,
             ),
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit f7d714ee13e76b95e2291a9f0c3a8a90e7ebfb68

Good catch — `AsyncStorageClient` inherits `KeboolaServiceClient.post()`, which was not forwarding the `timeout` kwarg to `raw_client.post()`, causing a `TypeError` at runtime. Fixed by adding `timeout: Optional[httpx.Timeout] = None` to `KeboolaServiceClient.post()` and forwarding it via `self.raw_client.post(..., timeout=timeout)`. The existing `test_cancel_job_method` assertion was also updated to expect `timeout=None` in the call.

tests/tools/test_sql.py
```diff
@@ -591,6 +592,11 @@ async def test_execute_query(
         actual = await m.execute_query(query, max_rows=max_rows, max_chars=max_chars)
 
         assert actual == expected
+        keboola_client.storage_client.workspace_query.assert_called_once_with(
+            workspace_id=1234,
+            query=query,
+            timeout=httpx.Timeout(connect=5.0, read=300.0, write=10.0, pool=5.0),
+        )
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit f7d714ee13e76b95e2291a9f0c3a8a90e7ebfb68

Updated the assertion to reference `_BigQueryWorkspace._QUERY_TIMEOUT` instead of the hard-coded `300.0`, so the test stays correct if the constant changes.

tests/tools/components/test_tools.py
```diff
@@ -829,6 +829,65 @@ async def test_update_sql_transformation_folder(
...
+@pytest.mark.parametrize(
+    ('folder', 'patched_fn'),
+    [
+        ('Sales', 'set_configuration_folder_metadata'),
+        ('', 'clear_configuration_folder_metadata'),
+    ],
+    ids=['set_raises', 'clear_raises'],
+)
+@pytest.mark.asyncio
+async def test_update_sql_transformation_folder_metadata_error_is_swallowed(
```
tomasfejfar, line 841: moc nerůzumím

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -871,9 +871,25 @@ async def update_sql_transformation(
     else:
         folder_stripped = folder.strip()
         if folder_stripped:
-            await set_configuration_folder_metadata(client, sql_transformation_id, configuration_id, folder_stripped)
+            try:
```
tomasfejfar, line 874: Proč by to mělo být silent failure?

tomasfejfar (approved):
Nebránim se tomu, ale nerozumím proč když se nepodaří updatnout folder, tak to potichu projde

tests/tools/components/test_tools.py
Matovidlo: **Already addressed** ✅

The test patches `set_configuration_folder_metadata` (or `clear_configuration_folder_metadata` for the empty-folder case) to raise a `RuntimeError`, then calls `update_sql_transformation` and asserts `result.success is True`. It directly verifies that the try/except guard works — a metadata API failure doesn't bubble up and fail the whole tool call.

The two parametrize cases (`set_raises` / `clear_raises`) mirror the two branches in production code: setting a non-empty folder vs. clearing it (empty string passed as `folder`).

src/keboola_mcp_server/tools/components/tools.py
Matovidlo: **Intentional** ℹ️

The folder metadata call happens **after** `configuration_update` has already been committed to the Storage API. If we let the exception propagate through `@tool_errors()`, the user sees `"Server error"` — but their transformation WAS already updated; only the cosmetic UI folder grouping failed. That's a misleading outcome.

The folder is non-critical metadata (it just groups configs in the UI sidebar; the transformation is fully functional without it), so a transient 500 from the metadata endpoint shouldn't mask a successful update. The error IS logged as `WARNING` with `exc_info` (full stack trace in server logs), so it's traceable — not truly silent.

The same pattern has been in `create_sql_transformation` since AI-2344 specifically to prevent orphaning already-created resources when the metadata endpoint is flaky. `update_sql_transformation` was just missing it.

PR #530, opened 2026-05-18
AI-3216: Fix get_flows validation errors on unknown variants

Matovidlo:
## Description

**Linear**: [AI-3216](https://linear.app/keboola/issue/AI-3216/fix-mcp-flow-validation-errors-in-get-flows)

### Change Type

### Summary

`get_flows` is failing on stack `com-keboola-gcp-europe-west3` with **18 pydantic validation errors** for a single `ConditionalFlowTask`. The errors all originate from one real mismatch — the rest are cascading noise from a non-discriminated `Union`.

#### Root cause

For a flow task with shape:

```jsonc
{
  "type": "variable",
  "name": "...",
  "source": {
    "type": "function",
    "function": "YEAR",        // <-- not in our Literal['COUNT', 'DATE']
    "operands": [{ "type": "const", "value": "U" }]
  }
}
```

1. `FunctionCondition.function` is declared `Literal['COUNT', 'DATE']` in `src/keboola_mcp_server/tools/flow/model.py:208`. The actual backend produces `YEAR` (and likely other PHP `DateTime::format`-style tokens). The bundled `conditional-flow-schema.json` also enums only `COUNT`/`DATE`, so the schema/backend are out of sync.
2. `TaskConfiguration = Union[JobTaskConfiguration, NotificationTaskConfiguration, VariableTaskConfiguration]` (model.py:280) has **no discriminator**. When the inner `VariableTaskConfiguration` fails (because of #1), pydantic also tries `Job` and `Notification`, producing 12 noisy literal/missing-field errors that hide the real problem.
3. `Flow.from_api_response` (model.py:394-396) validates every task via `ConditionalFlowTask.model_validate(...)`. A single bad task in a single flow takes down the **entire `get_flows` response** (`unwrap_results` re-raises).

#### Proposed fix (incremental, low-risk)

| Step | Change | Why |
|------|--------|-----|
| A | Relax `FunctionCondition.function` from `Literal['COUNT', 'DATE']` to `str` (and update the JSON schema enum or drop it) | The comment at model.py:106-107 already says these models will be replaced by AI-service-fetched schema. Until then, the MCP shouldn't be stricter than the backend. |
| B | Make `TaskConfiguration` a **discriminated union** on `type` via `Annotated[Union[...], Field(discriminator='type')]`. Same for `ConditionObject` and `VariableSourceObject`. | Collapses cascading 18-error reports into one targeted error; clear "unknown task type" when a new variant lands. |
| C | In `Flow.from_api_response`, **catch `ValidationError` per task/phase** and log a warning, keeping the flow returnable with skipped/raw-passthrough tasks. | One malformed flow today nukes `get_flows` for the whole project. The MCP server's job is to display flows, not gatekeep them. |
| D | Add parametrized tests in `tests/tools/flow/test_tools.py` covering: `YEAR` function value, unknown task `type`, malformed-task-tolerant `get_flows`. | Locks in resilience; prevents future regression of the cascade. |

#### Out of scope

- Confirming the full canonical list of `function` values (`YEAR`, `MONTH`, `DAY`, `HOUR`, ...) — that's a follow-up for the AI-service schema endpoint mentioned in the model comment.
- Reworking `get_flow_schema` output (this PR only relaxes runtime parsing, not the schema we hand back to agents).

## Testing

### Optional testing

## Checklist

- [ ] Self-review completed
- [x] Project version bumped according to the change type (1.61.0 -> 1.61.1)

### DIFF
--- a/src/keboola_mcp_server/resources/conditional-flow-schema.json
+++ b/src/keboola_mcp_server/resources/conditional-flow-schema.json
@@ -392,8 +392,7 @@
         },
         "function": {
           "type": "string",
-          "enum": ["COUNT", "DATE"],
-          "description": "Function type: 'COUNT' (counts elements in array, requires 1 operand), 'DATE' (formats current date/time using PHP DateTime::format, requires 1 operand with format string)"
+          "description": "Function name. Supported values: 'COUNT' (returns the number of elements in its single array operand) and 'DATE' (returns the current date/time formatted via PHP DateTime::format; takes one operand that resolves to a format string such as 'Y', 'm', 'd', 'H', 'i', 's', or 'U'). The enum is intentionally omitted so unknown server-side values do not break flow listing; only the two values above are accepted at execution time."
         },
         "operands": {
           "type": "array",
--- a/src/keboola_mcp_server/tools/flow/model.py
+++ b/src/keboola_mcp_server/tools/flow/model.py
@@ -2,16 +2,19 @@
 Flow models for Keboola MCP server.
 """
 
+import logging
 from datetime import datetime
-from typing import Any, Literal, Optional, Union
+from typing import Annotated, Any, Literal, Optional, Union
 
-from pydantic import AliasChoices, BaseModel, Field
+from pydantic import AliasChoices, BaseModel, Field, ValidationError
 
 from keboola_mcp_server.clients.client import ORCHESTRATOR_COMPONENT_ID, FlowType, get_metadata_property
 from keboola_mcp_server.clients.storage import APIFlowResponse
 from keboola_mcp_server.config import MetadataField
 from keboola_mcp_server.links import Link
 from keboola_mcp_server.tools.flow.scheduler_model import SchedulesOutput
+
+LOG = logging.getLogger(__name__)
 
 # =============================================================================
 # RESPONSE MODELS
@@ -205,7 +208,15 @@ class FunctionCondition(BaseModel):
     """Function-based condition."""
 
     type: Literal['function'] = Field(description='Condition type')
-    function: Literal['COUNT', 'DATE'] = Field(description='Function type')
+    function: str = Field(
+        description=(
+            "Function name. Supported values: 'COUNT' (returns the number of elements in its single "
+            "array operand) and 'DATE' (returns the current date/time formatted via PHP "
+            "DateTime::format; takes one operand that resolves to a format string such as 'Y', 'm', "
+            "'d', 'H', 'i', 's', or 'U'). Typed as str rather than Literal so unknown server-side "
+            'values do not break flow listing; only the two values above are accepted at execution time.'
+        )
+    )
     operands: list['VariableSourceObject'] = Field(description='List of operand conditions')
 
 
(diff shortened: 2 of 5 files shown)


src/keboola_mcp_server/resources/conditional-flow-schema.json
```diff
@@ -392,8 +392,7 @@
         },
         "function": {
           "type": "string",
-          "enum": ["COUNT", "DATE"],
-          "description": "Function type: 'COUNT' (counts elements in array, requires 1 operand), 'DATE' (formats current date/time using PHP DateTime::format, requires 1 operand with format string)"
+          "description": "Function name. Known values include 'COUNT' (counts elements in array, requires 1 operand) and 'DATE' (formats current date/time using PHP DateTime::format, requires 1 operand with format string). The backend may evolve to accept additional tokens (e.g. 'YEAR'); the enum is intentionally omitted so unknown server-side values do not break flow parsing."
```
tomasfejfar: This is also used for creation though, no? So there should be list of known values in the description. I disagree that it's out of scope

src/keboola_mcp_server/tools/flow/model.py
```diff
@@ -205,7 +208,14 @@ class FunctionCondition(BaseModel):
     """Function-based condition."""
 
     type: Literal['function'] = Field(description='Condition type')
-    function: Literal['COUNT', 'DATE'] = Field(description='Function type')
+    function: str = Field(
+        description=(
+            "Function name. Known values include 'COUNT' (counts elements in array operand) and "
```
tomasfejfar: same here

tomasfejfar (approved):
LGTM, but I would fill in the full list of currently available functions in the description

src/keboola_mcp_server/resources/conditional-flow-schema.json
Matovidlo: **Fixed** ✅ in commit bab0e4c2d365d9c36a085cd66b767f12b9f12e7b

Fair point — you're right, this schema is also handed to agents for creation, so the description needs to be authoritative, not hand-wavy. I cross-checked the canonical source in `keboola/job-queue-daemon/src/Flow/Conditional/ConditionFactory.php::createFunction()` and the UI types in `keboola/ui/.../flows-v2/builder/types.ts` — the backend currently accepts **exactly two** function names: `COUNT` and `DATE`. Anything else (including the `YEAR` value that triggered this PR) is a malformed config — the backend's match arm falls through to `InvalidArgumentException('Unknown function: …')` at execution time.

Updated the `function` description in `conditional-flow-schema.json` (lines 393–395) to enumerate both functions, their operand counts, what they return, and the PHP `DateTime::format` tokens that `DATE` accepts (`Y`/`m`/`d`/`H`/`i`/`s`/`U`, …). The enum is still intentionally omitted at the JSON-schema level so unknown server-side values don't break `get_flows` listing, but the documented set is now the authoritative one. The model-side description in `model.py:208-222` was tightened in parallel.

src/keboola_mcp_server/tools/flow/model.py
Matovidlo: **Fixed** ✅ in commit bab0e4c2d365d9c36a085cd66b767f12b9f12e7b

Tightened the parallel description in `model.py` (lines 208–222) the same way: `FunctionCondition.function` now lists `COUNT` and `DATE` explicitly with their operand semantics and the PHP `DateTime::format` tokens used by `DATE`. Field remains `str` (not `Literal`) so malformed configs carrying unknown function names still parse and don't take down `get_flows`, but the description is now sourced from `keboola/job-queue-daemon/src/Flow/Conditional/ConditionFactory.php::createFunction()` — the only two valid names — instead of guessing.

tomasfejfar (approved):
LGTM

tomasfejfar:
Hopefully it won't cause hallucinations for the field :(

PR #547, opened 2026-05-29
AI-3286: redesign python-js data app draft flow

AdamVyborny:
## Description

**Linear**: [AI-3286](https://linear.app/keboola/issue/AI-3286)

### Change Type

### Summary

Replaces the v1.63 MVP "dev twin" surface with a coherent **draft** flow. Three agent-facing scenarios are now first-class and taught entirely through tool docstrings (no MCP prompt):

1. **Create-from-scratch** — prod + draft on branch `init` → preview in dev mode → merge → redeploy prod → delete draft.
2. **Edit-existing** — fresh credential + new draft on an agent-chosen descriptive branch → preview → merge → redeploy → delete.
3. **Continue an unfinished draft** — `get_data_apps([PROD])` now returns `drafts: [...]` inline, so the agent finds an abandoned iteration in one round-trip without scanning all configs.

Cross-cutting rule, stated in every relevant docstring: **MCP never runs git on your behalf**. All git work (clone, branch, commit, push, merge, branch-delete) is the agent's. MCP owns credentials, configs, deploys, and draft lifecycle.

**Surface changes**

| Tool | Change |
|---|---|
| `modify_python_js_data_app` | Default draft branch is `'init'` (was random `iter-<6-hex>`). Draft configs persist `parameters.dataApp.parentConfigurationId`. Docstring rewritten with the three scenarios. |
| `get_data_apps` | Detail responses for python-js **prod** apps now include `drafts: [...]` of every draft parented to that prod. One extra Storage list call per prod detail fetch; skipped for drafts and Streamlit. |
| **NEW** `delete_python_js_data_app_draft` | Deletes a draft's data-app instance (DSAPI) + Storage config. Refuses prod apps (no `isDraft`) and Streamlit. Returns the parent's `configuration_id` so the agent can pivot back. |
| `create_python_js_data_app_git_credential` | Docstring tightened: always mint against **prod** (drafts have no managed repo of their own). |
| `deploy_data_app` | Docstring reframes `mode='dev'` as "deploys the draft as a dev version of the data app" (hot reload + auto-auth iframe preview). No behaviour change. |

**Other**

- Drop `_HARDCODED_PYTHON_JS_IMAGE_VERSION = '1.6.1'`. Platform now applies a default; legacy pins on existing configs are preserved verbatim via deepcopy.
- `docs/python-js-data-apps.md` rewritten for v1.64.
- `pyproject.toml`: 1.63.3 → 1.64.0; `uv.lock` refreshed.
- `TOOLS.md` regenerated.

## Testing

### Optional testing

End-to-end verification of the three-scenario flow against `canary-orion` (per the verification section of the AI-3286 plan) is still TODO before un-drafting.

## Checklist

- [x] Self-review completed
- [x] Unit tests added/updated — extended parametrized tests for draft create (parentConfigurationId, default `'init'`), drafts list (N=0/1/3, draft-of-draft empty, Streamlit empty), and `delete_python_js_data_app_draft` (success, reject prod with two `isDraft` shapes, reject Streamlit)
- [x] Integration tests updated — `test_python_js_data_app_prod_and_draft_lifecycle` now asserts drafts surface in prod detail, exercises the new delete tool, and verifies the draft disappears from `drafts: [...]` afterwards
- [x] Project version bumped (minor: new tool + new schema field)
- [x] Documentation updated — full rewrite of `docs/python-js-data-apps.md`

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -2004,6 +2004,11 @@ as a dev version of the data app (hot reload + auto-auth for iframe preview); us
 branch-delete — is yours. MCP gives you authenticated clone URLs and manages configs/deploys;
 it never invokes git.
 
+**The draft flow below is mandatory — never edit prod source directly.** Every source-code
+change goes through a draft branch that the user previews and explicitly approves first. NEVER
+push directly to `main`: `main` only ever advances by merging an approved draft branch, and
+only after the user has approved that draft's preview.
+
 Three scenarios the agent has to distinguish:
 
 ## Scenario A — Create a brand-new data app
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -38,8 +38,10 @@ name or description.
 - [create_oauth_url](#create_oauth_url): Generates an OAuth authorization URL for a Keboola component configuration.
 
 ### Other Tools
-- [create_python_js_data_app_git_credential](#create_python_js_data_app_git_credential): Mints a one-time HTTPS token on a python-js data app so the caller can clone, pull, and push
-to the app's managed git repo over HTTPS.
+- [create_python_js_data_app_git_credential](#create_python_js_data_app_git_credential): Mints a one-time HTTPS token on a python-js **prod** data app so the caller can clone, pull,
+and push to the app's managed git repo over HTTPS.
+- [delete_python_js_data_app_draft](#delete_python_js_data_app_draft): Deletes a python-js DRAFT data app — both the data-app instance (DSAPI) and its Storage
+configuration.
 - [deploy_data_app](#deploy_data_app): Deploys/redeploys a data app or stops a running data app in the Keboola environment asynchronously, given the
 action and the configuration ID.
 - [get_data_apps](#get_data_apps): Lists summaries of data apps in the project given the limit and offset or gets details of a data apps by
@@ -1712,8 +1714,15 @@ Example 4 - Update storage mappings:
 
 **Description**:
 
-Mints a one-time HTTPS token on a python-js data app so the caller can clone, pull, and push
-to the app's managed git repo over HTTPS.
+Mints a one-time HTTPS token on a python-js **prod** data app so the caller can clone, pull,
+and push to the app's managed git repo over HTTPS.
+
+**Always call against the prod app's configuration_id** — drafts have no managed repo of their
+own, so calling this on a draft fails. The prod app is the canonical repo owner; drafts
+iterate against branches of that same repo.
+
+**MCP never runs git on your behalf.** All git work — clone, branch, commit, push, merge,
+branch-delete — is yours. This tool only mints credentials.
 
 Returns a ready-to-use `git_clone_url` of the form `https://kai:<secret>@<host>/<path>.git`
 plus the raw `secret`. The token is returned **only** at creation — the platform cannot return
@@ -1725,19 +1734,21 @@ additional token without invalidating any tokens already held by other clients.
 
 ## When to call
 
-1. **Right after `modify_python_js_data_app` create** — the new app has a managed repo but no
-   credentials yet. Call this tool with the new app's `configuration_id` to enable git access.
+1. **Right after `modify_python_js_data_app` create of a prod app** — the new prod has a
+   managed repo but no credentials yet. Call this tool with the new app's `configuration_id`
+   to enable git access. (Note: when creating a **draft**, the prod-side token is minted and
+   embedded into the returned `git_clone_url` automatically — no separate call needed.)
 
-2. **Recovery when the cached token is gone** (e.g., a fresh Kai sandbox continuing an old draft
-   — the previous sandbox's filesystem was wiped, taking the cached `git_clone_url` with it).
-   If `git clone`/`pull`/`push` against a python-js app the LLM did not create in the current
-   session fails with `Authentication failed` (HTTP 401), call this tool with the same
-   `configuration_id` to mint a fresh token. Existing credentials remain valid, so other clients
-   are not disrupted.
+2. **Recovery when the cached token is gone / continuing an unfinished draft** — e.g., a fresh
+   sandbox continuing yesterday's work, with the previous sandbox's filesystem wiped. The
+   cached `git_clone_url` is lost; the configuration ID for the prod app is all you have.
+   Call this tool with the **prod app's** `configuration_id` to mint a fresh token (drafts
+   have no managed repo, so always mint against prod). Existing credentials remain valid, so
+   other clients are not disrupted.
 
 ## Constraints
-- Only python-js data apps have a managed git repo. Streamlit apps reject the call with a clear
-  error.
+- Only python-js prod data apps have a managed git repo. Streamlit apps reject the call with
+  a clear error.
 - Permissions are always `readWrite` — the LLM virtually always needs push access. The
   data-science API supports read-only credentials, but the tool does not expose that knob;
   revisit once a real use case appears.
@@ -1761,6 +1772,56 @@ additional token without invalidating any tokens already held by other clients.
 ```
 
 ---
+<a name="delete_python_js_data_app_draft"></a>
+## delete_python_js_data_app_draft
+**Annotations**: `destructive`
+
+**Tags**: `data-apps`
+
+**Description**:
+
+Deletes a python-js DRAFT data app — both the data-app instance (DSAPI) and its Storage
+configuration.
+
+**MCP never runs git on your behalf.** Deleting the feature branch on the remote is your job;
+this tool only tears down the draft config and its data-app instance.
+
+WHEN TO CALL: at the end of a promote-to-prod sequence, after you have merged the draft's
+branch into `main`, pushed, deleted the feature branch from the remote, and redeployed the
+prod app. The Keboola UI lists drafts under their parent prod app; once you call this tool,
+the draft disappears from that list.
+
+WHAT THIS TOOL REFUSES:
+  - prod apps (no `isDraft` flag) — protects against accidental prod deletion;
+  - Streamlit apps — they have no draft concept.
+
+WHAT THIS TOOL DOES NOT DO:
+  - Run git. Deleting the feature branch on the remote is your job.
+  - Revoke the prod-side git credential minted when the draft was created. Credential
+    rotation is the user's job via the Keboola UI.
+
+After a successful call, pivot back to the parent prod app (its configuration_id is returned
+in the response) or to `get_data_apps` for further work.
+
+
+**Input JSON Schema**:
+```json
+{
+  "additionalProperties": false,
+  "properties": {
+    "configuration_id": {
+      "description": "Storage configuration ID of the python-js draft data app to delete.",
+      "type": "string"
+    }
+  },
+  "required": [
+    "configuration_id"
+  ],
+  "type": "object"
+}
+```
+
+---
 <a name="deploy_data_app"></a>
 ## deploy_data_app
 **Annotations**: 
@@ -1772,13 +1833,21 @@ additional token without invalidating any tokens already held by other clients.
 Deploys/redeploys a data app or stops a running data app in the Keboola environment asynchronously, given the
 action and the configuration ID.
 
+**MCP never runs git on your behalf.** All git work — clone, branch, commit, push, merge,
+branch-delete — is yours. This tool only triggers deploys against existing git state.
+
 ## Mode and branch (python-js apps)
-- `mode='dev'` enables the in-platform preview for a python-js dev twin. Pair with `branch` to deploy a
-  specific git branch (edit flow); without `branch`, the dev twin deploys `main`.
-- For prod redeploys (including after merging a feature branch into `main` during the edit flow), use
-  no `mode` and no `branch` — the prod app picks up the current `main`.
-- python-js apps do NOT fetch a Storage `configVersion` for deployment (their source lives in git, not in
-  the Storage configuration); this is handled automatically.
+- `mode='dev'` deploys the target as a **dev version of the data app** — the runtime uses a
+  development `setup.sh` (hot reload) and the data-app proxy enables an auto-auth path so an
+  iframe preview can render without a manual login. Only meaningful on **draft** configs
+  (python-js apps with `isDraft=true`).
+- For prod redeploys (including after merging a draft's branch into `main`), use no `mode` and
+  no `branch` — the prod app picks up the current `main`.
+- The optional `branch=` argument overrides the branch the draft deploys from for this single
+  deploy. Normally unnecessary — drafts have their draft branch pinned in
+  `parameters.dataApp.git.branch` at create time.
+- python-js apps do NOT fetch a Storage `configVersion` for deployment (their source lives in
+  git, not in the Storage configuration); this is handled automatically.
 
 ## Streamlit apps
 Streamlit apps have no managed git repo, so `mode` and `branch` have no effect on the
@@ -1826,7 +1895,7 @@ error for any app type (Streamlit or python-js).
         }
       ],
       "default": null,
-      "description": "Deployment mode. Set to \"dev\" to enable the in-platform preview for python-js data apps. Leave None (default) for Streamlit apps and for production deploys."
+      "description": "Deployment mode. Set to \"dev\" to deploy a python-js draft as a **dev version of the data app** \u2014 the runtime uses a development `setup.sh` (hot reload), and the data-app proxy enables an auto-auth path so an iframe preview can render without a manual login. Only meaningful on **draft** configs (python-js apps with `isDraft=true`). Leave None (default) for prod redeploys and for Streamlit apps."
     },
     "branch": {
       "anyOf": [
@@ -1838,7 +1907,7 @@ error for any app type (Streamlit or python-js).
         }
       ],
       "default": null,
-      "description": "Git branch to deploy from. Only meaningful when `mode=\"dev\"` for python-js apps backed by a managed repo \u2014 the dev twin will deploy from this branch instead of `main`, enabling branch-based preview during the edit flow. Leave None for prod deploys and for Streamlit apps."
+      "description": "Git branch to deploy from. Only meaningful when `mode=\"dev\"` for python-js drafts. Normally unnecessary \u2014 drafts have their branch pinned in `parameters.dataApp.git.branch` at create time; this argument overrides that pin for this single deploy (escape hatch). Leave None for prod deploys and for Streamlit apps."
     }
   },
   "required": [
@@ -1875,6 +1944,12 @@ data app logs to investigate in-app errors. The logs may be updated after openin
   (when `configuration_ids` is provided). The inventory list always returns `repo_url=None`,
   even for python-js apps with a managed repo — to retrieve the URL, call this tool again
   with the target `configuration_ids`.
+- When called with `configuration_ids=[<prod-cfg>]` for a python-js **prod** app, the response
+  includes a `drafts: [...]` array of every draft (configs with `isDraft=true` and
+  `parentConfigurationId == <prod-cfg>`) currently in the project. Drafts in trash are not
+  included. Use this to discover existing drafts when continuing a previously abandoned
+  iteration (Scenario C in `modify_python_js_data_app`). The array is empty for drafts
+  themselves and for Streamlit apps.
 
 
 **Input JSON Schema**:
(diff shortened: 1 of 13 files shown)


src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -1415,12 +1563,55 @@ async def _fetch_data_app_details_task(
...
+        return data_app
     except Exception:
         LOG.exception(f'Failed to fetch data app by configuration ID: {configuration_id}')
         return configuration_id
 
 
+def _is_draft_config(configuration: Mapping[str, Any]) -> bool:
+    """True iff the data app's stored configuration carries `parameters.dataApp.isDraft = true`."""
+    return cast(Mapping[str, Any], configuration.get('parameters') or {}).get('dataApp', {}).get('isDraft') is True
+
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in cc995c9b

Made _is_draft_config shape-safe: parameters/dataApp that aren't mappings now return False instead of raising AttributeError. Added a parametrized unit test (test_is_draft_config) covering draft, non-draft, and malformed shapes.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -1415,12 +1563,55 @@ async def _fetch_data_app_details_task(
...
+    """List the drafts (configs with `parentConfigurationId == prod_configuration_id`) of a python-js
+    prod app. Returns full `DataAppSummary` entries (one extra DSAPI fetch per draft, capped at
+    10 parallel). Drafts in trash are not returned by `configuration_list` and so do not appear here.
+    """
+    configs = await client.storage_client.configuration_list(DATA_APP_COMPONENT_ID)
+    draft_cfg_ids: list[str] = []
+    for cfg in configs:
+        cfg_body = cast(Mapping[str, Any], cfg.get('configuration') or {})
+        data_app_block = cast(Mapping[str, Any], cfg_body.get('parameters') or {}).get('dataApp') or {}
+        if data_app_block.get('parentConfigurationId') == prod_configuration_id:
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in b9ad861e

_fetch_prod_drafts now requires _is_draft_config(cfg) AND parentConfigurationId == prod, so a config that kept the parent pointer but lost isDraft is no longer surfaced as a draft. Extended test_get_data_apps_detail_for_prod_lists_drafts with a non-draft child that must be excluded.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -968,19 +1027,21 @@ async def create_python_js_data_app_git_credential(
...
+       cached `git_clone_url` is lost; the configuration ID for the prod app is all you have.
+       Call this tool with the **prod app's** `configuration_id` to mint a fresh token (drafts
+       have no managed repo, so always mint against prod). Existing credentials remain valid, so
+       other clients are not disrupted.
 
     ## Constraints
-    - Only python-js data apps have a managed git repo. Streamlit apps reject the call with a clear
-      error.
+    - Only python-js prod data apps have a managed git repo. Streamlit apps reject the call with
+      a clear error.
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in acfc692b

Added an early _is_draft_config guard: drafts now get a clear 'this is a draft, mint against the parent prod app (parentConfigurationId=...)' error instead of falling through to the misleading https_url=None 'platform-side bug' path. Unit test for this is coming in a follow-up commit (separate review thread).

tests/tools/test_data_apps.py
```diff
@@ -2096,3 +2086,294 @@ async def test_create_python_js_data_app_git_credential_invalid_configuration_id
         )
 
     keboola_client.data_science_client.create_app_git_credential.assert_not_called()
+
+
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in 42679751

Added test_create_python_js_data_app_git_credential_rejects_draft: asserts a draft config raises the prod-only error (with the parentConfigurationId hint) and that neither get_app_git_repo nor create_app_git_credential is called. Pairs with the guard added in acfc692b.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -1270,6 +1346,78 @@ async def deploy_data_app(
...
+    data_app_block = cast(Mapping[str, Any], data_app.configuration.get('parameters') or {}).get('dataApp') or {}
+    parent_cfg_id = data_app_block.get('parentConfigurationId')
+    parent_configuration_id: Optional[str] = parent_cfg_id if isinstance(parent_cfg_id, str) else None
+
+    await client.data_science_client.delete_data_app(data_app.data_app_id)
+    await client.storage_client.configuration_delete(
+        component_id=DATA_APP_COMPONENT_ID,
+        configuration_id=configuration_id,
+        skip_trash=False,
+    )
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in 4bfdf109

Made the Storage delete idempotent rather than removing it. Investigation showed the call is load-bearing, not redundant: the integ test passes WITH the double-delete, which is only possible if the config still exists when configuration_delete runs (a single DELETE raises on 404 via _raise_for_status) — i.e. DSAPI's storage-config removal is not synchronous. So we keep the delete but swallow a 404 (already gone), while any other status still propagates. Added a parametrized success case for the 404 path and a test asserting non-404 errors propagate.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -834,27 +884,27 @@ async def modify_python_js_data_app(
...
-            # serialize an external-git block into the dev twin's config.
+            # Draft create path: resolve the parent's repo + mint a parent-side credential, then
+            # serialize an external-git block into the draft's config.
             parent = await _fetch_data_app(client, configuration_id=parent_configuration_id, data_app_id=None)
             if parent.type != 'python-js':
                 raise ValueError(
                     f'parent_configuration_id "{parent_configuration_id}" is type "{parent.type}", but only '
-                    f'python-js prod apps can parent a dev twin.'
+                    f'python-js prod apps can parent a draft.'
                 )
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in 68bfebdb

Added an explicit _is_draft_config guard on the parent in the draft-create path: a draft parent now raises 'is itself a python-js **draft**, not a prod app — pass the prod app's configuration_id' before reaching the repo_url check, so no misleading 'platform-side bug' error and no credential minted. Added test_modify_python_js_data_app_create_draft_rejects_when_parent_is_draft.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -1270,6 +1357,89 @@ async def deploy_data_app(
...
+            f'Storage config "{configuration_id}" was already deleted (404) while tearing down the '
+            'draft — treating as already removed.'
+        )
+
+    links = links_manager.get_data_app_links(
+        configuration_id=parent_configuration_id or configuration_id,
+        configuration_name=data_app.name,
+        deployment_link=None,
+        uses_basic_authentication=False,
+    )
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in 4a1e837c

The pivot link now uses an explicit 'parent prod app' label when parent_configuration_id is present (we don't have the parent's name here without an extra fetch, which isn't worth it for a link label). The link still points at the parent config but no longer carries the draft's name. Added assertions in the success test that the link targets the parent config and is labelled accordingly.

docs/python-js-data-apps.md
```diff
@@ -1,73 +1,74 @@
...
+| `deploy_data_app` | No behaviour change. Docstring reframes `mode='dev'` as "deploys the draft as a **dev version of the data app**" (hot reload + auto-auth for iframe preview). | Deploy/redeploy or stop. `mode='dev'` only meaningful on drafts. |
+| `create_python_js_data_app_git_credential` | No behaviour change. Docstring tightens the prod-only contract: drafts have no managed repo, always mint against prod. | Mint a one-time HTTPS token on a python-js prod app's managed repo. |
+| `get_data_apps` | Detail responses for python-js **prod** apps now include a `drafts: [...]` array of `DataAppSummary` entries — every draft (`isDraft=true`, `parentConfigurationId == <prod-cfg>`) parented to that prod, fetched with one extra `configuration_list` round-trip. Empty for drafts themselves and for Streamlit apps. | List or detail-fetch data apps; discover drafts. |
+| **NEW** `delete_python_js_data_app_draft` | New tool. Deletes a draft's data-app instance (DSAPI) and its Storage configuration. Refuses prod apps (no `isDraft` flag) and Streamlit apps. | Cleanup primitive — call after a draft's branch has been promoted to `main`. |
```
Matovidlo, line 64: 🔴 1. New delete_python_js_data_app_draft tool is not branch-gated like its siblings — src/keboola_mcp_server/mcp.py:377 and :449

  This is the one real bug. The redesign adds the delete_python_js_data_app_draft tool (registered at tools/data_apps.py:78), but it was not added to either branch guard:

  - mcp.py:373-384 (on_list_tools) filters out data-app tools on non-main branches — lists modify_streamlit_data_app, modify_python_js_data_app, create_python_js_data_app_git_credential, get_data_apps,
  deploy_data_app, but not delete_python_js_data_app_draft.
  - mcp.py:449-457 (on_call_tool) raises 'Data apps are supported only in the main production branch.' for that same set — again missing the delete tool.

  Failure scenario: a client on a dev branch (branch_id set) sees and can invoke delete_python_js_data_app_draft, which then runs delete_data_app + configuration_delete outside the main-branch invariant that every
  other data-app tool enforces. The whole AI-3286 premise is "main is merge-only / draft flow is mandatory" (commit 90703df8), so leaking a destructive draft tool onto non-main branches contradicts the design.
  Both lists should include the new tool name.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -1415,12 +1601,70 @@ async def _fetch_data_app_details_task(
             uses_basic_authentication=_uses_basic_authentication(data_app.configuration.get('authorization') or {}),
         )
         logs = await _fetch_logs(client, data_app.data_app_id)
-        return data_app.with_links(links).with_deployment_info(logs)
+        data_app = data_app.with_links(links).with_deployment_info(logs)
+        # Drafts of a python-js prod are surfaced inline so the agent can find them in one round-trip
+        # — see Scenario C in `modify_python_js_data_app`. Skip for drafts themselves and for Streamlit
+        # (neither has children).
+        if data_app.type == 'python-js' and not _is_draft_config(data_app.configuration):
+            data_app.drafts = await _fetch_prod_drafts(client, prod_configuration_id=data_app.configuration_id)
```
Matovidlo:
```
  except Exception:
      LOG.exception(...); return None
  ...
  return [s for s in results if isinstance(s, DataAppSummary)]
```
  A transient DSAPI failure (expired token, timeout, 5xx) on one draft's detail fetch removes it from the returned list with no signal to the caller. The user listing a prod app's drafts sees fewer than exist and
  can't distinguish "temporarily unreachable" from "deleted" — risky right before a teardown decision. 

It's logged but not surfaced. Consider returning a partial-failure marker or at least counting omissions in
  the response.

Matovidlo (commented):
Cool, did you test it on local MCP, and both backends ? Just to be sure there are not regressions

docs/python-js-data-apps.md
AdamVyborny: Fixed in 284dd9d5

Good catch — this was a real leak. Added `delete_python_js_data_app_draft` to the branch gate and, to prevent the same drift recurring, extracted the duplicated tool-name set into a single `DATA_APP_BRANCH_GATED_TOOLS` constant (mirroring `SEMANTIC_TOOL_NAMES`) now used by both `on_list_tools` and `on_call_tool`. Extended the parametrized branch-filter tests to cover the delete tool.

src/keboola_mcp_server/tools/data_apps.py
AdamVyborny: Fixed in 4887313a

Agreed — silently shrinking the list right before a teardown decision is the dangerous part. `_fetch_prod_drafts` now returns `(drafts, unavailable_count)` and the count is surfaced via a new `DataApp.drafts_unavailable` field. Its description warns that a non-zero value means `drafts` is incomplete and those drafts were NOT deleted (retry to get the full list), so the agent can distinguish "temporarily unreachable" from "deleted". Failures are still logged as before. Added a test covering one transient failure (surviving draft returned, `drafts_unavailable == 1`).

Matovidlo (approved):
LGTM

docs/python-js-data-apps.md
```diff
@@ -1,73 +1,74 @@
...
+Python-JS data apps are backed by a **git repository**: source code lives in the repo, not in the Storage configuration. The MCP server exposes a small set of primitives — `modify_python_js_data_app`, `deploy_data_app`, `create_python_js_data_app_git_credential`, `get_data_apps`, `delete_python_js_data_app_draft` — that together support a **two-app project model**:
 
 - A persistent **prod app** that users actually run. The prod app **owns the only managed git repo** in the project.
-- One or more **dev twins** that are *external-git* apps configured to clone the prod app's repo on every deploy. Dev twins serve as the LLM's iteration sandbox.
+- Zero or more **drafts** parented to that prod app. A draft is a Storage configuration with `parameters.dataApp.isDraft=true` and `parameters.dataApp.parentConfigurationId=<prod cfg id>`; it's an *external-git* app that clones the parent prod's repo at a pinned branch on every deploy. Drafts serve as the LLM's iteration sandbox.
```
tomasfejfar, line 14: unless there is any other app that has "parent" the "isDraft" is useless and may potentially drift (like when using vibecoded cli from padak)

docs/python-js-data-apps.md
```diff
@@ -1,73 +1,74 @@
...
-| `deploy_data_app` | No change | Same as before: `mode='dev'` plus optional `branch` to override the branch the dev twin deploys from. |
-| `create_python_js_data_app_git_credential` | No change | Mint a one-time HTTPS token on a prod app's managed repo. Used for the lost-token recovery flow; **not** needed for the standard dev-twin create flow (the token is minted and embedded into the returned `git_clone_url` automatically). |
+| `modify_python_js_data_app` | Default draft branch is now `'init'` (was `iter-<6-hex>`). Drafts persist `parameters.dataApp.parentConfigurationId` so they can be discovered cheaply. Docstring rewritten to drop "dev twin". | Create/update a prod app; create a draft bound to the parent prod app's managed repo. |
+| `deploy_data_app` | No behaviour change. Docstring reframes `mode='dev'` as "deploys the draft as a **dev version of the data app**" (hot reload + auto-auth for iframe preview). | Deploy/redeploy or stop. `mode='dev'` only meaningful on drafts. |
```
tomasfejfar, line 61: auto-auth for iframe works how exactly? Isn't it just security through obscurity?

docs/python-js-data-apps.md
tomasfejfar, line 64: Probably the first delete endppoint in MCP

docs/python-js-data-apps.md
```diff
@@ -81,15 +82,21 @@ Newly created python-js apps carry `runtime.workspace.enabled = true` in their S
...
 
 The platform supports multiple credentials per app, so minting a fresh credential never invalidates earlier ones — important for the recovery flow below and for parallel iteration.
 
+### Credential lifecycle
+
+- A new credential is minted on the parent **prod** app every time a draft is created. The prod-side credential is encrypted into the draft's `parameters.dataApp.git.#password`.
+- **Deleting a draft via `delete_python_js_data_app_draft` does NOT revoke that credential.** The MCP surface has no list/delete-credential affordance; rotation is the user's job via the Keboola UI. (DSAPI does support `DELETE /apps/{id}/git-repo/credentials/{credentialId}`, but the MCP server does not store the credential ID on the draft to make targeted revocation possible.)
```
tomasfejfar, line 91: Isn'T this a problem? It's a choice, but the credentials should live with the draft, no?

docs/python-js-data-apps.md
```diff
@@ -81,15 +82,21 @@ Newly created python-js apps carry `runtime.workspace.enabled = true` in their S
...
+- Because the platform accepts unlimited per-app credentials and ignores stale ones for auth purposes, accumulated drafts produce stale-but-harmless credentials on the prod app over time. Treat this as a known trade-off until the platform-side drafts feature lands ([AI-3240](https://linear.app/keboola/issue/AI-3240)).
```
tomasfejfar: It's already closed 🤷

TOOLS.md
```diff
@@ -1920,66 +1995,96 @@ data app logs to investigate in-app errors. The logs may be updated after openin
...
-
-## Update flow (modifying an existing app's deployment metadata)
-When `configuration_id` is set: updates the Storage configuration (auto-suspend, name, description,
-`authentication_type`). `slug`, `parent_configuration_id`, and `branch` are all rejected here —
-slug is immutable, the repo binding is fixed at creation, and the iteration branch only makes
-sense at create time. Use `authentication_type='default'` to keep the existing auth setup
-(including OIDC configured outside the MCP); pass `'no-auth'` or `'basic-auth'` to overwrite.
-After updating, ALWAYS call `deploy_data_app(action='deploy', ...)` to restart the app so the
-changes take effect.
+Two-app project model. Every python-js project has a persistent **prod app** that owns the
```
tomasfejfar, line 1994: This should be a skill - muddies any agent session with a workflow for data apps

tomasfejfar (approved):
I just did a high level review, seeing that Martin already reviewed.

docs/python-js-data-apps.md
AdamVyborny: IDK, done by Pepa. There is some handshake request on proxy with SAPI token, so it is not just obscurity

docs/python-js-data-apps.md
AdamVyborny: Credentials are always retrieved from prod version, as the endpoint is creating new repo for each app. I should probably make the credential tool to always retrieve credentials from parent config, so we don't rely on LLM to call it on parent ID

docs/python-js-data-apps.md
Matovidlo: Check proxy in keboola-as-code if interested :wink:

docs/python-js-data-apps.md
Matovidlo: We need to update the tooling to generally be accessible for deletion but that's future scope for tooling

docs/python-js-data-apps.md
AdamVyborny: Right that the two halves are partly redundant. We can't drop `isDraft` right now because the **Keboola UI now relies on it** to hide drafts from the prod app list. The MCP side already requires *both* `isDraft=true` AND `parentConfigurationId` to treat a config as a draft, so a drifted config (one set without the other) is treated as "not a draft" rather than surfacing inconsistently. Filed a follow-up to decide whether to derive draft-ness from the parent pointer alone (and migrate the UI off the flag) or add a drift guard: AI-3311 — https://linear.app/keboola/issue/AI-3311

docs/python-js-data-apps.md
AdamVyborny: Credentials are always retrieved from the **prod** version — the prod app owns the only managed repo, and each draft is an external-git app that clones the prod repo at its pinned branch. For now the credential tool **rejects a draft configuration_id with an actionable error** that points the caller at the parent prod app (it includes the `parentConfigurationId` in the message), so a wrong call fails loudly rather than minting against the wrong place. I think that's enough for now; if we see the tool erroring on drafts frequently in practice, we can refactor it to auto-resolve the parent from `parentConfigurationId` so we don't rely on the LLM to pass the prod ID.

TOOLS.md
AdamVyborny: Agreed — teaching the whole multi-step draft→prod choreography through tool docstrings means it's loaded into every agent session, even ones unrelated to data apps. Extracting it into an on-demand skill (and keeping the docstrings to the per-tool contract) is the right direction. Filed a follow-up: AI-3312 — https://linear.app/keboola/issue/AI-3312

docs/python-js-data-apps.md
AdamVyborny: Fixed in 5af11b5e

You're right — AI-3240 is closed (Done). It was a *design-tracking* ticket, not the platform implementation, and the native-drafts feature hasn't actually shipped (v1.64.0 is the MCP-side workaround built precisely because it isn't there yet), so citing it as "until AI-3240 lands" was misleading. Removed the AI-3240 references from the doc (it was cited in 4 places, incl. the header tracker line and an end-of-doc 'long-term fix' pointer) and reworded to describe the workaround/trade-off as the current state without tying it to a closed ticket.

docs/python-js-data-apps.md
tomasfejfar: The whole thing feels off (when seeing with fresh eyes). The fact that drafts and credentials to the repo are disconnected feels like a red flag. The acccumulated credentials without any path to kill them as well. Optimally the delete dev version would drop the credentials as well, no? (inside MCP)

AdamVyborny: Ideally we should want credentials for the draft (without creating new repo as it works now) and then we can delete them with draft. I open it on next stand-up with Pepa.

PR #562, opened 2026-06-04
feat: attach k8s SA step-up header to workspace provisioning for read-only users [PSGO-256]

Matovidlo:
## Description

**Linear**: https://linear.app/keboola/issue/PSGO-256/add-support-for-manage-sdk-in-go-monorepo

### Change Type

### Summary

When `KBC_KUBERNETES_TOKEN_PATH` points at the projected Kubernetes ServiceAccount token (mounted into the deployed MCP server), workspace provisioning keeps the **user's own Storage token** and additionally carries the SA JWT as the `X-Kubernetes-Authorization` **step-up header** on the billing-configuration create and the workspace create. Connection waives the permissions a read-only token lacks when the ServiceAccount is authorized for workspace provisioning. **No privileged token is ever minted** — every action runs under, and is audited to, the user's token.

The shared MCP workspace is **no longer tracked by a branch-metadata pointer** (which would have needed an elevated metadata write a read-only token can't do). It is a regular workspace created under a configuration of the `keboola.mcp-server-tool` component, and is rediscovered by listing that component's configurations and fetching each config's workspaces through the config-scoped workspaces endpoint (`WorkspaceManager._find_ws_in_branch`).

- `WorkspaceManager` takes `kubernetes_token_path`; the step-up provisioning client is built once per session (token file read then — kubelet rotation safe) and cached for the manager's lifetime
- the step-up client **propagates the read-only guard** of the user's Storage client — the step-up widens server-side permissions only, never the client-side write guard
- `KBC_KUBERNETES_TOKEN_PATH` is read from the process environment only — deliberately **not** a `Config` field so it can never be injected via HTTP headers or per-request config
- missing/empty token file fails loudly — the step-up header is never silently dropped
- docs: `docs/kubernetes-sa-auth.md` incl. **Drawbacks / Limitations**: locally-run MCP servers have no projected SA token, so a local user must connect at least once to the deployed MCP server so the workspace is created there; afterwards local `query_data` works via query-service

> Note: requires the Connection-side step-up support; until that support covers all provisioning endpoints, the workspace-credentials endpoint is the first one honoring the header (query-service path).

## Testing

### Optional testing

Unit tests: step-up header construction (user token + `Bearer <jwt>`), read-only guard propagation, token-file read + per-session caching, fail-loud on missing/empty file, fallback to the user client when the env is unset. E2E smoke planned after the paired backend changes land.

## Checklist

- [x] Self-review completed

### DIFF
--- a/docs/kubernetes-sa-auth.md
+++ b/docs/kubernetes-sa-auth.md
@@ -12,8 +12,11 @@ requests keep the **user's own Storage token** and additionally carry the SA JWT
 the `X-Kubernetes-Authorization` header:
 
 - creating the billing configuration (`keboola.mcp-server-tool`),
-- creating the workspace itself,
-- writing the workspace ID into the branch metadata.
+- creating the workspace itself.
+
+The workspace is a regular workspace created under the `keboola.mcp-server-tool`
+configuration and is rediscovered by listing workspaces and matching that component —
+no branch-metadata pointer is written anymore.
 
 Connection validates the JWT and, when the ServiceAccount is authorized for workspace
 provisioning, waives the permissions the user's token lacks —
@@ -48,6 +51,6 @@ Key properties:
 - Read-only users who only ever use a locally-run MCP server and never connect to
   the deployed one cannot get a workspace created — their own token lacks the
   required permissions, and there is no ServiceAccount token to step up with.
-- For read-only users the step-up is intentionally limited to workspace provisioning.
-  Any other write operation still requires adequate permissions on the user's own
-  token and keeps failing for a read-only user.
+- For read-only users the step-up is intentionally limited to workspace provisioning
+  and storage event emission. Any other write operation still requires adequate
+  permissions on the user's own token and keeps failing for a read-only user.
--- a/docs/kubernetes-sa-auth.md
+++ b/docs/kubernetes-sa-auth.md
@@ -1,0 +1,53 @@
+# Kubernetes ServiceAccount Authentication (Deployed MCP Server)
+
+The MCP server deployed in Keboola infrastructure can attach its projected Kubernetes
+ServiceAccount JWT as a **step-up credential** to Connection Storage API requests
+(`X-Kubernetes-Authorization: Bearer <jwt>`).
+
+## How it works
+
+When the `KBC_KUBERNETES_TOKEN_PATH` environment variable points at the projected
+ServiceAccount token (mounted into the deployed MCP server), workspace provisioning
+requests keep the **user's own Storage token** and additionally carry the SA JWT as
+the `X-Kubernetes-Authorization` header:
+
+- creating the billing configuration (`keboola.mcp-server-tool`),
+- creating the workspace itself,
+- writing the workspace ID into the branch metadata.
+
+Connection validates the JWT and, when the ServiceAccount is authorized for workspace
+provisioning, waives the permissions the user's token lacks —
+workspace provisioning works even for read-only project members. **No privileged token
+is ever minted**; every action runs under, and is audited to, the user's own token.
+Everything else (queries, reads, tools) is untouched.
+
+Key properties:
+
+- The token file is read per provisioning flow — kubelet rotation is picked up
+  automatically (no caching across flows).
+- `KBC_KUBERNETES_TOKEN_PATH` is read from the process environment only; it cannot be
+  set or overridden via HTTP headers or per-request config.
+- A missing or empty token file fails loudly — the step-up header is never silently
+  dropped.
+- When the env variable is not set (any locally-run server), the behavior is exactly
+  as before — provisioning uses the user's token alone.
+- Requires the Connection-side step-up support; until that support covers all
+  provisioning endpoints, the workspace-credentials endpoint is the first one
+  honoring the header.
+
+## Drawbacks / Limitations
+
+- **Locally-run MCP servers have no projected ServiceAccount token**, so the
+  step-up provisioning path only works on the MCP server deployed in Keboola
+  infrastructure. A user running the MCP server locally (stdio/custom deployment)
+  must connect **at least once to the deployed (remote) MCP server**, so that their
+  workspace is created there (with proper billing configuration and credentials).
+  After that, the local MCP server works for querying: its `query_data` tool sends
+  queries to the in-infrastructure query-service, which attaches its own step-up
+  header when obtaining workspace credentials.
+- Read-only users who only ever use a locally-run MCP server and never connect to
+  the deployed one cannot get a workspace created — their own token lacks the
+  required permissions, and there is no ServiceAccount token to step up with.
+- For read-only users the step-up is intentionally limited to workspace provisioning.
+  Any other write operation still requires adequate permissions on the user's own
+  token and keeps failing for a read-only user.
--- a/docs/kubernetes-sa-auth.md
+++ b/docs/kubernetes-sa-auth.md
@@ -14,9 +14,10 @@ the `X-Kubernetes-Authorization` header:
 - creating the billing configuration (`keboola.mcp-server-tool`),
 - creating the workspace itself.
 
-The workspace is a regular workspace created under the `keboola.mcp-server-tool`
-configuration and is rediscovered by listing workspaces and matching that component —
-no branch-metadata pointer is written anymore.
+The workspace is a regular workspace created under a configuration of the
+`keboola.mcp-server-tool` component, and is rediscovered by listing that component's
+configurations and fetching each config's workspaces through the config-scoped
+workspaces endpoint — no branch-metadata pointer is written anymore.
 
 Connection validates the JWT and, when the ServiceAccount is authorized for workspace
 provisioning, waives the permissions the user's token lacks —
@@ -26,8 +27,10 @@ Everything else (queries, reads, tools) is untouched.
 
 Key properties:
 
-- The token file is read per provisioning flow — kubelet rotation is picked up
-  automatically (no caching across flows).
+- The token file is read when the step-up provisioning client is first built, once per
+  session (`WorkspaceManager` lifetime). Provisioning happens at most once per session,
+  well within the projected token's rotation window; a new session re-reads the file, so
+  kubelet rotation is picked up without restarting the server.
 - `KBC_KUBERNETES_TOKEN_PATH` is read from the process environment only; it cannot be
   set or overridden via HTTP headers or per-request config.
 - A missing or empty token file fails loudly — the step-up header is never silently
--- a/integtests/test_workspace.py
+++ b/integtests/test_workspace.py
@@ -20,17 +20,11 @@ async def dynamic_manager(
     storage_client = sync_storage_client
     token_info = storage_client.tokens.verify()
     project_id: str = token_info['owner']['id']
+    component_id = WorkspaceManager.MCP_WORKSPACE_COMPONENT_ID
 
-    def _get_workspace_meta() -> list[Mapping[str, Any]]:
-        metadata: list[Mapping[str, Any]] = []
-        for m in storage_client.branches.metadata('default'):
-            if m.get('key') == WorkspaceManager.MCP_META_KEY:
-                metadata.append(m)
-        return metadata
-
-    metas = _get_workspace_meta()
-    if metas:
-        pytest.fail(f'Expecting empty Keboola project {project_id}, but found {metas} in the default branch')
+    def _mcp_workspaces() -> list[Mapping[str, Any]]:
+        """MCP workspaces, discovered by their component id (no branch-metadata pointer)."""
+        return [w for w in storage_client.workspaces.list() if w.get('component') == component_id]
 
     workspaces = storage_client.workspaces.list()
     # ignore the static workspaces
@@ -50,7 +44,6 @@ async def dynamic_manager(
             f'{[{"id": w["id"], "name": w["name"]} for w in workspaces]}'
         )
 
-    component_id = WorkspaceManager.MCP_WORKSPACE_COMPONENT_ID
     existing_configs = list(storage_client.configurations.list(component_id=component_id))
     if existing_configs:
         pytest.fail(
@@ -61,24 +54,16 @@ async def dynamic_manager(
     yield await WorkspaceManager.create(keboola_client)
 
     LOG.info(f'Cleaning up workspaces in Keboola project with ID={project_id}')
-    metas = _get_workspace_meta()
-    if len(metas) > 1:
-        LOG.info(f'Multiple metadata entries found: {metas}')
-    for meta in metas:
+    # The MCP server no longer stamps a branch-metadata pointer; its workspaces are discovered
+    # (and cleaned up) by their component id.
+    for ws in _mcp_workspaces():
         try:
-            storage_client.workspaces.delete(meta['value'])
-            LOG.info(f'Deleted workspaces: {meta["value"]}')
+            storage_client.workspaces.delete(ws['id'])
+            LOG.info(f'Deleted workspace: {ws["id"]}')
         except requests.HTTPError:
-            LOG.exception(f'Failed to delete workspace {meta["value"]}')
-        try:
-            url = storage_client.branches.base_url.rstrip('/')
-            storage_client.branches._delete(f'{url}/branch/default/metadata/{meta["id"]}')
-            LOG.info(f'Deleted workspaces metadata: {meta["id"]}')
-        except requests.HTTPError as e:
-            LOG.exception(f'Failed to delete workspace metadata {meta["id"]}: {e}')
+            LOG.exception(f'Failed to delete workspace {ws["id"]}')
 
     # Clean up configurations created under the MCP workspace component
-    component_id = WorkspaceManager.MCP_WORKSPACE_COMPONENT_ID
     try:
         configs = storage_client.configurations.list(component_id=component_id)
         for cfg in configs:
--- a/src/keboola_mcp_server/clients/client.py
+++ b/src/keboola_mcp_server/clients/client.py
@@ -1,6 +1,7 @@
 """Keboola Storage API client wrapper."""
 
 import logging
+from pathlib import Path
 from typing import Any, Literal, Mapping, Sequence, TypeVar
 from urllib.parse import urlparse, urlunparse
 
@@ -266,6 +267,35 @@ class KeboolaClient:
     @property
     def storage_client(self) -> 'AsyncStorageClient':
         return self._storage_client
+
+    def step_up_storage_client(self, kubernetes_token_path: str) -> 'AsyncStorageClient':
+        """
+        Returns a Storage client that keeps this client's user token and additionally
+        sends the projected Kubernetes ServiceAccount JWT as the X-Kubernetes-Authorization
+        step-up header. Connection waives the permissions the user's token lacks on the
+        step-up-enabled actions (workspace / config / event creation) when the
+        ServiceAccount is authorized for them — no privileged token is minted and the
+        user's token stays the audited principal. The read-only write guard of the user's
+        Storage client is preserved; the header only widens server-side permissions.
+
+        The token file is read on each call so kubelet rotation needs no restart.
+
+        :param kubernetes_token_path: Path to the projected ServiceAccount token file.
+        :raises ValueError: If the token file is empty.
+        """
+        jwt = Path(kubernetes_token_path).read_text().strip()
+        if not jwt:
+            raise ValueError(f'Kubernetes ServiceAccount token file is empty: {kubernetes_token_path}')
+
+        headers = dict(self._headers or {})
+        headers['X-Kubernetes-Authorization'] = f'Bearer {jwt}'
+        return AsyncStorageClient.create(
+            root_url=self._storage_api_url,
+            token=self._token,
+            branch_id=self._branch_id,
+            headers=headers,
+            readonly=self._storage_client.raw_client.readonly,
+        )
 
     @property
     def jobs_queue_client(self) -> 'JobsQueueClient':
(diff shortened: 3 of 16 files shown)


Matovidlo:
@claude review

docs/kubernetes-sa-auth.md
```diff
@@ -0,0 +1,54 @@
...
+  must connect **at least once to the deployed (remote) MCP server**, so that their
+  workspace is created there (with proper billing configuration and credentials).
+  After that, the local MCP server works for querying: its `query_data` tool sends
+  queries to the in-infrastructure query-service, which attaches its own step-up
+  header when obtaining workspace credentials.
+- Read-only users who only ever use a locally-run MCP server and never connect to
+  the deployed one cannot get a workspace created — their own token lacks the
+  required permissions, and there is no ServiceAccount token to step up with.
+- The step-up is intentionally limited to workspace provisioning. Other write
+  operations still require adequate permissions on the user's own token.
```
Matovidlo: This appleis only for RO user

Matovidlo: **Fixed** ✅ in commit 3ed1853ef3fac57a44c324e3ce9d648777e8a3a9

Reworded the Drawbacks bullet to make the read-only-user framing explicit: "For read-only users the step-up is intentionally limited to workspace provisioning. Any other write operation still requires adequate permissions on the user's own token and keeps failing for a read-only user."

src/keboola_mcp_server/workspace.py
```diff
@@ -506,12 +518,59 @@ def __init__(self, client: KeboolaClient, workspace_schema: str | None = None):
...
+            raise ValueError(f'Kubernetes ServiceAccount token file is empty: {self._kubernetes_token_path}')
+
+        headers = dict(self._client.headers or {})
+        headers['X-Kubernetes-Authorization'] = f'Bearer {jwt}'
+        self._provisioning_client = AsyncStorageClient.create(
+            root_url=self._client.storage_api_url,
+            token=self._client.token,
+            branch_id=self._client.branch_id,
+            headers=headers,
+        )
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 3ed1853ef3fac57a44c324e3ce9d648777e8a3a9

The step-up provisioning client now propagates the read-only guard of the user's Storage client: `AsyncStorageClient.create(..., readonly=self._client.storage_client.raw_client.readonly)`. The step-up header widens server-side permissions only — it never bypasses the client-side write guard, so a `readonly=True` `KeboolaClient` can no longer produce a write-enabled step-up client.

tests/test_workspace.py
```diff
@@ -203,6 +204,119 @@ def _rebind(target_branch_id: str | None) -> Mock:
...
+    with patch.object(AsyncStorageClient, 'create', return_value=mock_writer) as mock_create:
+        await manager._create_ws()
+
+    # The provisioning client keeps the user's token and adds the step-up header ...
+    mock_create.assert_called_once_with(
+        root_url='https://connection.keboola.com',
+        token='user-token',
+        branch_id=None,
+        headers={'User-Agent': 'test', 'X-Kubernetes-Authorization': 'Bearer sa-jwt'},
+    )
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 3ed1853ef3fac57a44c324e3ce9d648777e8a3a9

The `assert_called_once_with(...)` in `test_workspace_creation_sends_step_up_header` now includes the `readonly` kwarg, and the test is parametrized over `readonly in (None, True)` to verify the guard is propagated in both modes.

docs/kubernetes-sa-auth.md
```diff
@@ -0,0 +1,53 @@
...
+## How it works
+
+When the `KBC_KUBERNETES_TOKEN_PATH` environment variable points at the projected
+ServiceAccount token (mounted into the deployed MCP server), workspace provisioning
+requests keep the **user's own Storage token** and additionally carry the SA JWT as
+the `X-Kubernetes-Authorization` header:
+
+- creating the billing configuration (`keboola.mcp-server-tool`),
+- creating the workspace itself,
+- writing the workspace ID into the branch metadata.
```
tomasfejfar: It's still in metadata?

docs/kubernetes-sa-auth.md
```diff
@@ -0,0 +1,53 @@
...
+
+Connection validates the JWT and, when the ServiceAccount is authorized for workspace
+provisioning, waives the permissions the user's token lacks —
+workspace provisioning works even for read-only project members. **No privileged token
+is ever minted**; every action runs under, and is audited to, the user's own token.
+Everything else (queries, reads, tools) is untouched.
```
tomasfejfar, line 26: RO user doesn't have access to the workspace, right? And for example query service can't run DGOM or such things. If Query service is only limited by the workspace privileges and would allow read-only user, it might easily lead to privilege escalation for RO users.

docs/kubernetes-sa-auth.md
```diff
@@ -0,0 +1,53 @@
...
+## Drawbacks / Limitations
+
+- **Locally-run MCP servers have no projected ServiceAccount token**, so the
+  step-up provisioning path only works on the MCP server deployed in Keboola
+  infrastructure. A user running the MCP server locally (stdio/custom deployment)
+  must connect **at least once to the deployed (remote) MCP server**, so that their
+  workspace is created there (with proper billing configuration and credentials).
+  After that, the local MCP server works for querying: its `query_data` tool sends
+  queries to the in-infrastructure query-service, which attaches its own step-up
+  header when obtaining workspace credentials.
```
tomasfejfar, line 53: So query service works. Does it have any ability to change data in project?

integtests/test_workspace.py
```diff
@@ -61,24 +54,16 @@ def _get_workspace_meta() -> list[Mapping[str, Any]]:
     yield await WorkspaceManager.create(keboola_client)
 
     LOG.info(f'Cleaning up workspaces in Keboola project with ID={project_id}')
-    metas = _get_workspace_meta()
-    if len(metas) > 1:
-        LOG.info(f'Multiple metadata entries found: {metas}')
-    for meta in metas:
+    # The MCP server no longer stamps a branch-metadata pointer; its workspaces are discovered
```
tomasfejfar, line 57: As expected - so it should be removed in the sa-auth document

src/keboola_mcp_server/clients/client.py
```diff
@@ -267,6 +268,35 @@ def headers(self) -> dict[str, Any] | None:
     def storage_client(self) -> 'AsyncStorageClient':
         return self._storage_client
 
+    def step_up_storage_client(self, kubernetes_token_path: str) -> 'AsyncStorageClient':
```
tomasfejfar, line 271: I think this should be part of the client, not a random "let's add headers" wrapper

src/keboola_mcp_server/tools/components/tools.py
```diff
@@ -173,7 +173,12 @@ def add_component_tools(mcp: KeboolaMcpServer) -> None:
         FunctionTool.from_function(
             run_sync_action,
             tags={COMPONENT_TOOLS_TAG},
-            annotations=ToolAnnotations(readOnlyHint=True),
+            # A sync action (testConnection, getTables, ...) does not persist changes to the project,
+            # but it is NOT a safe read: it POSTs an arbitrary component action with caller-supplied
+            # configData (reaching external systems, credentials, OAuth, etc.). readOnlyHint=True lets
+            # MCP clients auto-run it without user approval — a (limited) security gap. Mark it
+            # non-read-only so clients prompt for consent before executing it.
+            annotations=ToolAnnotations(readOnlyHint=False),
```
tomasfejfar: This change needs to be synced with kai release so that the workflow doesn't break for data apps and other tools using sync actions no?

docs/kubernetes-sa-auth.md
```diff
@@ -0,0 +1,53 @@
...
+- Read-only users who only ever use a locally-run MCP server and never connect to
+  the deployed one cannot get a workspace created — their own token lacks the
+  required permissions, and there is no ServiceAccount token to step up with.
+- For read-only users the step-up is intentionally limited to workspace provisioning.
```
tomasfejfar: Events as well, no?

src/keboola_mcp_server/mcp.py
```diff
@@ -283,7 +284,16 @@ async def create_session_state(
...
 
         try:
-            workspace_manager = await WorkspaceManager.create(client, config.workspace_schema)
+            # The Kubernetes ServiceAccount token path is read from the process environment
+            # only (KBC_KUBERNETES_TOKEN_PATH), never from `Config`/HTTP headers — it is a
+            # deployment-level credential of the MCP server itself and must not be
+            # overridable per request.
+            kubernetes_token_path = os.environ.get('KBC_KUBERNETES_TOKEN_PATH')
+            if kubernetes_token_path:
+                LOG.info('Kubernetes SA token configured; provisioning will send the step-up header.')
```
tomasfejfar: The k8s step up is on multiple places and only this one logs

src/keboola_mcp_server/workspace.py
```diff
@@ -614,14 +656,21 @@ async def _find_ws_by_id(self, workspace_id: str | int) -> _WspInfo | None:
...
-            if m.get('key') == meta_key and (raw_value := m.get('value')):
-                if (info := await self._find_ws_by_id(raw_value)) and info.readonly:
-                    return info
+        The MCP server creates its workspace under the MCP_WORKSPACE_COMPONENT_ID
+        component, so it is rediscovered by listing workspaces and matching that
+        component (plus read-only storage access). This needs only the read-only
+        `workspace_list` endpoint — no branch-metadata pointer, and therefore no
+        elevated metadata write that a read-only user's token would be denied.
+        """
+        for sapi_wsp_info in await self._client.storage_client.workspace_list():
```
tomasfejfar: the filtering didn't land yet?

src/keboola_mcp_server/workspace.py
```diff
@@ -651,9 +700,11 @@ async def _create_ws(self, *, timeout_sec: float = 300.0) -> _WspInfo | None:
         else:
             raise ValueError(f'Unexpected default backend: {default_backend}')
 
+        writer = await self._provisioning_storage_client()
```
tomasfejfar: why it's called writer? 😕

docs/kubernetes-sa-auth.md
Matovidlo: **Fixed** ✅ in commit b7fe2f46707866d46b3c5f59c536647c75e0a87b

Good catch — it isn't anymore. The workspace is now a regular workspace created under the `keboola.mcp-server-tool` configuration and rediscovered by listing workspaces and matching that component (see `_find_ws_in_branch`), so no branch-metadata pointer is written. Removed the stale 'writing the workspace ID into the branch metadata' step from the doc.

integtests/test_workspace.py
Matovidlo: **Fixed** ✅ in commit b7fe2f46707866d46b3c5f59c536647c75e0a87b

Right — the branch-metadata pointer is gone, so the doc no longer mentions it. Updated the sa-auth doc in the same commit to describe component-based discovery instead.

docs/kubernetes-sa-auth.md
Matovidlo: **Intentional** ℹ️

RO users don't get direct access to the workspace credentials — they reach the workspace only through Query service calls. The Query service is not a raw pass-through of workspace privileges: it exposes a constrained query API, and any metadata mutation (DGOM) is gated behind the explicit `refreshMetadataOnSuccess` field on the query endpoint, which only takes effect when the workspace has a configuration wired for it (transformations only, for now). So a plain RO query cannot escalate to a write; escalation would require a workspace deliberately set up with DGOM + refresh, which is not the RO default. See the query-service docs: https://query.canary-orion.keboola.dev/api/v1/documentation#/jobs/post_api_v1_branches__branchId__workspaces__workspaceId__queries

docs/kubernetes-sa-auth.md
Matovidlo: **Intentional** ℹ️

Only in a narrow, opt-in case: if the workspace has DGOM and a refresh is configured, the `refreshMetadataOnSuccess` flag on the query endpoint can refresh metadata (and that path currently applies to transformations only). For a normally-provisioned RO workspace there is no such setup, so the Query service cannot change project data.

src/keboola_mcp_server/clients/client.py
Matovidlo: **Question** ❓

It already is a method on `KeboolaClient` (`step_up_storage_client`), not a free-standing helper — it constructs a proper `AsyncStorageClient` with the same token/branch/readonly flags plus the step-up header. It's deliberately a separate factory rather than the default client because the step-up path only exists on the remotely-deployed MCP server (it requires the projected SA token); the default client must stay unchanged for local runs. Did you have a specific shape in mind — e.g. folding the header into `AsyncStorageClient.create` behind a flag? Happy to refactor if so.

src/keboola_mcp_server/tools/components/tools.py
Matovidlo: **Already addressed** ✅ in commit debbc60310e5eaa53e4f3b4c4d77481ef99b4267

Agreed this needs to be coordinated with the Kai release so sync-action-driven workflows (data apps, etc.) don't start prompting for approval unexpectedly. The change itself landed as a separate commit ("don't mark run_sync_action read-only") so it can be sequenced with the Kai rollout — it won't be released ahead of that coordination. Flagging to sync the release timing before this ships.

docs/kubernetes-sa-auth.md
Matovidlo: **Fixed** ✅ in commit b7fe2f46707866d46b3c5f59c536647c75e0a87b

Yes — storage event emission is also stepped up for RO users (see `errors.py` using `step_up_storage_client`), and those events aren't possible for an RO user on a purely local server either. Updated the doc line to say the step-up covers 'workspace provisioning and storage event emission'.

src/keboola_mcp_server/mcp.py
Matovidlo: **Fixed** ✅ in commit 328bf0e8c7aa4ba5e68aaa6b378de7a9718a093e

Right — the step-up is set up in several places and only this one logged, which was inconsistent. Removed this log entirely and kept a single `LOG.debug` at the point the provisioning step-up client is actually created (`_provisioning_storage_client`).

src/keboola_mcp_server/workspace.py
Matovidlo: **Already addressed** ✅

It did — the filtering is on the lines just below this `for`: it skips any workspace whose `component != MCP_WORKSPACE_COMPONENT_ID` and only returns one with `info.readonly` (workspace.py:669-672). GitHub's diff cut the hunk at the loop header; the component + read-only match is right after it.

src/keboola_mcp_server/workspace.py
Matovidlo: **Fixed** ✅ in commit 328bf0e8c7aa4ba5e68aaa6b378de7a9718a093e

Leftover misleading name — it's the provisioning storage client (user token + optional step-up header), not a 'writer'. Renamed the local to `provisioning_client` throughout `_create_ws`.

src/keboola_mcp_server/clients/client.py
tomasfejfar: Oh, sorry, I misread it.

src/keboola_mcp_server/tools/components/tools.py
tomasfejfar: The linked commit doesn't revert it so I don't understand the comment

src/keboola_mcp_server/workspace.py
tomasfejfar: I though that SAPI should have support for filtering workspaces by componentId directly in the request

Matovidlo: [https://api.keboola.com/?service=storage#storage/tag/component-configurations/GET/v2/storage/components/{componentId}/configs/{configurationId}/workspaces](<https://api.keboola.com/?service=storage#storage/tag/component-configurations/GET/v2/storage/components/%7BcomponentId%7D/configs/%7BconfigurationId%7D/workspaces>) We may use this

tomasfejfar: yes!

Matovidlo: **Fixed** ✅ in commit 569cf4bdbd6357ebffa18d10183fb155a1513951

Reworked this to fetch through the proper scoped endpoint rather than listing the whole project. `_find_ws_in_branch` now lists the MCP component's own configs and calls `GET branch/{branch}/components/{componentId}/configs/{configId}/workspaces` (new `AsyncStorageClient.workspace_list_for_config`) per config, matching read-only access — instead of `workspace_list()` over every workspace in the project and filtering client-side. So discovery needs only read access to the MCP component's configs. Tests updated accordingly.

src/keboola_mcp_server/tools/components/tools.py
Matovidlo: **Fixed** ✅ in commit 61d8b041a6d7e96796defb173093904ef385235f

Agreed — rather than ship it here and risk breaking sync-action workflows (data apps, Kai) before it's coordinated, I've reverted the change. run_sync_action is back to readOnlyHint=True (revert of debbc603, incl. TOOLS.md + annotation test). The consent-gating change can be reintroduced later, sequenced with a Kai release.

docs/kubernetes-sa-auth.md
```diff
@@ -0,0 +1,56 @@
...
+ServiceAccount token (mounted into the deployed MCP server), workspace provisioning
+requests keep the **user's own Storage token** and additionally carry the SA JWT as
+the `X-Kubernetes-Authorization` header:
+
+- creating the billing configuration (`keboola.mcp-server-tool`),
+- creating the workspace itself.
+
+The workspace is a regular workspace created under the `keboola.mcp-server-tool`
+configuration and is rediscovered by listing workspaces and matching that component —
+no branch-metadata pointer is written anymore.
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit f72f59ade358b42214e4740e65c3c2d3e0692328

Updated the doc to match `_find_ws_in_branch`: discovery now lists the `keboola.mcp-server-tool` component's configurations and fetches each config's workspaces through the config-scoped workspaces endpoint, not a project-wide workspace list.

docs/kubernetes-sa-auth.md
```diff
@@ -0,0 +1,56 @@
...
+Connection validates the JWT and, when the ServiceAccount is authorized for workspace
+provisioning, waives the permissions the user's token lacks —
+workspace provisioning works even for read-only project members. **No privileged token
+is ever minted**; every action runs under, and is audited to, the user's own token.
+Everything else (queries, reads, tools) is untouched.
+
+Key properties:
+
+- The token file is read per provisioning flow — kubelet rotation is picked up
+  automatically (no caching across flows).
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit f72f59ade358b42214e4740e65c3c2d3e0692328

Corrected the 'no caching across flows' bullet. The step-up provisioning client is cached for the `WorkspaceManager` (session) lifetime, so the token file is read once per session — not per flow. Provisioning happens at most once per session, well within the token rotation window; a new session re-reads the file, so kubelet rotation is still picked up without a restart.

src/keboola_mcp_server/workspace.py
```diff
@@ -577,12 +587,44 @@ def __init__(self, client: KeboolaClient, workspace_schema: str | None = None):
...
+        additionally sends the projected SA JWT as the X-Kubernetes-Authorization
+        step-up header — Connection waives permissions the user's token lacks when the
+        ServiceAccount is authorized for workspace provisioning. No privileged token
+        is ever minted; the audit trail stays on the user's token.
+        Otherwise the user's own Storage client is used unchanged.
+
+        The step-up client is cached for this manager's lifetime — provisioning happens
+        at most once per manager, well within the projected token's rotation window.
+        The token file is read here (per provisioning flow), so kubelet rotation needs
+        no restarts.
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit f72f59ade358b42214e4740e65c3c2d3e0692328

Reworded the `_provisioning_storage_client` docstring: the client is cached for the manager's lifetime, so the token file is read once (when the client is first built), not on every provisioning attempt; a fresh manager (new session) re-reads it for rotation.

src/keboola_mcp_server/workspace.py
```diff
@@ -764,14 +818,14 @@ async def _get_workspace(self) -> _Workspace:
...
-            # All tokens share the same read-only workspace
-            # Race conditions during initialization are acceptable (last-write-wins)
-            meta = await self._client.storage_client.branch_metadata_update({self.MCP_META_KEY: info.id})
-            LOG.info(f'Set metadata in the default branch: {meta}')
-            # use the newly created workspace
+            # All tokens share the same read-only workspace, rediscovered by its
+            # component id (see _find_ws_in_branch) — no branch-metadata pointer is
+            # written, so no elevated metadata write is needed. Concurrent first-use
+            # may create more than one workspace; that is acceptable, _find_ws_in_branch
+            # returns the first match on the next lookup.
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit f72f59ade358b42214e4740e65c3c2d3e0692328

Updated the PR summary — the branch-metadata pointer is gone. It now states the shared MCP workspace is rediscovered via the component's config-scoped workspaces endpoint (`_find_ws_in_branch`), and the step-up header is attached only on the billing-config create and the workspace create (no metadata write).

tomasfejfar (approved):

PR #564, opened 2026-06-05
AJDA-2810: Fetch conditional-flow schema live from the Developer Portal

MiroCillik:
## Description

**Linear**: [AJDA-2810](https://linear.app/keboola/issue/AJDA-2810/cf-variables-update-mcp)
**RFC**: `feature_spec/conditional_flow_schema_from_devportal/RFC.md` (approved)

### Change Type

- [ ] Major (breaking changes, significant new features)
- [x] Minor (new features, enhancements, backward compatible)
- [ ] Patch (bug fixes, small improvements, no new features)

### Summary

The MCP server shipped a **static, bundled copy** of the `keboola.flow` (conditional-flow) JSON schema and read it for both `get_flow_schema` and conditional-flow validation. Because it was bundled, the server could not pick up new schema fields (e.g. the upcoming **Variables** capability, AJDA-2351) until someone re-bundled the file and cut a release.

This PR sources the conditional-flow schema **live from the Developer Portal**:

- Adds `resolve_flow_schema(client, flow_type)` in `tools/flow/utils.py` — for `keboola.flow` it fetches the live `configuration_schema` via the existing `fetch_component()` path (AI Service + Storage 404-fallback) and **caches it per session** on `KeboolaClient` (mirrors `_features_cache`; never stale across runs). Legacy `keboola.orchestrator` stays bundled.
- On fetch failure (non-404 HTTP error, network/transport error, or empty schema) it **hard-fails** with a clear, agent-recoverable message — there is no stale bundled fallback for conditional flows.
- `get_schema_as_markdown` is now async/client-aware; `get_flow_schema` awaits it (tool signature unchanged).
- `validate_flow_configuration_against_schema` gains an optional `schema` param; the three conditional call sites (`create_flow`, `create_conditional_flow`, `update_flow_internal`) resolve the schema and validate against it **before** any Storage write.
- **Removes** the bundled `resources/conditional-flow-schema.json`, the `ConfigurationSchemaResources.FLOW` enum member, and the `FLOW_SCHEMAS` conditional entry. Legacy orchestrator and storage schemas are untouched.

## Testing

- [x] Unit tests: resolver (live fetch, cache hit, empty-schema/HTTP/network hard-fail), `get_flow_schema` live (mocked), conditional validation against an offline fixture; legacy paths unchanged. Full `tox` green (pytest + black + isort + flake8 + check-tools-docs; TOOLS.md unchanged).
[...]

PR #574, opened 2026-06-09
feat(flow): AJDA-2860 support CF variables (variableOverrides + JMESPath) in flow models

MiroCillik:
## Description

**Linear**: [AJDA-2860](https://linear.app/keboola/issue/AJDA-2860) (follow-up to AJDA-2810, parent AJDA-2290, related AJDA-2351)
**RFC**: `feature_spec/conditional_flow_variables_model_support/RFC.md` (included in this PR)

### Change Type

### Summary

AJDA-2810 (#564) made the MCP server source the `keboola.flow` schema live, so `get_flow_schema` and jsonschema validation surface the new **Variables** capability. But the create/update path also routes `phases`/`tasks` through the Pydantic models in `tools/flow/model.py` (`get_flow_configuration()` → `ConditionalFlowTask.model_validate(...).model_dump(...)`) **before** schema validation, and those models didn't carry the variables shapes — so flows using variables were silently corrupted or rejected:

1. **`variableOverrides` was silently dropped.** `JobTaskConfiguration` didn't model it and the models used Pydantic's default `extra='ignore'`, so a job task passing a flow variable into a component lost the wiring with no error.
2. **JMESPath `value`s were rejected.** `TaskCondition.value` was a closed `Literal` of ~13 fixed paths, so a dynamic value like `sum(job.result.output.tables[].metrics[?name=='importedRowsCount'][].value)` raised a `ValidationError` — breaking both the `variable` task `source.value` and JMESPath-driven phase-transition conditions.

This PR fixes the models so the variables feature works end-to-end through `create_conditional_flow` / `update_flow`:

- Add `variableOverrides: Optional[list[str]]` to `JobTaskConfiguration`.
- Relax `TaskCondition.value` from the closed `Literal` to a free-form `str` (the former enum paths are kept as documented examples; JMESPath now allowed). This also covers the variable-task `source`, since it discriminates onto `TaskCondition`.
- Set `model_config = ConfigDict(extra='allow')` on the three task-configuration models (`Job`/`Notification`/`Variable`) so fields the live schema may add in future pass through instead of being silently dropped — matching AJDA-2810's "don't lag the live schema" goal. Validation against the live schema remains the real gate.

## Testing

- [x] New round-trip unit tests in `tests/tools/flow/test_utils.py` (`TestConditionalFlowVariablesRoundTrip`) built from a real CF-variables flow: assert `get_flow_configuration` preserves `variableOverrides`, the JMESPath `value` in the variable `source` and in a phase `condition`, that `validate_flow_structure` accepts it, and that an unknown future job field is preserved (extra='allow'). Full `tox` green (pytest, black, isort, flake8, check-tools-docs; TOOLS.md unchanged).

## Checklist

- [x] Self-review completed
- [ ] Integration tests added/updated (real variables create/validate can be added once exercised on a CF-enabled stack)
- [x] Project version bumped (minor: 1.67.0 → 1.68.0) + `uv.lock` synced
- [x] RFC included

## Release Notes

- **Justification**
    - **Why:** After AJDA-2810 the server advertises the conditional-flow Variables fields, but the MCP create/update tools couldn't actually build a variables flow — `variableOverrides` was silently dropped and JMESPath dynamic values were rejected by the local Pydantic models before schema validation.
    - **What:** Extend the flow models to carry `variableOverrides` and free-form/JMESPath condition values, and tolerate future schema fields, so agents can create/update flows that set and consume flow variables.
- **Plans for Customer Communication**
    - Enables the CF Variables capability via MCP; coordinate with the Conditional Flows feature announcement.
- **Impact Analysis**
    - Affects conditional flows only. `extra='allow'` widens what the task-config models accept; the live `keboola.flow` schema validation (added in AJDA-2810) remains the authoritative gate, so malformed configs are still rejected there. No change to legacy orchestrator flows or storage.
- **Deployment Plan**
    - Continuous; independent change. Verify a variables flow create/validate on a CF-enabled stack after deploy.
- **Rollback Plan**
    - Two-way door: revert restores the previous model definitions; no data migration.
- **Post-Release Support Plan**
    - None beyond standard; if a variables flow fails to validate, the live-schema validation error is the source of truth.

### DIFF
--- a/src/keboola_mcp_server/tools/flow/model.py
+++ b/src/keboola_mcp_server/tools/flow/model.py
@@ -6,7 +6,7 @@ import logging
 from datetime import datetime
 from typing import Annotated, Any, Literal, Optional, Union
 
-from pydantic import AliasChoices, BaseModel, Field, ValidationError
+from pydantic import AliasChoices, BaseModel, ConfigDict, Field, ValidationError
 
 from keboola_mcp_server.clients.client import ORCHESTRATOR_COMPONENT_ID, FlowType, get_metadata_property
 from keboola_mcp_server.clients.storage import APIFlowResponse
@@ -146,21 +146,16 @@ class TaskCondition(BaseModel):
 
     type: Literal['task'] = Field(description='Condition type')
     task: str = Field(description='ID of the task to evaluate, or "*" when used with phase operators')
-    value: Literal[
-        'taskId',
-        'phaseId',
-        'status',
-        'job.id',
-        'job.componentId',
-        'job.configId',
-        'job.status',
-        'job.result',
-        'job.startTime',
-        'job.endTime',
-        'job.duration',
-        'job.result.output.tables',
-        'job.result.message',
-    ] = Field(description='Property path to retrieve from the task context')
+    value: str = Field(
+        description=(
+            'Property path or JMESPath expression to retrieve from the task context. Common simple '
+            "paths: 'taskId', 'phaseId', 'status', 'job.id', 'job.componentId', 'job.configId', "
+            "'job.status', 'job.result', 'job.startTime', 'job.endTime', 'job.duration', "
+            "'job.result.output.tables', 'job.result.message'. JMESPath expressions over the job "
+            'result are also supported, for example: '
+            "sum(job.result.output.tables[].metrics[?name=='importedRowsCount'][].value)"
+        )
+    )
 
 
 class PhaseCondition(BaseModel):
@@ -248,6 +243,10 @@ ConditionObject = Union[
 class JobTaskConfiguration(BaseModel):
     """Job task configuration."""
 
+    # Forward-compat: keep fields the live keboola.flow schema may add (and that we don't model yet)
+    # instead of silently dropping them on the model_dump round-trip in get_flow_configuration().
+    model_config = ConfigDict(extra='allow')
+
     type: Literal['job'] = Field(description='Task type')
     component_id: str = Field(description='Component ID', alias='componentId')
     config_id: Optional[str] = Field(default=None, description='Configuration ID', alias='configId')
@@ -255,6 +254,11 @@ class JobTaskConfiguration(BaseModel):
     mode: Literal['run'] = Field(description='Execution mode')
     delay: Optional[Union[str, int]] = Field(default=None, description='Initial delay in seconds')
     retry: Optional[RetryConfiguration] = Field(default=None, description='Retry configuration')
+    variable_overrides: Optional[list[str]] = Field(
+        default=None,
+        description='Names of flow variables to pass into this job as variable overrides',
+        alias='variableOverrides',
+    )
 
 
 class NotificationRecipient(BaseModel):
@@ -266,6 +270,8 @@ class NotificationRecipient(BaseModel):
 
 class NotificationTaskConfiguration(BaseModel):
     """Notification task configuration."""
+
+    model_config = ConfigDict(extra='allow')
 
     type: Literal['notification'] = Field(description='Task type')
     recipients: list[NotificationRecipient] = Field(description='List of notification recipients', min_length=1)
@@ -284,6 +290,8 @@ VariableSourceObject = Annotated[
 
 class VariableTaskConfiguration(BaseModel):
     """Variable task configuration."""
+
+    model_config = ConfigDict(extra='allow')
 
     type: Literal['variable'] = Field(description='Task type')
     name: str = Field(description='Variable name')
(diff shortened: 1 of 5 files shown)


src/keboola_mcp_server/tools/flow/model.py
```diff
@@ -146,21 +146,16 @@ class TaskCondition(BaseModel):
...
-        'job.configId',
-        'job.status',
-        'job.result',
-        'job.startTime',
-        'job.endTime',
-        'job.duration',
-        'job.result.output.tables',
-        'job.result.message',
-    ] = Field(description='Property path to retrieve from the task context')
+    value: str = Field(
```
MiroCillik, line 161: Relaxing `value` from the closed `Literal` to free-form `str` is the right fix for JMESPath, but it removes the only client-side validation of this field — correctness now rests entirely on the live `keboola.flow` schema (AJDA-2810) as the gate, so a mistyped path is caught server-side only. Intended per the RFC; just flagging it. (Good that `PhaseCondition.value` stays an enum — that one really is a fixed set.)

src/keboola_mcp_server/tools/flow/model.py
```diff
@@ -248,13 +243,22 @@ class ArrayCondition(BaseModel):
 class JobTaskConfiguration(BaseModel):
     """Job task configuration."""
 
+    # Forward-compat: keep fields the live keboola.flow schema may add (and that we don't model yet)
+    # instead of silently dropping them on the model_dump round-trip in get_flow_configuration().
+    model_config = ConfigDict(extra='allow')
```
MiroCillik: `extra='allow'` is the correct forward-compat choice, but note the side effect: a misspelled field such as `variableOverride` (singular) now passes through silently as extra data instead of being rejected — client-side field-name typos won't surface here, only via live-schema validation. Acceptable per the RFC trade-off; worth keeping in mind.

MiroCillik (commented):
Inline review notes (self-review; Copilot generated no comments). All are non-blocking — the change is correct and well-scoped. Summary: model relaxation is correctly targeted; `extra='allow'` trade-off is acceptable; main suggestion is a read/display-path test (RFC item 4).

src/keboola_mcp_server/tools/flow/model.py
```diff
@@ -267,6 +271,8 @@ class NotificationRecipient(BaseModel):
 class NotificationTaskConfiguration(BaseModel):
     """Notification task configuration."""
 
+    model_config = ConfigDict(extra='allow')
```
MiroCillik: Consistency nit: `JobTaskConfiguration` carries the `# Forward-compat: …` comment explaining `extra='allow'`, but this model (and `VariableTaskConfiguration` below) don't. Consider adding the same one-line rationale at each site so the intent is obvious.

tests/tools/flow/test_utils.py
```diff
@@ -645,3 +645,104 @@ async def test_returns_bundled_schema_for_legacy(self, mocker):
...
+                'componentId': 'keboola.snowflake-transformation',
+                'configId': 'abc',
+                'variableOverrides': ['importedRowsSum'],
+            },
+        },
+    ]
+    return phases, tasks
+
+
+class TestConditionalFlowVariablesRoundTrip:
```
MiroCillik, line 717: Solid round-trip coverage. One gap vs. RFC item 4: the read/display path (`Flow.from_api_response` / `get_flows`) uses these same models. Consider a small test that an API response containing `variableOverrides` + a JMESPath `value` round-trips through `Flow.from_api_response` without dropping fields, to guard the read direction too.

Matovidlo (approved):

src/keboola_mcp_server/tools/flow/model.py
odinuv: Souhlasím s tím aby se to sem dalo, ale mělo by se to dát univerzálně - proč je to teď na JobTaskConfiguration a ne na PhaseCondtion? Čili udělal bych asi BaseExtraModel, který se použije na všechny ty konfigurační nody

MiroCillik: Good call — done in bc5b7d84. Introduced a shared `BaseExtraModel(BaseModel)` with `extra='allow'` and applied it to **all** conditional-flow config nodes: every condition (incl. `PhaseCondition`), the retry models, task configs, transition, task and phase. Removed the three ad-hoc per-class `model_config` declarations. Added a test that an unknown field on a condition node round-trips too.

Scoped it to the conditional-flow (keboola.flow) config nodes since the forward-compat motivation is the *live* schema; I kept the MCP output models (`Flow`, `FlowToolOutput`, `GetFlows*`) and the legacy orchestrator models (`FlowPhase/FlowTask/FlowConfiguration`, whose schema is bundled & static) strict. Happy to extend to legacy too if you'd prefer truly everything.

PR #578, opened 2026-06-11
feat: cancel running query when the MCP client disconnects mid-query_data (AI-3204)

Matovidlo:
## Description

**Linear**: [AI-3204](https://linear.app/keboola/issue/AI-3204/cancel-query-data-queries-when-chat-is-stopped)

### Change Type

### Summary

Stopping a Kai chat during `query_data` did not cancel the underlying Query Service job — long scans (a reported 400GB case) kept running with no way to stop them from the UI.

This PR makes the dropped HTTP connection itself the cancellation signal:

1. **`_Workspace.execute_query`** wraps the job-status polling loop in an `asyncio.CancelledError` handler: on cancellation it fires `cancel_job` on the Query Service inside `asyncio.shield(...)`, so the backend cancel survives the task teardown. Since polling is now unified in the base class, this covers Snowflake and BigQuery alike.
2. **`query_data`** races the query task against `_watch_for_http_disconnect()`, which polls the underlying ASGI request's `is_disconnected()`. When the client tears down its `tools/call` connection (Kai kills the sandbox SDK process on STOP; the kai-agent proxy now propagates the abort upstream — see companion PR keboola/ui#6349), the query task is cancelled, tripping the handler from (1).

Why disconnect and not MCP `notifications/cancelled`: the server runs streamable-HTTP in **stateless** mode (`stateless_http=True`), so the cancel notification arrives on a fresh transport instance and never reaches the in-flight tool call. The connection drop is the only signal that reliably reaches the running request. An earlier attempt at this watcher (on the abandoned `martinvasko-ai-3204-cancel-query_data-queries-when-chat-is-stopped` branch) "never fired" only because the kai-agent proxy held its upstream socket open after the sandbox died — that is fixed on the Kai side now.

Safety net: the existing 5-minute `_QUERY_TIMEOUT` in `execute_query` still hard-cancels the QS job (via `_cancel_job_with_timeout`) even if disconnect detection never fires, so an orphaned query can never run longer than 5 minutes.

The `notifications/progress` emission with `_meta.keboola.queryJobId` / `cancellationUrl` stays: it is informational for other MCP clients and keeps the SSE stream non-idle (which helps disconnect detection through load balancers). It is no longer load-bearing for Kai cancellation.

## Testing

### Optional testing

## Checklist

- [x] Self-review completed

### DIFF
--- a/src/keboola_mcp_server/tools/sql.py
+++ b/src/keboola_mcp_server/tools/sql.py
@@ -108,9 +108,12 @@ async def _execute_watching_disconnect(
     # query either way, but keep the log honest so a watcher bug isn't disguised as a disconnect.
     watcher_exc = disconnect_task.exception()
     if watcher_exc is not None:
+        # Pass an explicit (type, exc, tb) tuple: stdlib logging treats a truthy non-tuple
+        # `exc_info` as True and falls back to sys.exc_info(), which is empty here, so the
+        # watcher's traceback would otherwise be lost.
         LOG.warning(
             f'HTTP disconnect watcher for query_data "{query_name}" failed; cancelling underlying query',
-            exc_info=watcher_exc,
+            exc_info=(type(watcher_exc), watcher_exc, watcher_exc.__traceback__),
         )
     else:
         LOG.info(f'HTTP client disconnected during query_data "{query_name}"; cancelling underlying query')
--- a/src/keboola_mcp_server/tools/sql.py
+++ b/src/keboola_mcp_server/tools/sql.py
@@ -3,10 +3,9 @@ import contextlib
 import csv
 import logging
 from io import StringIO
-from typing import Annotated
+from typing import Annotated, Awaitable
 
 from fastmcp import Context, FastMCP
-from fastmcp.server.dependencies import get_http_request
 from fastmcp.tools import FunctionTool
 from mcp.types import (
     ProgressNotification,
@@ -16,9 +15,11 @@ from mcp.types import (
     ToolAnnotations,
 )
 from pydantic import BaseModel, Field
+from starlette.requests import Request
 
 from keboola_mcp_server.errors import tool_errors
-from keboola_mcp_server.workspace import JobSubmittedInfo, SqlSelectData, WorkspaceManager
+from keboola_mcp_server.mcp import get_http_request_or_none
+from keboola_mcp_server.workspace import JobSubmittedInfo, QueryResult, SqlSelectData, WorkspaceManager
 
 LOG = logging.getLogger(__name__)
 
@@ -26,12 +27,12 @@ SQL_TOOLS_TAG = 'sql'
 MAX_ROWS = 1_000
 MAX_CHARS = 50_000
 # How often to check whether the HTTP client has disconnected during a long query.
-# Mirrors the 1 s job-poll cadence in `_SnowflakeWorkspace.execute_query`.
+# Mirrors the 1 s job-poll cadence in `_Workspace.execute_query`.
 _DISCONNECT_POLL_INTERVAL = 1.0
 
 
-async def _watch_for_http_disconnect(poll_interval: float = _DISCONNECT_POLL_INTERVAL) -> None:
-    """Return when the underlying HTTP request is torn down, or block forever otherwise.
+async def _watch_for_http_disconnect(request: Request, poll_interval: float | None = None) -> None:
+    """Return when the underlying HTTP `request` is torn down.
 
     In stateless streamable-HTTP mode (`stateless_http=True` in `cli.py`), the MCP
     `notifications/cancelled` payload arrives on a fresh transport instance and cannot
@@ -41,27 +42,102 @@ async def _watch_for_http_disconnect(poll_interval: float = _DISCONNECT_POLL_INT
     hit "stop" in Kai, lost network) and trigger the same cancellation path we
     already have for SDK-driven cancels.
 
-    Returns silently when disconnect is detected. Blocks forever if there is no HTTP
-    request bound (e.g. stdio transport, background workers) — in that case the caller
-    will only stop on normal task completion or its own cancellation.
+    Only started when there is an HTTP request bound (see `query_data`); on stdio /
+    background workers there is nothing to watch, so the caller skips the race entirely.
 
     Any error from `is_disconnected()` is treated as "still connected" so a transient
     ASGI hiccup never cancels an otherwise-working query.
-    """
-    try:
-        request = get_http_request()
-    except RuntimeError:
-        # No HTTP request context — never fire (e.g. stdio transport).
-        await asyncio.Event().wait()
-        return  # unreachable; satisfies the type checker
 
+    `poll_interval` is resolved at call time (not bound as a default) so tests can
+    patch `_DISCONNECT_POLL_INTERVAL` and have it take effect here.
+    """
+    interval = poll_interval if poll_interval is not None else _DISCONNECT_POLL_INTERVAL
     while True:
         try:
             if await request.is_disconnected():
                 return
         except Exception:
             LOG.debug('HTTP is_disconnected() check failed; treating as still-connected', exc_info=True)
-        await asyncio.sleep(poll_interval)
+        await asyncio.sleep(interval)
+
+
+async def _drain(*tasks: asyncio.Task) -> None:
+    """Cancel and drain `tasks` under a shield so their cancellation cleanup — notably the
+    backend `cancel_job` in `_Workspace.execute_query` — finishes even when `query_data` is
+    itself being torn down.
+
+    The shield protects the in-flight cleanup, but the `await` here is deliberately NOT
+    protected: if `query_data` is cancelled at this boundary, the `CancelledError` propagates
+    out (the caller's cancellation must win — we must not go on to return a result or raise a
+    synthetic error as if nothing happened) while the drain runs to completion in the background.
+    """
+    await asyncio.shield(asyncio.gather(*(_cancel_and_drain(t) for t in tasks)))
+
+
+async def _execute_watching_disconnect(
+    query_coro: Awaitable[QueryResult], request: Request, query_name: str
+) -> QueryResult:
+    """Run `query_coro`, cancelling it — and thus the backend job, via `execute_query`'s
+    `CancelledError` handler — if the HTTP client disconnects first.
+
+    Returns the query result when the query wins the race. Raises `ValueError('Query was
+    cancelled')` if the client disconnected (or the disconnect watcher itself failed) before the
+    query finished: a plain `ValueError` keeps the cancelled call on the normal `@tool_errors`
+    path, so it is logged as an error (not a success) and the client still gets a response — a
+    raised `CancelledError` would be a `BaseException` that bypasses the decorator entirely.
+    """
+    query_task = asyncio.create_task(query_coro)
+    disconnect_task = asyncio.create_task(_watch_for_http_disconnect(request))
+    try:
+        done, _pending = await asyncio.wait([query_task, disconnect_task], return_when=asyncio.FIRST_COMPLETED)
+    except BaseException:
+        # `query_data` itself was cancelled (e.g. SDK-driven MCP cancellation on a non-stateless
+        # transport). Drain both tasks so `execute_query`'s CancelledError handler runs long enough
+        # to propagate `cancel_job` and neither task leaks as pending; suppress the drain's own
+        # CancelledError so the original exception propagates via `raise`.
+        with contextlib.suppress(asyncio.CancelledError):
+            await _drain(query_task, disconnect_task)
+        raise
+
+    if query_task in done:
+        # The query won; the watcher is still pending, so drain it.
+        await _drain(disconnect_task)
+        return query_task.result()
+
+    # The watcher won: either the client disconnected, or the watcher itself raised. Cancel the
+    # query either way, but keep the log honest so a watcher bug isn't disguised as a disconnect.
+    watcher_exc = disconnect_task.exception()
+    if watcher_exc is not None:
+        LOG.warning(
+            f'HTTP disconnect watcher for query_data "{query_name}" failed; cancelling underlying query',
+            exc_info=watcher_exc,
+        )
+    else:
+        LOG.info(f'HTTP client disconnected during query_data "{query_name}"; cancelling underlying query')
+    await _drain(query_task)
+    raise ValueError('Query was cancelled')
+
+
+async def _cancel_and_drain(task: asyncio.Task) -> None:
+    """Cancel a task and await its unwind, suppressing the resulting error.
+
+    Draining (rather than fire-and-forget `task.cancel()`) lets any shielded
+    cleanup the task runs on cancellation — notably the backend `cancel_job` in
+    `_Workspace.execute_query` — finish before we move on. `KeyboardInterrupt` /
+    `SystemExit` are intentionally NOT suppressed.
+
+    Only the expected `CancelledError` is swallowed. Any other exception raised
+    while the task unwinds is a real bug in the cancellation cleanup path (e.g. the
+    shielded backend cancel) — log it with a traceback rather than silently dropping
+    it, but still don't re-raise so draining stays best-effort.
+    """
+    task.cancel()
+    try:
+        await task
+    except asyncio.CancelledError:
+        pass
+    except Exception:
+        LOG.warning('Unexpected error while draining cancelled task', exc_info=True)
 
 
 def _client_progress_token(ctx: Context) -> ProgressToken | None:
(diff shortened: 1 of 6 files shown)


AdamVyborny:
@Matovidlo máš tu nějaký copilot commenty a je potřeba rebase

src/keboola_mcp_server/tools/sql.py
```diff
@@ -22,6 +25,60 @@
...
+    """Cancel a task and await its unwind, suppressing the resulting error.
+
+    Draining (rather than fire-and-forget `task.cancel()`) lets any shielded
+    cleanup the task runs on cancellation — notably the backend `cancel_job` in
+    `_Workspace.execute_query` — finish before we move on. `KeyboardInterrupt` /
+    `SystemExit` are intentionally NOT suppressed.
+    """
+    task.cancel()
+    with contextlib.suppress(asyncio.CancelledError, Exception):
+        await task
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 51f744f2017b86410f4f51c7f9a3404eb98d2dbd

`_cancel_and_drain` no longer suppresses all exceptions. It now only swallows the expected `asyncio.CancelledError`; any other `Exception` raised while the task unwinds (including in the shielded backend-cancel path) is logged at WARNING with a traceback (`exc_info=True`) instead of being silently dropped. `KeyboardInterrupt`/`SystemExit` still propagate. Draining stays best-effort (we don't re-raise).

src/keboola_mcp_server/tools/sql.py
```diff
@@ -175,12 +232,39 @@ async def query_data(
...
+    disconnect_task = asyncio.create_task(_watch_for_http_disconnect())
+    try:
+        done, _pending = await asyncio.wait(
+            [query_task, disconnect_task],
+            return_when=asyncio.FIRST_COMPLETED,
+        )
+    except BaseException:
+        query_task.cancel()
+        disconnect_task.cancel()
+        raise
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 51f744f2017b86410f4f51c7f9a3404eb98d2dbd

The outer `except BaseException` now drains both tasks instead of just cancelling them. It runs `asyncio.gather(_cancel_and_drain(query_task), _cancel_and_drain(disconnect_task))` wrapped in `asyncio.shield(...)` so the drain completes under the outer cancellation that is unwinding us — letting `execute_query`'s `CancelledError` handler propagate `cancel_job` and leaving no pending task (no "Task was destroyed but it is pending" warnings). The `CancelledError` that the shielded await itself re-raises is suppressed so the original exception still propagates via `raise`.

src/keboola_mcp_server/tools/sql.py
```diff
@@ -175,12 +232,39 @@ async def query_data(
...
+
+    if query_task not in done:
+        LOG.info(f'HTTP client disconnected during query_data "{query_name}"; cancelling underlying query')
+        await _cancel_and_drain(query_task)
+        raise asyncio.CancelledError(f'HTTP client disconnected during query_data "{query_name}"')
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 51f744f2017b86410f4f51c7f9a3404eb98d2dbd

In the disconnect path we now consume `disconnect_task`'s outcome before raising: `disconnect_task.exception()` is inspected and, if `_watch_for_http_disconnect` finished with an exception, it is logged at WARNING with `exc_info=...`. This prevents the "Task exception was never retrieved" warning and the silent loss of the error. Added a regression test (`test_query_data_logs_when_disconnect_watcher_raises`) covering this branch.

src/keboola_mcp_server/tools/sql.py
```diff
@@ -175,12 +241,59 @@ async def query_data(
...
+        # (rather than returning on a detected disconnect), surface it in the logs instead
+        # of leaking a "Task exception was never retrieved" warning and silently losing it.
+        disconnect_exc = disconnect_task.exception()
+        if disconnect_exc is not None:
+            LOG.warning(f'HTTP disconnect watcher for query_data "{query_name}" failed', exc_info=disconnect_exc)
+        LOG.info(f'HTTP client disconnected during query_data "{query_name}"; cancelling underlying query')
+        await _cancel_and_drain(query_task)
+        raise asyncio.CancelledError(f'HTTP client disconnected during query_data "{query_name}"')
+
+    await _cancel_and_drain(disconnect_task)
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 59ffd896390bb953ffd75cf035873a1bb3787755

Wrapped the post-query drain in `asyncio.shield`: `await asyncio.shield(_cancel_and_drain(disconnect_task))`. Now, if `query_data` is cancelled at this await boundary (client cancels right after the query completes), the `CancelledError` propagates out of `query_data` instead of being swallowed inside `_cancel_and_drain` — so we no longer return a result despite cancellation. The watcher cleanup still runs to completion best-effort in the background under the shield. Added regression test `test_query_data_propagates_cancel_during_post_query_drain` that cancels the task at the drain boundary and asserts both that `CancelledError` propagates and that the shielded drain still finishes.

src/keboola_mcp_server/tools/sql.py
```diff
@@ -175,12 +241,65 @@ async def query_data(
...
+        # We are here because `disconnect_task` finished first, so it is already done.
+        # Consume its outcome: if `_watch_for_http_disconnect` finished with an exception
+        # (rather than returning on a detected disconnect), surface it in the logs instead
+        # of leaking a "Task exception was never retrieved" warning and silently losing it.
+        disconnect_exc = disconnect_task.exception()
+        if disconnect_exc is not None:
+            LOG.warning(f'HTTP disconnect watcher for query_data "{query_name}" failed', exc_info=disconnect_exc)
+        LOG.info(f'HTTP client disconnected during query_data "{query_name}"; cancelling underlying query')
+        await _cancel_and_drain(query_task)
+        raise asyncio.CancelledError(f'HTTP client disconnected during query_data "{query_name}"')
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit b6d672d632c879e07f5ac9c4c045976345d8137c

The disconnect-first branch now distinguishes the two cases. When `disconnect_task.exception()` is None we keep the INFO "HTTP client disconnected ... cancelling underlying query" log and raise `CancelledError("HTTP client disconnected during query_data ...")`. When the watcher actually failed with an exception, we instead log at WARNING "HTTP disconnect watcher ... failed; cancelling underlying query" (with `exc_info`) and raise `CancelledError("HTTP disconnect watcher failed during query_data ...")`. The query is still cancelled either way, but the reason is now honest so a watcher bug isn't disguised as a client disconnect.

tests/tools/test_sql.py
```diff
@@ -1142,6 +1143,134 @@ async def test_cancellation_polling_terminal_statuses(
...
+        drain_started = asyncio.Event()
+        drain_finished = asyncio.Event()
+        release = asyncio.Event()
+
+        async def slow_drain(task: asyncio.Task) -> None:
+            task.cancel()
+            drain_started.set()
+            await release.wait()
+            drain_finished.set()
+
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit b6d672d632c879e07f5ac9c4c045976345d8137c

`slow_drain` now awaits the cancelled task after `release` (under `contextlib.suppress(asyncio.CancelledError)`), mirroring the real `_cancel_and_drain`. This drains the task so it can't linger pending and emit intermittent "Task exception was never retrieved" / pending-task warnings at the end of the run, while preserving the start/release/finish coordination the test relies on.

src/keboola_mcp_server/workspace.py
```diff
@@ -249,38 +249,52 @@ async def execute_query(
...
-                reason = f'Query timeout exceeded after {elapsed_time:.2f} seconds'
-                cancellation_confirmed, query_completed = await self._cancel_job_with_timeout(job_id, reason)
-
-                # If query completed during cancellation, fetch and return results
-                if query_completed:
-                    LOG.info(f'Query completed during cancellation polling, returning results: job_id={job_id}')
-                    # Break out of the polling loop to fetch results below
-                    job_status = await self._qsclient.get_job_status(job_id)
-                    break
+        try:
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 3293d9c1

Pulled the `on_job_submitted` callback inside the same `try` that wraps the poll loop, so the existing `except asyncio.CancelledError` handler (which shields and awaits `_cancel_job_with_timeout`) now also covers cancellation during the callback. Previously, if the task was cancelled while awaiting `on_job_submitted` — after `submit_job` had already registered the job — the `CancelledError` re-raised straight past the handler and the QS job was left running. Now the backend cancel fires in that window too. Added regression test `test_execute_query_cancels_backend_when_cancelled_during_callback` asserting `cancel_job` is issued for the submitted job (and no results fetch happens) when the callback is cancelled.

src/keboola_mcp_server/tools/sql.py
```diff
@@ -175,12 +241,74 @@ async def query_data(
...
+            LOG.warning(
+                f'HTTP disconnect watcher for query_data "{query_name}" failed; cancelling underlying query',
+                exc_info=disconnect_exc,
+            )
+            cancel_reason = f'HTTP disconnect watcher failed during query_data "{query_name}"'
+        else:
+            LOG.info(f'HTTP client disconnected during query_data "{query_name}"; cancelling underlying query')
+            cancel_reason = f'HTTP client disconnected during query_data "{query_name}"'
+        await _cancel_and_drain(query_task)
+        raise asyncio.CancelledError(cancel_reason)
```
AdamVyborny: 🚫 **Blocking** — raising `asyncio.CancelledError` here has two real consequences:

1. **Cancelled queries get logged as `success`.** `@tool_errors` only catches `except Exception`; `CancelledError` is a `BaseException`, so it skips the error branch and `exception` stays `None`. The decorator's `finally` then runs `_trigger_event(..., exception=None)` → emits a `keboola.mcp-server-tool` **success** event. And because this raise is synthetic (the task wasn't actually cancelled), that `await` completes normally rather than being interrupted — so the success event really does fire. Since STOP-during-query is the main path this feature targets, success metrics get systematically polluted.
2. **Watcher-failed case hangs a connected client.** In the `disconnect_exc is not None` branch just above, the client has *not* disconnected, yet we still raise `CancelledError` → it lands in the MCP server's `except cancelled: if message.cancelled: … else: raise` path (reserved for transport-close TG teardown), which skips `message.respond()`. The connected client gets no error result.

Suggest `raise ValueError('Query was cancelled')` (consistent with the existing cancel path at the bottom of `query_data`): flows through `tool_errors` cleanly, logs an error event, and returns a proper client response. Same for the watcher-failed message.

src/keboola_mcp_server/tools/sql.py
```diff
@@ -175,12 +241,74 @@ async def query_data(
...
+            # raised error so a watcher bug isn't disguised as a client disconnect.
+            LOG.warning(
+                f'HTTP disconnect watcher for query_data "{query_name}" failed; cancelling underlying query',
+                exc_info=disconnect_exc,
+            )
+            cancel_reason = f'HTTP disconnect watcher failed during query_data "{query_name}"'
+        else:
+            LOG.info(f'HTTP client disconnected during query_data "{query_name}"; cancelling underlying query')
+            cancel_reason = f'HTTP client disconnected during query_data "{query_name}"'
+        await _cancel_and_drain(query_task)
```
AdamVyborny: Non-blocking (symmetry): the query-won path shields its drain — `await asyncio.shield(_cancel_and_drain(disconnect_task))` — precisely so `_cancel_and_drain`'s `except CancelledError: pass` can't swallow a real caller cancellation. This disconnect-won drain isn't shielded, so if `query_data` is cancelled during this (up-to-30s) drain, the genuine cancellation gets masked by the synthetic one below. Cheap to make symmetric: `await asyncio.shield(_cancel_and_drain(query_task))`. Not a blocker.

src/keboola_mcp_server/workspace.py
```diff
@@ -232,55 +232,74 @@ async def execute_query(
...
+                        )
+        except asyncio.CancelledError:
+            # Client (e.g. MCP `notifications/cancelled`) cancelled the in-flight tool call.
+            # Propagate the cancel to the backend so the query doesn't keep scanning data.
+            # `asyncio.shield` keeps the cancel HTTP call alive even though our own task
+            # is being cancelled; without it the request would be torn down immediately.
+            LOG.info(f'Query cancelled by client: job_id={job_id}')
             try:
-                await on_job_submitted(info)
+                await asyncio.shield(self._cancel_job_with_timeout(job_id, reason='Client cancelled the request'))
```
AdamVyborny, line 300: Non-blocking (awareness): `asyncio.shield` protects against the *first* cancellation only. If a second `.cancel()` lands while `_cancel_job_with_timeout` is mid-poll (up to `_CANCELLATION_TIMEOUT` = 30s), this `await shield(...)` is cancelled, the `except CancelledError: pass` fires, and `_cancel_job_with_timeout` is left running detached — it may not finish if the loop tears down, leaving the QS job scanning until the 5-min `_QUERY_TIMEOUT` backstop kills it. Acceptable given the backstop; flagging only for awareness.

src/keboola_mcp_server/tools/sql.py
```diff
@@ -22,6 +25,69 @@
...
+
+    Any error from `is_disconnected()` is treated as "still connected" so a transient
+    ASGI hiccup never cancels an otherwise-working query.
+
+    `poll_interval` is resolved at call time (not bound as a default) so tests can
+    patch `_DISCONNECT_POLL_INTERVAL` and have it take effect here.
+    """
+    interval = poll_interval if poll_interval is not None else _DISCONNECT_POLL_INTERVAL
+    try:
+        request = get_http_request()
```
AdamVyborny: Non-blocking (reuse): `mcp.py` already has `get_http_request_or_none()` wrapping `get_http_request()` + the `RuntimeError` handling (used by `errors.py`). Could reuse it here instead of re-importing `get_http_request` and re-catching `RuntimeError`, so the no-HTTP-context detection lives in one place.

AdamVyborny (changes requested):
Reviewed the disconnect-cancellation change. The direction is right and the 5-min `_QUERY_TIMEOUT` backstop keeps it safe from runaway queries regardless. Two blockers before merge (inline), plus non-blocking cleanups below. The stateless-HTTP design is a deliberate team decision — not relitigating that here.

## 🚫 Blocking

**1. Don't signal disconnect with `asyncio.CancelledError`** (inline on `sql.py:301`). It's a `BaseException`, so it bypasses `@tool_errors` (`except Exception`): the decorator's `finally` then logs the cancelled query as a **`success`** Storage event — and STOP-during-query is the *primary* path this feature targets, so success telemetry is systematically wrong. In the watcher-*failed* branch it also leaves a still-connected client with no error response. Raise a normal `ValueError('Query was cancelled')` instead (matches the existing cancel path at the end of `query_data`).

**2. Verify the watcher actually fires in stateless-SSE before merge.** The earlier branch "never fired"; the Kai-proxy fix (keboola/ui#6349) addresses one cause (socket held open) but not the in-process one: in SSE mode `sse_starlette._listen_for_disconnect()` is permanently parked on the *same* single-consumer ASGI `receive()` that the watcher polls via `is_disconnected()`. It may still work (uvicorn re-delivers `http.disconnect` on each poll), but given the history + zero integration coverage, please confirm with a canary test on `canary-orion` together with the Kai proxy change, and ideally add a prod log/metric when the watcher fires. Not a runaway-query risk (the 5-min timeout covers that) — but if this link is silently broken the feature buys ~nothing over the existing timeout.

## Non-blocking (nice-to-have, not blockers)

- **`sql.py:300`** — shield the disconnect-won drain for symmetry with the query-won path (inline).
- **`workspace.py:297`** — double-cancel can orphan the shielded backend cancel; bounded by the 5-min backstop, so just for awareness (inline).
- **`sql.py` `_watch_for_http_disconnect`** — could reuse the existing `get_http_request_or_none()` helper in `mcp.py` instead of re-catching `RuntimeError` (inline).
- **`workspace.py:266`** — the timeout-branch `reason` local now shares scope with the new `except CancelledError` block (which hardcodes its own reason); a future edit referencing `reason` there would hit `UnboundLocalError` on the pre-timeout cancel path. Minor fragility.
- The three drain sites (`except BaseException`, disconnect-won, query-won) repeat the cancel+shield idiom — a single `try/finally` draining both tasks would be easier to keep correct over time. (TaskGroup isn't an option — project supports Python 3.10.)

src/keboola_mcp_server/tools/sql.py
Matovidlo: **Fixed** ✅ in commit e97ab0cf424f727334f282d8cc2be0cdb4dfdcb7

Fixed. The disconnect/watcher-failed paths now `raise ValueError('Query was cancelled')` instead of `asyncio.CancelledError`, matching the existing cancel path at the bottom of `query_data`. Confirmed your analysis against `errors.py`: `@tool_errors` only catches `Exception`, so the old `CancelledError` (a `BaseException`) skipped the error branch and `_trigger_event` fired a `success` event — and left a connected client with no response in the watcher-failed branch. Both gone now: the cancelled call flows through `tool_errors` as an error and returns a proper response. The honest client-disconnect vs watcher-failed distinction is preserved in the logs (INFO vs WARNING+exc_info). Tests `test_query_data_cancels_on_http_disconnect` and `test_query_data_logs_when_disconnect_watcher_raises` updated to assert `ValueError`.

src/keboola_mcp_server/tools/sql.py
Matovidlo: **Fixed** ✅ in commit e97ab0cf424f727334f282d8cc2be0cdb4dfdcb7

Fixed — and made uniform. All three drain sites now go through a single `_drain(*tasks)` helper that does `await asyncio.shield(asyncio.gather(_cancel_and_drain(t) for t in tasks))`. The shield protects the in-flight cleanup (backend `cancel_job`), but the `await` itself is deliberately NOT suppressed, so a genuine caller-cancellation at any of these boundaries propagates out rather than being masked by the drain. That covers the disconnect-won drain you flagged here, the query-won drain, and the `except BaseException` path (which keeps its own `contextlib.suppress` so the original exception re-raises).

src/keboola_mcp_server/tools/sql.py
Matovidlo: **Fixed** ✅ in commit e97ab0cf424f727334f282d8cc2be0cdb4dfdcb7

Fixed. `query_data` now resolves the request once via `get_http_request_or_none()` from `mcp.py` — the no-HTTP-context detection lives in one place again. This also let me gate the whole disconnect race on HTTP presence: with no request bound (stdio / background workers) the query runs directly, so the watcher's `asyncio.Event().wait()` block-forever branch and its `RuntimeError` handling are deleted. `_watch_for_http_disconnect(request, ...)` now just takes the resolved request.

src/keboola_mcp_server/workspace.py
Matovidlo: **Intentional** ℹ️

Leaving as-is, per your own framing — the double-`.cancel()` window can detach the shielded `_cancel_job_with_timeout`, but it's bounded by the 5-min `_QUERY_TIMEOUT` backstop, so the QS job can't scan indefinitely. Capturing it here for awareness rather than adding a second shield layer that would buy little over the backstop. Happy to revisit if we ever see it in practice.

src/keboola_mcp_server/tools/sql.py
```diff
@@ -13,15 +15,129 @@
...
+        return query_task.result()
+
+    # The watcher won: either the client disconnected, or the watcher itself raised. Cancel the
+    # query either way, but keep the log honest so a watcher bug isn't disguised as a disconnect.
+    watcher_exc = disconnect_task.exception()
+    if watcher_exc is not None:
+        LOG.warning(
+            f'HTTP disconnect watcher for query_data "{query_name}" failed; cancelling underlying query',
+            exc_info=watcher_exc,
+        )
```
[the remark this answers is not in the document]
Matovidlo: Good catch — fixed in 293e7fe7. Now passes an explicit `(type(watcher_exc), watcher_exc, watcher_exc.__traceback__)` tuple so the watcher's traceback is logged instead of falling back to an empty `sys.exc_info()`.

AdamVyborny (approved):
LGTM

src/keboola_mcp_server/tools/sql.py
```diff
@@ -13,15 +15,132 @@
...
+from keboola_mcp_server.workspace import JobSubmittedInfo, QueryResult, SqlSelectData, WorkspaceManager
 
 LOG = logging.getLogger(__name__)
 
 SQL_TOOLS_TAG = 'sql'
 MAX_ROWS = 1_000
 MAX_CHARS = 50_000
+# How often to check whether the HTTP client has disconnected during a long query.
+# Mirrors the 1 s job-poll cadence in `_Workspace.execute_query`.
+_DISCONNECT_POLL_INTERVAL = 1.0
```
AdamVyborny, line 31: Nitpick; I don't like mirroring of consts just based on reading comments, it will diverge...

Matovidlo: I will fix it later hopefully :sweat_smile:

PR #588, opened 2026-06-16
feat(search): adopt server-side global search with safe fallback [AI-3393]

AdamVyborny:
## Description

**Linear**: [AI-3393](https://linear.app/keboola/issue/AI-3393/adopt-server-side-storage-search-endpoints-in-the-mcp-search-tool) (resolves [AI-3236](https://linear.app/keboola/issue/AI-3236/search-is-very-slow))

Takes over #579 (rebased onto `main`) and adds the changes the AI-3236 benchmark showed are required before server-side global search can replace the client-side `search` enumeration.

### Change Type

### Summary

Continues #579: textual `search` uses the server-side `global-search` endpoint instead of client-side enumeration. The AI-3236 benchmark (~70× faster on a 103-bucket project) confirmed it is a large win but **not a strict 1:1 replacement**, so this PR adds the safety net:

- **Fallback to client-side enumeration on zero hits *or* error** — not only when the `global-search` feature is absent. The feature flag does not guarantee the project's index is populated (the backfill is asynchronous) and the endpoint can 5xx transiently; without this, an indexed-but-not-yet-backfilled project silently returns nothing.
- **Corrected the tool docstring**: global search is a tokenized full-text *name* search (case/diacritics-insensitive), **not** typo-corrected, and does not search IDs / columns / descriptions. (The previous "fuzzy / typo-tolerant" wording was disproven by the benchmark — `custommer` matched nothing.)
- Does **not** gate on `global-search-backend` — empirically an unreliable readiness signal.
- Keeps the legacy enumeration path until global search is GA (platform side DMD-4 is pre-GA; DMD-772 is an open production-miss bug).

Full method, speed/recall tables and landing notes are in `feature_spec/global_search_adoption/RFC.md` → "Benchmark results & landing plan (AI-3236)" and on the [AI-3236 issue](https://linear.app/keboola/issue/AI-3236/search-is-very-slow).

## Testing

Verified end-to-end against `canary-orion` projects by driving the real `search()` code:
- Project 198 (feature on, index empty) → now returns **15** table hits via the enumeration fallback (was **0**).
- Project 21 (indexed, 103 buckets) → uses the fast global-search path (~0.3s vs ~22s enumeration).

## Checklist

- [x] Self-review completed
- [x] Unit tests added/updated (fallback on zero-hits, fallback on error)
- [ ] Integration tests added/updated — deferred: the CI integ failure is a dirty shared CI project (`integtests/conftest.py:324`), not a search defect
- [x] Project version bumped (1.70.0, minor)
- [x] Documentation updated (`TOOLS.md` regenerated, RFC extended)

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -3485,8 +3485,8 @@ This tool supports two complementary search types:
 
 1) textual
 - Searches items by name, server-side (fast, independent of project size).
-- Full-text (fuzzy) matching: tokenized and typo/similarity tolerant — pass the name as the user said it;
-  do NOT "fix" spelling or build regex. It is NOT substring matching and does NOT support regex.
+- Tokenized full-text name matching, case- and diacritics-insensitive. Pass the plain name; do NOT build
+  regex (rejected). It is NOT typo-corrected — misspellings may not match.
 - Prefers the current branch context; when nothing is found there, automatically widens the search to all
   branches of the project — such hits carry `branch_id`/`branch_name` so you can tell where they live.
 
@@ -3515,9 +3515,9 @@ data-apps, flows, or transformations
 
 HOW IT WORKS:
 - Supports two types:
-  - search_type="textual": full-text (fuzzy) name search, server-side. Names only — descriptions, column names
-    and configuration contents are NOT searched (use config-based search for configuration contents). Matching
-    is tokenized and typo/similarity tolerant, so approximate names still match; it is not substring matching.
+  - search_type="textual": tokenized full-text name search, server-side. Names only — descriptions, column
+    names, IDs and configuration contents are NOT searched (use config-based search for configuration contents,
+    or get_tables for columns). Matching is case- and diacritics-insensitive but NOT typo-corrected.
   - search_type="config-based": matches inside configuration JSON objects, optionally narrowed by JSON path `scopes`
 - case-insensitive search
 - mode for pattern search: applies to config-based only — `literal` (default) or `regex`. Textual search ignores
@@ -3536,9 +3536,9 @@ IMPORTANT:
 - The search returns IDs that you can use with other tools (e.g., get_tables, get_configs, get_flows)
 - Results are ordered by the `updated` field, most recent first. `updated` is the item's last update time
   when available, or its creation time otherwise (textual/global-search hits expose only the creation time).
-- Textual search matches names only, with fuzzy full-text matching (typo/similarity tolerant; no regex). To find
-  items by description or by table column, use get_tables or config-based search; to find items by configuration
-  content, use config-based search.
+- Textual search matches names only, with tokenized full-text matching (case/diacritics-insensitive; not
+  typo-corrected; no regex). It may not return every item the legacy enumeration did. To find items by
+  description or by table column, use get_tables; to find items by configuration content, use config-based search.
 - For exact ID lookups, use specific tools like get_tables, get_configs, get_flows instead
 - Use specific `scopes` only when you know the config structure (schema or real example); otherwise run config-based
   search without scopes.
@@ -3656,7 +3656,7 @@ scopes=["storage"]
     },
     "mode": {
       "default": "literal",
-      "description": "How to interpret patterns. Applies to config-based search only: \"regex\" for regular expressions or \"literal\" for exact text (default: \"literal\"). Ignored by textual search, which is always a full-text (fuzzy) query and rejects \"regex\".",
+      "description": "How to interpret patterns. Applies to config-based search only: \"regex\" for regular expressions or \"literal\" for exact text (default: \"literal\"). Ignored by textual search, which is always a tokenized full-text name query (not typo-corrected) and rejects \"regex\".",
       "enum": [
         "regex",
         "literal"
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -3485,14 +3485,14 @@ This tool supports two complementary search types:
 
 1) textual
 - Searches items by name, server-side (fast, independent of project size).
-- Full-text (fuzzy) matching: tokenized and typo/similarity tolerant — pass the name as the user said it; do
-NOT "fix" spelling or build regex. It is NOT substring matching and does NOT support regex.
+- Full-text (fuzzy) matching: tokenized and typo/similarity tolerant — pass the name as the user said it;
+  do NOT "fix" spelling or build regex. It is NOT substring matching and does NOT support regex.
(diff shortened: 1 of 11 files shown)


src/keboola_mcp_server/tools/search.py
```diff
@@ -741,13 +947,60 @@ async def search(
...
+                'configurations.'
+            )
+        # The global-search feature flag does not guarantee the project's index is populated (the bulk
+        # backfill is asynchronous) and the endpoint can fail transiently, so global search is a fast path
+        # with a safety net: fall back to client-side enumeration on any error, or when it finds nothing.
+        try:
+            output = await _global_textual_search(client, spec, limit=limit, offset=offset)
+        except Exception as exc:
+            LOG.warning(f'Global search failed ({exc!r}); falling back to client-side enumeration.')
+            output = await _enumeration_search(client, spec, limit=limit, offset=offset)
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in 1f361f2c — the fallback warning now logs with `exc_info=True`, so the traceback is captured and transient 5xx can be told apart from mapping/parsing bugs in production.

TOOLS.md
```diff
@@ -3508,50 +3511,53 @@ data-apps, flows, or transformations
...
-  - search_type="textual": matches against id, name, displayName, and description, for tables also column names
-  and column descriptions
+  - search_type="textual": tokenized full-text name search, server-side. Names only — descriptions, column
+    names, IDs and configuration contents are NOT searched (use config-based search for configuration contents,
+    or get_tables for columns). Matching is case- and diacritics-insensitive but NOT typo-corrected.
   - search_type="config-based": matches inside configuration JSON objects, optionally narrowed by JSON path `scopes`
 - case-insensitive search
-- mode for pattern search: `literal` (default) or `regex`
+- mode for pattern search: applies to config-based only — `literal` (default) or `regex`. Textual search ignores
+  `mode` (always full-text) and rejects `regex`.
```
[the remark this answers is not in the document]
AdamVyborny: Fixed in 1f361f2c — the `patterns` field description now distinguishes textual (matches item names) from config-based (matches configuration JSON content); TOOLS.md regenerated so the schema block matches the narrative.

src/keboola_mcp_server/tools/search.py
```diff
@@ -741,13 +947,60 @@ async def search(
...
+    client = KeboolaClient.from_state(ctx.session.state)
+
+    if search_type == 'textual' and await client.storage_client.is_enabled(GLOBAL_SEARCH_FEATURE):
+        if mode == 'regex':
+            raise ToolError(
+                'Regex patterns are not supported for textual search — it is a tokenized full-text name search. '
+                'Pass the plain name as the pattern, or use search_type="config-based" for regex matching inside '
+                'configurations.'
+            )
+        # The global-search feature flag does not guarantee the project's index is populated (the bulk
```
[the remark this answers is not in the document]
AdamVyborny: By design: regex is intentionally honored on the legacy enumeration path (see `test_search_with_regex_pattern`) — only the global-search path can't express regex, so the rejection is scoped there. The legacy path is removed at GA, after which textual regex goes away entirely. Leaving behavior as-is to avoid regressing the enumeration path.

src/keboola_mcp_server/tools/search.py
```diff
@@ -603,20 +803,23 @@ async def search(
         ),
     ] = DEFAULT_GLOBAL_SEARCH_LIMIT,
     offset: Annotated[int, Field(description='Number of matching items to skip for pagination (default: 0).')] = 0,
-) -> list[SearchHit]:
+) -> SearchOutput:
```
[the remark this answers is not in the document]
AdamVyborny: Intentional output-shape change — the `SearchOutput` envelope (RFC §3) is a documented, minor-version-bumped change. The `integtests/tools/test_search.py` assertion update is part of the deferred integration-test work (the CI integ failure is the dirty shared project at `conftest.py:324`, unrelated to search).

src/keboola_mcp_server/clients/storage.py
```diff
@@ -932,6 +960,36 @@ async def table_metadata_update(
 
         return cast(JsonDict, await self.post(endpoint=f'tables/{table_id}/metadata', data=payload))
 
+    async def tables_search(
```
Matovidlo: This is dead code

src/keboola_mcp_server/clients/storage.py
```diff
@@ -614,26 +630,34 @@ async def configuration_list(self, component_id: str) -> list[JsonDict]:
     async def component_configurations_search(
         self,
         component_id: str | None = None,
+        configuration_id: str | None = None,
```
Matovidlo: This is unused in utils (not propagated)

src/keboola_mcp_server/tools/search.py
Matovidlo, line 1: This is becomming bloated, we should extract the global search parts somewhere in separate tool 🤔

Matovidlo (changes requested):

src/keboola_mcp_server/clients/storage.py
AdamVyborny: Removed in b6188bb4 — `tables_search` had no caller (added as groundwork only), so dropped the method and its tests.

src/keboola_mcp_server/clients/storage.py
AdamVyborny: Fixed in b6188bb4 — dropped the unused `configuration_id` and `include` params (not propagated by any caller). The only caller (data-app folder lookup) uses `component_id` + `metadata_keys`, which stay server-side.

src/keboola_mcp_server/tools/search.py
AdamVyborny: Done in 62b1a02d — extracted the global-search path into `search_global.py` and the shared models/aliases/constants into `search_models.py`. `search.py` drops 1094 → ~590 lines (tool entry point + legacy enumeration path), re-exporting the public names via `__all__` so importers (server.py, generate_tool_docs, tools/storage/usage.py, tests) are unaffected. The split is models ← global ← tool, so no import cycle. Pure refactor — full unit suite green and the real search() verified unchanged against canary-orion (not-indexed project still falls back, indexed project uses the fast path).

Kept it as one internal `search` tool (not a separate MCP tool) to preserve the 1:1-replacement contract; shout if you'd prefer a different module boundary.

Matovidlo (approved):
LGTM

AdamVyborny:
@Matovidlo can you help me with integ test failing pls

PR #591, opened 2026-06-17
AJDA-2852: fix: don't report modify_streamlit_data_app failure after a committed write

sykora-ji:
[link to issue](https://linear.app/keboola/issue/AJDA-2852/fix-mcp-for-streamlit-and-cc)

## Description

Fixes point 2 of AJDA-2852 (from a customer support request): `modify_streamlit_data_app` could silently apply changes while still reporting an error. The tool performed the API write (`configuration_update` on the update path, `create_data_app` on the create path) and then built its response (re-fetch, metadata, folder, links, model validation). Any failure in a post-write step was wrapped by the `@tool_errors` decorator into a `ToolError`, so the tool reported **failure** even though the write had already committed. Agents then retried and re-applied the change — observed as a config version jumping v27 → v29 and deploying broken code (wrong packages), with no way to tell from the error whether the write had landed.

Changes in `src/keboola_mcp_server/tools/data_apps.py`:

- **Hard split between the committing write and response building.** In both branches the write (`configuration_update` / `create_data_app`) is now the only step allowed to fail the whole tool — if it raises, nothing was committed and the `ToolError` is truthful. Everything after it is wrapped in a best-effort `try/except`.
- **Truthful partial-success on post-write failure.** New helpers `_partial_update_output` / `_partial_create_output` return a `ModifiedDataAppOutput` whose `change_summary` states the change WAS applied (with the new version) and that the agent must **NOT** retry, pointing it to `deploy_data_app`. This breaks the retry/double-write loop.
- **Version known without the re-fetch.** The new config version is read directly from the `configuration_update` response (`update_resp.get('version')`), so it is reported even when the re-fetch fails. The partial-update response mirrors the success-path "redeploy required" wording based on the pre-update app state.
- **Metadata stamp reordered.** On the update path, `set_cfg_update_metadata` now runs *before* `_fetch_data_app` (previously after). This is intentional — the metadata version now comes from the write response rather than the re-fetch — and harmless, but flagged here for reviewers as the ordering changed.
- **Helpers never raise.** Both fall back to a `DataAppSummary` built from primitives (`SafeState`/`SafeType` accept any string), so a committed write can never surface as a `ToolError` even if `DataAppSummary` construction fails.

Review hardening (follow-up commit):

- The new-version read is now defensive (`isinstance(update_resp, dict)`), so a non-dict response can't raise after the committing write and reintroduce the bug.
- `set_cfg_update_metadata` is skipped when the update response carries no numeric version, instead of stamping the stale pre-update version (which would record a misleading `UPDATED_BY_MCP` version). Covered by a new regression test.
- The raw exception is no longer interpolated into the agent-facing `change_summary` (it is already logged at the call site), keeping the message stable and free of internal detail.

Tests (`tests/tools/test_data_apps.py`):

- `test_modify_streamlit_data_app_partial_success_when_response_building_fails` — parametrized over update/create and over which post-write step fails; asserts no `ToolError`, success response with state-correct wording, the new version surfaced, "do not retry" guidance, and that the committing write ran exactly once.
- `test_partial_output_helpers_never_raise_when_summary_construction_fails` — forces the primary `DataAppSummary` construction to raise and asserts the helpers still return a sane fallback.
- `test_modify_streamlit_data_app_update_skips_metadata_when_version_missing` — asserts `set_cfg_update_metadata` is skipped when the update response carries no numeric version.

Patch version bump `1.72.1` → `1.72.2`.

## Release Notes

- **Justification**
    - A committed data-app change must never be reported as a failure. The old behavior made agents believe the write had failed, so they retried and applied the change twice — bumping the config version and deploying broken code with no signal that anything had landed.
    - For a non-developer: when the assistant updates or creates a Streamlit data app, it now always tells you truthfully whether the change was saved — even if it couldn't load all the follow-up details — instead of showing a confusing error that hides the fact the change already happened.
- **Plans for Customer Communication**
    - No customer communication needed. This is a reliability fix to an MCP tool's error reporting; no API surface, parameters, or success behavior change.
- **Impact Analysis**
    - Scope is limited to `modify_streamlit_data_app`. The success path is unchanged. The only behavioral change is on post-write failure: the tool now returns a partial-success result instead of raising. No effect on other tools; not single-tenant-specific.
    - Not behind a feature flag.
- **Deployment Plan**
    - Continuous deployment with the regular MCP server release; no stack-by-stack rollout required.
- **Rollback Plan**
    - Fully reversible by reverting this commit and redeploying. Not a one-way-door change.
- **Post-Release Support Plan**
    - No special monitoring required. If desired, the `change_summary` "partial success" wording can be watched in tool events to confirm the degraded path behaves as expected. Support team notification not needed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.72.1"
+version = "1.72.2"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/src/keboola_mcp_server/tools/data_apps.py
+++ b/src/keboola_mcp_server/tools/data_apps.py
@@ -551,8 +551,14 @@ async def modify_streamlit_data_app(
     )
 
     if configuration_id:
-        # Update existing data app
-        data_app, updated_config, _ = await modify_streamlit_data_app_internal(
+        # Update existing data app.
+        #
+        # `configuration_update` below is the ONLY step allowed to fail the whole tool: if it raises, nothing
+        # was committed and a ToolError truthfully reports the failure. Everything after it merely enriches the
+        # response (re-fetch, metadata, folder, links) and must NEVER turn a committed write into a reported
+        # failure -- otherwise the agent, seeing an error, retries and re-applies the change, producing the
+        # v27->v29 double-write that deployed broken code (AJDA-2852). Hence the best-effort block below.
+        data_app_pre, updated_config, _ = await modify_streamlit_data_app_internal(
             client=client,
             workspace_manager=workspace_manager,
             name=name,
@@ -563,43 +569,64 @@ async def modify_streamlit_data_app(
             configuration_id=configuration_id,
             change_description=change_description,
         )
-        await client.storage_client.configuration_update(
+        update_resp = await client.storage_client.configuration_update(
             component_id=DATA_APP_COMPONENT_ID,
             configuration_id=configuration_id,
             configuration=updated_config,
             change_description=change_description or 'Change Data App',
-            updated_name=name or data_app.name,
-            updated_description=description or data_app.description,
+            updated_name=name or data_app_pre.name,
+            updated_description=description or data_app_pre.description,
         )
-        data_app = await _fetch_data_app(client, configuration_id=configuration_id, data_app_id=None)
-        await set_cfg_update_metadata(
-            client=client,
-            component_id=DATA_APP_COMPONENT_ID,
-            configuration_id=configuration_id,
-            configuration_version=int(data_app.config_version),
-        )
-        folder_hint = await apply_folder_metadata(
-            client, DATA_APP_COMPONENT_ID, configuration_id, folder, 'data apps', 'modify_streamlit_data_app'
-        )
-        links = links_manager.get_data_app_links(
-            configuration_id=data_app.configuration_id,
-            configuration_name=name,
-            deployment_link=data_app.deployment_url,
-            uses_basic_authentication=_uses_basic_authentication(data_app.configuration.get('authorization') or {}),
-        )
-        response = (
-            'updated (redeploy required to apply changes in the running app)'
-            if data_app.state in ('running', 'starting')
-            else 'updated'
-        )
-        return ModifiedDataAppOutput(
-            response=response,
-            change_summary=folder_hint,
-            data_app=DataAppSummary.model_validate(data_app.model_dump()),
-            links=links,
-        )
+        # --- write committed past this point; response building is strictly best-effort ---
+        # The new version comes straight from the PUT response, so it is known even if the re-fetch below fails.
+        new_version = str(update_resp.get('version') or '')
+        try:
+            await set_cfg_update_metadata(
+                client=client,
+                component_id=DATA_APP_COMPONENT_ID,
+                configuration_id=configuration_id,
+                configuration_version=int(new_version) if new_version.isdigit() else int(data_app_pre.config_version),
+            )
+            folder_hint = await apply_folder_metadata(
+                client, DATA_APP_COMPONENT_ID, configuration_id, folder, 'data apps', 'modify_streamlit_data_app'
+            )
+            data_app = await _fetch_data_app(client, configuration_id=configuration_id, data_app_id=None)
+            links = links_manager.get_data_app_links(
+                configuration_id=data_app.configuration_id,
+                configuration_name=name,
+                deployment_link=data_app.deployment_url,
+                uses_basic_authentication=_uses_basic_authentication(data_app.configuration.get('authorization') or {}),
+            )
+            response = (
+                'updated (redeploy required to apply changes in the running app)'
+                if data_app.state in ('running', 'starting')
+                else 'updated'
+            )
+            return ModifiedDataAppOutput(
+                response=response,
+                change_summary=folder_hint,
+                data_app=DataAppSummary.model_validate(data_app.model_dump()),
+                links=links,
+            )
+        except Exception as exc:
+            LOG.exception(
+                'Data app configuration %s was updated (version %s) but building the response failed; '
+                'returning a partial success so the change is not retried.',
+                configuration_id,
+                new_version or '?',
+            )
+            return _partial_update_output(
+                data_app_pre=data_app_pre,
+                links_manager=links_manager,
+                configuration_id=configuration_id,
+                name=name,
+                new_version=new_version,
+                error=exc,
+            )
     else:
-        # Create new data app
+        # Create new data app. As with the update path, `create_data_app` is the committing write; the
+        # metadata/folder/links steps after it are best-effort and must not fail the tool once the app exists,
+        # so a failed retry cannot create a duplicate app (AJDA-2852).
         config = _build_data_app_config(name, source_code, packages, authentication_type, secrets, sql_dialect)
         config = await client.encryption_client.encrypt(
             config, component_id=DATA_APP_COMPONENT_ID, project_id=project_id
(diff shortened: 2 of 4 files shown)


src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -502,43 +508,64 @@ async def modify_streamlit_data_app(
...
-            response=response,
-            change_summary=folder_hint,
-            data_app=DataAppSummary.model_validate(data_app.model_dump()),
-            links=links,
+            updated_name=name or data_app_pre.name,
+            updated_description=description or data_app_pre.description,
         )
+        # --- write committed past this point; response building is strictly best-effort ---
+        # The new version comes straight from the PUT response, so it is known even if the re-fetch below fails.
+        new_version = str(update_resp.get('version') or '')
```
MiroCillik: This line sits *past* the "write committed past this point" marker but *outside* the best-effort `try`. If `update_resp` were ever not a dict, `.get` would raise after the write has already committed — reintroducing exactly the bug this PR fixes (committed write reported as a `ToolError`).

In practice `configuration_update` is typed `-> JsonDict`, so this is safe today. But conceptually everything after the committing write should be inside the best-effort block. Suggest moving this into the `try` and falling back to `data_app_pre.config_version` on the rare parse failure.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -502,43 +508,64 @@ async def modify_streamlit_data_app(
...
+        try:
+            await set_cfg_update_metadata(
```
MiroCillik: Ordering change worth flagging: `set_cfg_update_metadata` now runs *before* `_fetch_data_app` (previously after). It's harmless — arguably more correct, since the metadata version now comes from the write response rather than the re-fetch — but it's an unflagged behavioral reorder. A one-line mention in the PR description would help future readers.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -639,6 +682,115 @@ async def modify_streamlit_data_app_internal(
...
+) -> ModifiedDataAppOutput:
+    """Build a truthful partial-success response after a committed Streamlit update whose response building failed.
+
+    The configuration write already landed, so this MUST NOT raise -- it is built entirely from the pre-update
+    data app (fetched before the write) plus the new version from the update response. The ``change_summary``
+    tells the agent the change IS applied and must not be retried, preventing the duplicate-write loop.
+    """
+    try:
+        summary = DataAppSummary.model_validate(data_app_pre.model_dump())
+        summary.config_version = new_version or summary.config_version
```
MiroCillik, line 768: Expected, just noting for consumers: on the partial path the `DataAppSummary` is the *pre-update* app with only `config_version` patched to the new version — so other fields (`configuration`, folder, `state`) reflect pre-write state. This is the right trade-off (the re-fetch is exactly what failed) and `change_summary` is honest about it. No change needed; callers should just rely on `config_version`/`change_summary` rather than the other summary fields on this path.

MiroCillik (commented):
Inline review notes from a manual review of the fix. The core approach (hard split between the committing write and best-effort enrichment, returning a truthful partial-success) is sound and well-documented; these are minor hardening/cleanup points, none blocking.

Also a non-inline nit: the PR description says the bump is `1.69.5` → `1.69.6`, but the diff bumps `1.70.0` → `1.70.1` (correct against `origin/main`). Likely a post-rebase artifact — worth updating the body so it isn't misleading.

MiroCillik (approved):
Just minor issues in the comments. Approving.

src/keboola_mcp_server/tools/data_apps.py
sykora-ji: Good catch — fixed in 8f6e778e. `new_version` is now read defensively (`str(update_resp.get('version') or '') if isinstance(update_resp, dict) else ''`), so a non-dict response can't raise after the committing write and reintroduce the bug. It stays just outside the `try` because the `except` handler needs it for the partial response, but it can no longer throw.

src/keboola_mcp_server/tools/data_apps.py
sykora-ji: Agreed — this reorder is intentional (the metadata version now comes from the write response rather than the re-fetch). Flagged it explicitly in the PR description so future readers see the ordering changed.

src/keboola_mcp_server/tools/data_apps.py
sykora-ji: Correct, that's the intended trade-off — the re-fetch is exactly what failed, so on the partial path only `config_version` is patched and `change_summary` is explicit about it. No change.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -502,43 +508,64 @@ async def modify_streamlit_data_app(
...
+        # --- write committed past this point; response building is strictly best-effort ---
+        # The new version comes straight from the PUT response, so it is known even if the re-fetch below fails.
+        new_version = str(update_resp.get('version') or '')
+        try:
+            await set_cfg_update_metadata(
+                client=client,
+                component_id=DATA_APP_COMPONENT_ID,
+                configuration_id=configuration_id,
+                configuration_version=int(new_version) if new_version.isdigit() else int(data_app_pre.config_version),
+            )
```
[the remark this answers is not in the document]
sykora-ji: Fixed in 8f6e778e. `set_cfg_update_metadata` is now skipped entirely when the update response carries no numeric version, instead of falling back to the (stale) pre-update version — so we never stamp a misleading UPDATED_BY_MCP version. Added a regression test (`test_modify_streamlit_data_app_update_skips_metadata_when_version_missing`).

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -639,6 +682,115 @@ async def modify_streamlit_data_app_internal(
...
+        if data_app_pre.state in ('running', 'starting')
+        else 'updated'
+    )
+    return ModifiedDataAppOutput(
+        response=response,
+        change_summary=(
+            f'The configuration WAS updated (version {new_version or "unknown"}), but loading the full app '
+            f'details failed, so this response is partial: {error}. Do NOT retry the update -- the change is '
+            f'already applied. Call deploy_data_app to apply it to the running app.'
+        ),
```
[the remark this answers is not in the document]
sykora-ji: Fixed in 8f6e778e — the raw exception is no longer interpolated into `change_summary`. It's already logged via `LOG.exception` at the call site, so the agent-facing message is now stable and free of internal detail.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -639,6 +682,115 @@ async def modify_streamlit_data_app_internal(
...
+        )
+    except Exception:
+        links = []
+    return ModifiedDataAppOutput(
+        response='created',
+        change_summary=(
+            f'The data app WAS created (configuration {data_app_resp.config_id}), but building the full response '
+            f'failed, so this response is partial: {error}. Do NOT retry creation -- it would create a duplicate. '
+            f'Call deploy_data_app to start the app.'
+        ),
```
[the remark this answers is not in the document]
sykora-ji: Same fix as the update helper — dropped the raw exception from `change_summary` in 8f6e778e (kept the stable 'WAS created … do NOT retry' guidance; the exception is logged at the call site).

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.70.0"
+version = "1.70.1"
```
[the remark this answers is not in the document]
sykora-ji: Thanks — the actual bump is `1.70.0` → `1.70.1` (correct against `origin/main`); the `1.69.x` in the description was a stale post-rebase artifact. Updated the PR description to match.

sykora-ji:
@keboola-pr-reviewer-bot review

sykora-ji:
@keboola-pr-reviewer-bot review

PR #593, opened 2026-06-19
fix(storage): prefer KBC metadata over legacy description for buckets/tables [AI-3423]

Matovidlo:
## Description

**Linear**: [AI-3423](https://linear.app/keboola/issue/AI-3423/get-buckets-returns-legacy-bucket-description-instead-of)

### Change Type

### Summary

`get_buckets` (and `get_tables`) returned the stale legacy top-level `description`
field — auto-generated text like `"Bucket created by Transformation API"` — instead
of the curated `KBC.description` / `KBC.sharedDescription` value stored in bucket/table
metadata. For linked and shared buckets the legacy field is always non-empty, so the
priority logic in `set_description` never reached the metadata fallback.

`update_descriptions` writes the user's description to `KBC.description` **metadata**
(not the legacy field), so metadata is the source of truth. This PR inverts the
priority in both `BucketDetail.set_description` and `TableDetail.set_description`:

- Buckets: `KBC.sharedDescription` → `KBC.description` → legacy `description`
- Tables: own `KBC.description` → `sourceTable` `KBC.description` → legacy `description`

The legacy field is now only used as a last resort when no KBC metadata is present.

## Testing

- [ ] Tested with Cursor AI desktop (\`Streamable-HTTP\` transports)

Unit tests: updated the parametrized description-resolution suites for both
`BucketDetail` and `TableDetail` to assert metadata wins over the stale legacy field
(the exact AI-3423 scenario), plus a case confirming the legacy field is still used
when no metadata exists. All 57 storage tests pass; black + flake8 clean.

## Checklist

- [x] Self-review completed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.72.2"
+version = "1.72.3"
 description = "MCP server for interacting with Keboola Connection"
 readme = "README.md"
 requires-python = ">=3.10"
--- a/src/keboola_mcp_server/tools/storage/tools.py
+++ b/src/keboola_mcp_server/tools/storage/tools.py
@@ -213,12 +213,15 @@ class BucketDetail(BaseModel):
     @model_validator(mode='before')
     @classmethod
     def set_description(cls, values: dict[str, Any]) -> dict[str, Any]:
-        description = values.get('description')
+        # KBC metadata holds the curated, user-editable description (update_descriptions writes here).
+        # The legacy top-level `description` field is auto-generated (e.g. "Bucket created by
+        # Transformation API") and stale for linked/shared buckets, so metadata must take precedence.
         metadata = values.get('metadata', [])
-        if not description:
-            description = get_metadata_property(metadata, MetadataField.SHARED_DESCRIPTION)
-        if not description:
-            description = get_metadata_property(metadata, MetadataField.DESCRIPTION)
+        description = (
+            get_metadata_property(metadata, MetadataField.SHARED_DESCRIPTION)
+            or get_metadata_property(metadata, MetadataField.DESCRIPTION)
+            or values.get('description')
+        )
         values['description'] = description or None
         return values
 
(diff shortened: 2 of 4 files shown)


Matovidlo:
@claude review

tomasfejfar (approved):
## ✅ Approving — verified live against the PR build

Tested end-to-end on Canary (project 499) with the MCP server running this branch's code (local `:8000`, `appVersion: DEV`), driving it through the Kai agent in kbc-ui so the call path matches real usage. Linked bucket `in.c-linked-pick-sample-data` (source: catalog 431 / `in.c-pick-sample-data`).

### Verification result

| Scenario | Surface | Description returned | Resolves to | Result |
| -- | -- | -- | -- | -- |
| Linked bucket description | **`get_buckets` (this PR)** | `"Description set when sharing the bucket"` | `KBC.sharedDescription` | ✅ curated value, not the stale legacy field |
| Same bucket | UI – Data Catalog (Shared with you) | `"Description set when sharing the bucket"` | `KBC.sharedDescription` | ✅ MCP matches |
| Same bucket | UI – Storage bucket preview (consumer) | `"Desciption on source project"` | `KBC.description` (inherited from source) | ⚠️ MCP **differs** from this view |

The fix does what AI-3423 intended: `get_buckets` now returns the curated `KBC.sharedDescription` / `KBC.description` instead of the auto-generated legacy `description`. Precedence `sharedDescription → description → legacy` confirmed active.

### ⚠️ Note: MCP intentionally diverges from one of the UI views

For a linked/shared bucket the platform shows **two different "Bucket Description" values** depending on the surface:

- **Data Catalog (shared-with-you)** shows the **sharing-time** description — this PR's `get_buckets` agrees with it. 👍
- **Bucket preview** of the linked bucket shows the **source project's** description — this PR's `get_buckets` does **not** match it.

So an agent answering "what's this bucket's description?" will report the sharing-time text, which can differ from what a user sees on the bucket's own Storage/preview page. That's a defensible choice (the sharing-time description is the producer's consumer-facing text), but it's a deliberate mismatch worth knowing about.

This UI-side divergence — and the open question of a single, consistent resolution order across views (incl. whether a linked-project-local description should be allowed to override) — is tracked in **[UT-4341](https://linear.app/keboola/issue/UT-4341/inquiry-which-bucket-description-should-win-for-sharedlinked-buckets)**. Relevant to that discussion: the precedence in this PR has **no rung above `sharedDescription`**, so a consumer-local override could never win today — and in practice the linked-bucket Storage page exposes no editor to set one.

None of the above blocks this PR; the fix is correct for its stated scope. Approving.

PR #620, opened 2026-07-10
CFTL-714: allow changing a python-js data app's git branch via MCP

soustruh:
## Description

**Linear**: CFTL-714

### Change Type

### Summary

**The problem**

Changing the git branch of a python-js data app had no supported MCP path:

- `update_config` refuses `keboola.data-apps` and tells you to use the data-app tools — by design.
- But the data-app tool couldn't do it either: `modify_python_js_data_app` rejected `branch` on update, and its update path never wrote `parameters.dataApp.git`.
- So the only way to point a data app at a different branch was a raw Storage API config PUT (or the UI).

**The fix**

- `modify_python_js_data_app` now accepts `branch` on the update path (when you pass `configuration_id`). For an external-git app it writes `parameters.dataApp.git.branch`.
- Only the branch changes. The encrypted `#password`, `repository` and `username` are left exactly as stored — the update re-PUTs the config and the encryption service leaves an already-encrypted value alone, so the token is never touched or re-encrypted.
- The repoint is allowed only for **external-git** apps — a draft, or an app bound to an external repository. An app on a **Keboola-managed** git repo is rejected with a clear message before anything is written: its branch is owned by the platform and this tool must not change that. The check uses the authoritative `is_managed_git_repo` flag from the data-science `.../git-repo` endpoint (not the presence of a git block — managed apps can carry one too).
- On update any branch (including `main`) is allowed; creating a new draft still reserves `main` for the prod app.

Redeploy the app afterwards to actually serve the new branch. (Whether a redeploy reliably picks up the change is a separate issue, CFTL-716.)

## Testing

- Unit (`tests/tools/test_data_apps.py`): the update path writes only `git.branch` and preserves the encrypted credential, and rejects blank/whitespace branch names; the tool repoints an external-git app (the PATCH carries the new branch, encryption is never called) and rejects the repoint on a Keboola-managed app (via `is_managed_git_repo`, even when it carries a git block) and when the flag can't be determined — without writing anything.
- Integration (`integtests/tools/test_data_apps.py`): the prod + external-git draft lifecycle now repoints the draft's branch through the update path and checks the stored config reflects it with the credential preserved. Needs live `INTEGTEST_*` credentials — not run locally.
- No manual MCP-client / canary-orion E2E run.

## Checklist

- [x] Self-review completed
- [x] Integration tests added/updated
- [x] Project version bumped according to the change type (patch)
- [x] Documentation updated (TOOLS.md, docstrings)

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -2255,10 +2255,10 @@ draft handle.
 - `slug` is required on create and immutable after.
 - The **update path** (passing `configuration_id`) is for changing `name`, `description`,
   `authentication_type`, `auto_suspend_after_seconds`, `storage` on either a prod app or
-  a draft, and for repointing an **external-git** app's `branch` (drafts, or apps bound to
-  an external repository — a managed-repo prod app has no editable branch). Source code
-  changes go through the git flow above, not this tool. After a `branch` repoint, call
-  `deploy_data_app` to serve the new branch.
+  a draft, and for repointing an **external-git** app's `branch` (a draft, or an app bound
+  to an external repository — an app on a Keboola-managed git repo is rejected, its branch
+  is owned by the platform). Source code changes go through the git flow above, not this
+  tool. After a `branch` repoint, call `deploy_data_app` to serve the new branch.
 
 ## Authentication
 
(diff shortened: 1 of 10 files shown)


tomasfejfar:
@AdamVyborny might this be a problem when agent has this tooling for our app workflow? I'm not sure if "external git" is only actually external or (most likely) it also is available for our "internal" git. I think we assume that production version is on main no?

AdamVyborny:
@tomasfejfar @soustruh Yeah, this statement in PR desc:

> An app with a Keboola-managed repo has no git block in its config and can't have its branch changed this way; it's rejected with a clear message before anything is written.

Is not true

soustruh:
@AdamVyborny @tomasfejfar You're right, thanks — the "managed-repo app has
no git block" claim is wrong, and it means the code guards on the wrong
thing: it tells external-git from managed by the presence of
`parameters.dataApp.git` in the config, so a managed/internal app would pass
the guard and get its branch rewritten — exactly the prod-should-stay-on-main
risk Tomáš raised.

Proposed fix: gate on the authoritative `is_managed_git_repo` from the
`.../git-repo` endpoint (we already fetch it in `_build_data_app_with_repo`
but currently discard it) instead of the config git block — allow the branch
repoint only for genuinely external-git apps, and reject managed/internal
ones with a clear message.

Two things I'd like to confirm before pushing that:

1. Is `is_managed_git_repo` the right discriminator, or is there a better
   signal? I don't want to guess the data model.
2. For internal/managed apps, should the branch stay locked to `main` (never
   repointable via this tool), or is there a legitimate case for changing it?

AdamVyborny:
@soustruh 
1. Yep, sounds good to me...
2. IDK, we are in phase of reinventing the flow there. I would keep it as is for now to not modify the behaviour until we decide how it should work.

soustruh:
Pushed the fix. It's now gated on `is_managed_git_repo` from the `.../git-repo`
endpoint (threaded onto `DataApp`), not the presence of a config git block: the
branch repoint is allowed only for external-git apps (a draft, or an app on an
external repo), rejected for Keboola-managed ones, and refused when the flag
can't be determined. Dropped the now-unused git-block helper and corrected the
PR description.

Re point 2: kept managed/internal apps' behaviour unchanged — not repointable
via this tool — as you suggested, so we don't touch the flow while it's being
reworked. A unit test now gives a managed app a git block and still rejects it,
so the discriminator can't silently regress to block-presence.

AdamVyborny (approved):
Reviewed the diff, the review-thread fix, and CFTL-714. The `is_managed_git_repo` gate is correct: the update path populates it via `_fetch_data_app` → `_build_data_app_with_repo`, validation runs before anything is written, and the regression test proves a managed app with a git block is still rejected. Verified against the sandboxes-service OpenAPI spec that `isManagedGitRepo` is a **required** field of the `GET /apps/{appId}/git-repo` 200 response, and the 409 (no git repo) path flows into the fail-closed `None` refusal.

One optional hardening for a follow-up: `AppGitRepoResponse.is_managed_git_repo` defaults to `False` when the field is missing — per spec it never should be, but flipping the default to `None` would turn any spec drift (e.g. an older sandboxes-service) into a refusal instead of a write. Non-blocking.

soustruh:
Thanks for the approval. Applied the optional hardening too: `AppGitRepoResponse.is_managed_git_repo` now defaults to `None` instead of `False`, so a response missing the (spec-required) `isManagedGitRepo` field flows into the fail-closed "could not determine" refusal rather than a write. Added a unit test that a response without the field parses as `None`.

PR #634, opened 2026-07-21
feat: query_data + validate_semantic_query echo optional query_ref

davidesner:
## Description

**Linear**: AI-3576

### Change Type

### Summary

The AI Workspace (business-user Kai) surfaces semantic-layer grounding on chat query results: a badge showing whether a query was validated against the semantic model, which metrics/datasets it used, and any constraint violations, plus inline `[[q:<ref>]]` citations in the answer text. To render that, the UI must know which `validate_semantic_query` call belongs to which `query_data` call — but today the two tool calls share nothing to join on.

This adds an optional `query_ref` parameter to both tools: an opaque correlation token the agent chooses and passes to both calls, echoed back verbatim in each tool's output. The UI joins the two results on it deterministically (with a normalized-SQL fallback when absent). Purely a passthrough — it does not affect execution, and both parameters default to `None`, so existing callers are unaffected.

This correlation is an explicit interim bridge: the north-star is a kai-agent-side "validated query" tool that returns the joined result pre-assembled. The UI contract built on top of `query_ref` is the durable investment.

## Testing

Tested end-to-end with the In Platform Agent running against this server locally (Streamable-HTTP): kbc-ui → kai-agent → local MCP, with the agent instructed to validate-then-run using a shared `query_ref`; the UI correlates and renders the grounding badge. Existing unit tests for both tools pass (18 passed, `pytest -k "query_data or validate_semantic"`).

### Optional testing

## Checklist

- [x] Self-review completed
- [ ] Unit tests added/updated (if applicable) — no new tests; the field is a verbatim passthrough covered by existing tool tests


cjayyy:
Small, well-scoped, low-risk — design is sound. Two notes before merge:

**1. Echo behavior is untested.** `grep query_ref tests/` returns nothing — existing tests assert `isinstance(result, QueryDataOutput)` but never that the token round-trips. The checklist rationale ("covered by existing tool tests") does not hold for the echo, which is the actual contract the UI join depends on. Two ~3-line cases close it:

```python
result = await query_data(sql_query=..., query_name=..., ctx=..., query_ref="abc123")
assert result.query_ref == "abc123"
# + same for validate_semantic_query
```

**2. `query_ref` only echoes on the success path** (`sql.py:336`); the failure/cancel branch raises. Fine if the UI fallback handles "validation present, query failed" gracefully — worth a one-line confirmation.

Everything else is nits (description string duplicated 4×, but inline tool descriptions are the convention here).

cjayyy:
Is this PR R4R?

Matovidlo:
@davidesner is this still relevant ?

davidesner:
> @davidesner is this still relevant ?

@Matovidlo yes, unless we decide to implement the query validation and referencing using a different mechanism inside Kai.

Mystypix:
We need this field for Business Workspace (`apps/business-chat` in `keboola/ui`), where an answer cites the query behind a number. Rather than leave it to rot we'd like to pick the PR up — tracked on our side as UT-5308.

State of the branch, for whoever looks next: ahead 6 / behind 242 and `CONFLICTING` (`pyproject.toml` pinned to 1.74.0 against main's 1.80.1, and `TOOLS.md` is generated), plus an unrelated data-apps commit (`allow_main_branch`) riding along. @cjayyy's two notes from 21 Jul still stand and both look right to me: the echo has no test, and `query_ref` is echoed only on the success path — the failure and cancel branches raise before it.

One thing that isn't in the thread yet and should gate the merge as written. The parameter description instructs the model to *"cite it as `[[q:<query_ref>]]`"*. On our side a single agent prompt serves both the in-platform Kai chat in `kbc-ui` and business-chat, and only business-chat parses that marker (`remarkCitations.ts`); `[[q:` appears nowhere else in the monorepo. Merged as is, kbc-ui users would start seeing literal `[[q:…]]` in answers. So the citation instruction needs to be host-gated or moved out of the tool description into the prompt that knows which chat it is serving.

Two questions before I spend time on it:

1. @Matovidlo — with the 1:1 TypeScript rewrite in progress, should this land in the Python server at all, or go straight to the TS one?
2. Is an agent-chosen opaque token the shape you want long term, or would you rather the server return an id it generates? A model reusing the same ref across turns is a real failure mode for a citation registry, and a server-generated id doesn't have it.

davidesner:
ad 2. the reasoning for model generated was that using long UUIDs means that the model can make a mistake when populating it where as a short sequential id has this risk lower. I have not tested how real this concern is though by running some benchmark. My intuition is that the current models don't really struggle with replicating long strings. BTW having a unique tool call ID that is being used for these references is how this functionality is implemented in the AI chat.

davidesner:
And my take on this is that the citation feature should be available both in BU and normal Kai chat. I see no reason why there should be a chat functionality drift between BU and normal project Kai chat, other than different set of tools and system prompt (persona).

> One thing that isn't in the thread yet and should gate the merge as written. The parameter description instructs the model to "cite it as [[q:<query_ref>]]". On our side a single agent prompt serves both the in-platform Kai chat in kbc-ui and business-chat, and only business-chat parses that marker (remarkCitations.ts); [[q: appears nowhere else in the monorepo. Merged as is, kbc-ui users would start seeing literal [[q:…]] in answers. So the citation instruction needs to be host-gated or moved out of the tool description into the prompt that knows which chat it is serving.

Matovidlo:
1. The typescript rework wont happen very likely I have quite a lot of thigns on me. 
2. I thought this PR is possible to resolve conflicts test and merge + release. 

I see no problem doing this if we need it

Matovidlo (approved):

PR #641, opened 2026-07-27
ci: allow KaiBench evaluation to run on a labelled pull request

jordanrburger:
## Why

KaiBench could only be evaluated on a release tag (`release.yml` calls `kaibench.yml` on `v*` pushes) or by dispatching `kaibench.yml` by hand. **There was no way to trigger it from a pull request.**

This surfaced while trying to validate #640 against the suite: automation that holds `pull_requests: write` but not `actions: write` gets a 403 on `workflow_dispatch`, so it could not start a run and had to hand the commands to a human.

## What

Labelling a PR **`run-kaibench`** now calls `kaibench.yml` via `workflow_call`.

- **Opt-in, not automatic** — a full run takes ~30–60 min and spends real Vertex quota.
- **Forks excluded** — the evaluation needs secrets.
- **Own concurrency group with `cancel-in-progress`** — a new commit supersedes an in-flight evaluation instead of queueing 45 minutes behind it.
- **Job-level `permissions`** (`contents`/`actions`/`packages: read`), mirroring `release.yml`'s top-level block, so CI's global permissions are untouched.
- `build` now skips `labeled` events, so labelling a fork PR doesn't rerun the 3-version matrix for nothing.

Results are reported through a separate `kaibench_gate` job that **fails on regressions, not on absolute failures** — the suite doesn't pass 100% of its questions, so gating on `failed == 0` would be permanently red. A regression is a question that passed on the baseline run and no longer does.

## Three defects found while wiring this up

1. **The baseline lookup never worked under `workflow_call`.** It used `gh run list --workflow kaibench.yml`, which returns nothing — `kaibench.yml` is always invoked via `workflow_call`, so runs are recorded under the *calling* workflow. Now located through the artifact index. This must be paginated: the repo has ~3,300 artifacts and the KaiBench ones sit well past the first page, so a `per_page=100` lookup silently finds nothing.
2. **"0 regressions" was indistinguishable from "no baseline to compare against."** That would have made the new gate pass silently whenever no baseline existed — the worst failure mode for a merge gate. `kaibench-parse-results.py` now emits `baseline_run`, and the gate warns loudly instead of going quietly green.
3. **`--regression-only` is not a valid CLI option** — it's `--regression` (`kaibench/cli.py`). Typer rejects the unknown option; because the step is `continue-on-error`, the run failed silently. Only reachable via the `regression_only` dispatch input, so default runs were unaffected.

## Verification

- Both workflows parse; `kaibench-parse-results.py` compiles.
- Parse script on a fixture where Q2 went pass→fail and Q3 fail→pass: `regressions=1`, `baseline_run=run_old`. With no baseline: `regressions=0`, `baseline_run=` empty.
- Gate script exercised across all four paths: regression → exit 1; clean → exit 0; no baseline → exit 0 **with a warning**; evaluation didn't complete → exit 1.
- Baseline lookup dry-run against this repo resolves run `28088724816` — the v1.72.6 release evaluation from 2026-06-24, unexpired. So the first labelled PR has a real baseline.
- `--regression-only` rejection reproduced directly: `No such option: --regression-only  Did you mean --regression?`

## Follow-up, deliberately not in scope

Baselines come only from release tags, so a PR is compared against the last release (currently v1.72.6, ~5 weeks old). A scheduled run on `main` would keep the baseline current; a stale baseline still needs solver match and sufficient question overlap to be meaningful.

Requires the `run-kaibench` label to exist on this repo.

### DIFF
--- a/.github/scripts/kaibench-parse-results.py
+++ b/.github/scripts/kaibench-parse-results.py
@@ -32,11 +32,15 @@ status = 'passed' if m['failed'] == 0 and m.get('errors', 0) == 0 and partial_co
 print(f"status={status}")
 
 # Count regressions vs previous run (downloaded into prev-results/)
+# `baseline_run` stays empty when no comparison happened, so callers can tell "0 regressions"
+# apart from "never compared" — the two look identical otherwise.
 regressions = 0
+baseline_run = ''
 prev_runs = sorted(Path('prev-results').glob('run_*'), key=lambda p: p.stat().st_mtime) if Path('prev-results').exists() else []
 if prev_runs:
     prev_file = prev_runs[-1] / 'results.jsonl'
     if prev_file.exists() and results_file.exists():
+        baseline_run = prev_runs[-1].name
         prev_by_qid = {}
         for line in prev_file.read_text().splitlines():
             if line.strip():
@@ -51,3 +55,4 @@ if prev_runs:
                 if prev_by_qid[qid].get('status') == 'passed' and r.get('status') not in ('passed', 'skipped'):
                     regressions += 1
 print(f"regressions={regressions}")
+print(f"baseline_run={baseline_run}")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,12 @@ name: CI
 # - For pull requests from branches in this repository, the workflow is skipped to avoid running it twice (once for the
 #   push and once for the PR).
 # - The full workflow (including jobs requiring secrets) runs only on pushes events to branches in this repository
-on: [ push, pull_request ]
+# - KaiBench evaluation is opt-in per pull request: add the `run-kaibench` label. It is the one job that
+#   deliberately runs on `pull_request` (it needs the PR's labels), so it is not double-run by the push event.
+on:
+  push:
+  pull_request:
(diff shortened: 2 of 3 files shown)


jordanrburger:
Found a fourth defect while testing this end-to-end: **the KaiBench checkout is currently broken on `main`.**

Dispatching `kaibench.yml` on a branch fails at *Checkout KaiBench* with `Not Found` — GitHub's 404-for-inaccessible-private-repo. The same checkout succeeded on 2026-06-24 (v1.72.6 release run `28088724816`), so this broke between then and now: the KaiBench repo has since been transferred out of the `keboola-rnd` org and now lives at `keboola/KaiBench`.

**`KAIBENCH_REPO_TOKEN` needs re-issuing** with read access to `keboola/KaiBench` — either it was scoped to `keboola-rnd` resources and the transfer orphaned it, or it expired. That's a secret rotation, so it can't be fixed in this PR, and **KaiBench evaluation cannot run until it is** — including via the new label path.

I've also updated the `repository:` value to the canonical `keboola/KaiBench`. To be clear, that is *not* the fix: the token 404s under either name (the redirect itself resolves fine). It just removes a red herring from the next person's diagnosis.

Matovidlo (approved):

PR #646, opened 2026-07-28
feat(storage): add get_shared_buckets and link_shared_bucket tools [AI-3243]

Matovidlo:
## Description

**Linear**: [AI-3243](https://linear.app/keboola/issue/AI-3243/support-16377-add-keboola-mcp-tool-to-enumerate-shared-with-me-data) (SUPPORT-16377)

**RFC**: [feature_spec/list_shared_buckets/RFC.md](feature_spec/list_shared_buckets/RFC.md)

### Change Type

### Summary

Adds two new Storage tools for the Data Catalog "Shared with me" view:

- **`get_shared_buckets`** — wraps `GET /v2/storage/branch/{branch}/shared-buckets` to list buckets other projects have shared with this one that aren't linked yet. Today `get_buckets` only returns buckets already in the project (including already-linked shares) — there was no way for an agent to answer "what could I link?".
- **`link_shared_bucket`** — wraps `POST /v2/storage/branch/{branch}/buckets` (verified against `keboola/storage-api-php-client`'s `linkBucket()` — the ticket's suggested `/buckets/{id}/link` path does not exist) to link a discovered shared bucket into the project.

**Context-size guardrail**: when this ticket was triaged, the objection to shipping it as-is was that an unbounded shared-buckets list could dump too much into the agent's context for large organizations. `get_shared_buckets` is paginated (`limit`/`offset`, default 50, max 100, mirroring the `search` tool's convention) specifically to close that objection — `total_count` and a `message` field tell the agent whether to page further.

New module `tools/storage/shared_buckets.py` (kept separate from `tools/storage/tools.py`, already 1000+ lines) wired in via `tools/storage/__init__.py`.

## Testing

- [x] Unit tests added/updated:
  - `tests/tools/storage/test_shared_buckets.py` — model field mapping, pagination/limit-clamping, stage-derivation-from-ID-prefix, link success path
  - `tests/clients/test_storage.py` — `shared_bucket_list`/`bucket_link` endpoint/payload assertions
  - `tests/test_server.py` — updated tool list

### Optional testing

## Checklist

- [x] Self-review completed
- [ ] Integration tests added/updated (if applicable) — deferred: no test project with a real pending share available; `get_shared_buckets`/`link_shared_bucket` are unit-tested against mocked responses matching the verified API contract
- [x] Project version bumped according to the change type (1.73.4 → 1.74.0)
- [x] Documentation updated (`TOOLS.md` regenerated)


Matovidlo:
@claude review

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,189 @@
...
+            tags={STORAGE_TOOLS_TAG},
+        )
+    )
+    mcp.add_tool(
+        FunctionTool.from_function(
+            link_shared_bucket,
+            annotations=ToolAnnotations(destructiveHint=False),
+            serializer=toon_serializer_compact,
+            tags={STORAGE_TOOLS_TAG},
+        )
```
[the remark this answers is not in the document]
Matovidlo: `destructiveHint=False` here is intentional, not an oversight. Per the MCP tool-annotation semantics, `destructiveHint` marks whether the tool performs *destructive updates* vs *purely additive* ones — `link_shared_bucket` only creates a brand-new local bucket, it doesn't modify or delete any existing resource. That matches this repo's existing convention for creation tools: `create_config` uses the identical `ToolAnnotations(destructiveHint=False)` (`tools/components/tools.py:146`) despite also being a side-effecting POST. `update_descriptions` (`destructiveHint=True`) is a different case — it overwrites existing state. Keeping `False` here for consistency with the create-tool pattern.

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,189 @@
...
+        limit = DEFAULT_SHARED_BUCKETS_LIMIT
+    offset = max(0, offset)
+
+    client = KeboolaClient.from_state(ctx.session.state)
+    raw_shared_buckets = await client.storage_client.shared_bucket_list()
+    all_buckets = sorted((SharedBucketDetail.model_validate(raw) for raw in raw_shared_buckets), key=lambda b: b.id)
+
+    total_count = len(all_buckets)
+    page = all_buckets[offset : offset + limit]
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit 058ca512f0536ace8bd0cf24ef54804691b3a723

Sort/slice the raw dicts first, then validate only the requested page into `SharedBucketDetail` (shared_buckets.py:134-138), instead of building Pydantic models for the full list before slicing.

cjayyy (commented):
Review of the shared-buckets tools. Overall a well-researched change — verifying the link endpoint against `keboola/storage-api-php-client` rather than trusting the ticket's suggested `/buckets/{id}/link` path was the right call, and I independently confirmed both contract claims hold: `Client::linkBucket()` does post to `buckets` with exactly that payload, and `BranchAwareClient::request()` does rewrite it to `branch/{id}/buckets` (`buckets` is not in `START_ENDPOINTS_WITHOUT_BRANCH`). The tool-side pagination is also a correct read of the endpoint — `listSharedBuckets()` accepts only `include`, no `limit`/`offset` — and it genuinely closes the "non-pageable" triage objection.

One blocker (inline on `shared_buckets.py`): `BucketDetail.model_validate()` on the link response assumes a shape the reference client treats as `{id}` only, and the unit test mock hides it. Everything else is minor.

Security: nothing concerning. No new auth surface, no injection vectors (IDs go into a JSON body), no secrets logged, and `linked_by` only exposes what the Data Catalog UI already shows.

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,201 @@
...
+
+    client = KeboolaClient.from_state(ctx.session.state)
+    raw_bucket = await client.storage_client.bucket_link(
+        name=target_bucket_name,
+        stage=stage,
+        source_project_id=source_project_id,
+        source_bucket_id=source_bucket_id,
+        display_name=display_name,
+    )
+    return BucketDetail.model_validate(raw_bucket)
```
cjayyy: **Blocker: this will likely raise `ValidationError` — `POST /buckets` appears to return only `{id}`.**

`BucketDetail` has required fields `id`, `name`, `display_name`, `stage`, `created`. But in `keboola/storage-api-php-client`, both `linkBucket()` and `createBucket()` are typed and used as if the response carries `id` alone:

```php
/** @var array{id: string} $result */
$result = $this->apiPostJson($url, $options, $async);
return $result['id'];
```

The reference client never reads any other field off this response. The RFC states "the link response has the same shape as `bucket_detail`" — that's the one contract claim in the RFC that isn't backed by the reference client, and the unit test mocks a full payload so it passes without telling us anything about the real shape.

The failure mode is bad: the link **already succeeded** server-side (non-idempotent, no retry path), and the agent gets an opaque validation error. Suggest fetching the detail explicitly so the documented return type is guaranteed:

```python
raw_bucket = await client.storage_client.bucket_link(...)
bucket_id = cast(str, raw_bucket['id'])
return BucketDetail.model_validate(await client.storage_client.bucket_detail(bucket_id))
```

One extra call, and the tool's contract holds regardless of what `POST /buckets` returns. Alternatively, verify against a live response first — the RFC's step 5 (manual smoke test against a project with a real share) would settle it.

Related: `linkBucket($name, $stage, $sourceProjectId, $sourceBucketId, $displayName = null, $async = false)` takes an `$async` flag and `apiPostJson(..., $handleAsyncTask)`, so it's worth confirming the sync path doesn't return a job envelope (`{id, status, url}`) for large shares. If it can, this fix becomes mandatory rather than defensive.

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,201 @@
...
+        LOG.warning(
+            f'The "limit" parameter is out of range (0, {MAX_SHARED_BUCKETS_LIMIT}], setting to default value '
+            f'{DEFAULT_SHARED_BUCKETS_LIMIT}.'
+        )
+        limit = DEFAULT_SHARED_BUCKETS_LIMIT
+    offset = max(0, offset)
+
+    client = KeboolaClient.from_state(ctx.session.state)
+    raw_shared_buckets = await client.storage_client.shared_bucket_list()
+    raw_shared_buckets = sorted(raw_shared_buckets, key=lambda raw: raw['id'])
```
cjayyy: `raw['id']` raises a bare `KeyError` if any item lacks `id`, taking down the whole listing. Every other field in the model tolerates absence, so this is the one spot where a malformed payload fails without a useful message.

```suggestion
    raw_shared_buckets = sorted(raw_shared_buckets, key=lambda raw: raw.get('id') or '')
```

`SharedBucketDetail.id` is required so validation would still fail downstream — but with a Pydantic error that names the field.

(The stable sort itself is the right call — without it, pagination over an unpaginated upstream would be non-deterministic across calls.)

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,201 @@
...
+    stage: str = Field(description='Stage of the shared bucket (in for input stage, out for output stage).')
+    description: str | None = Field(None, description='Description of the shared bucket.')
+    project_id: str | None = Field(default=None, description='The ID of the project the shared bucket belongs to.')
+    project_name: str | None = Field(default=None, description='The name of the project the shared bucket belongs to.')
+    sharing: str | None = Field(
+        default=None,
+        description=(
+            "Sharing scope: 'organization' (all projects in the organization), "
+            "'organization-project' (all projects in the source project's organization), or "
+            "'specific-projects' (only projects explicitly targeted)."
```
cjayyy: Two issues with this enum documentation — and since it's what the agent reasons over, imprecision has real cost:

1. **A fourth scope is missing.** The PHP client exposes four sharing paths: `shareOrganizationBucket`, `shareOrganizationProjectBucket`, `shareBucketToProjects`, and `shareBucketToUsers` — so a `specific-users` scope exists and isn't documented here.
2. **The two org descriptions don't distinguish themselves.** "all projects in the organization" vs "all projects in the source project's organization" reads as the same thing; the actual difference isn't conveyed.

Worth pinning the exact server-side value set (and their real semantics) before shipping, since the agent may filter or explain based on this text.

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,201 @@
...
+    """A bucket shared with this project (Data Catalog "Shared with me") that may not be linked yet."""
+
+    id: str = Field(description='Unique identifier of the shared bucket in its source project.')
+    display_name: str = Field(
+        description='The display name of the shared bucket.',
+        validation_alias=AliasChoices('displayName', 'display_name'),
+    )
+    stage: str = Field(description='Stage of the shared bucket (in for input stage, out for output stage).')
+    description: str | None = Field(None, description='Description of the shared bucket.')
+    project_id: str | None = Field(default=None, description='The ID of the project the shared bucket belongs to.')
```
cjayyy: Storage API project IDs come back as JSON integers. `ProjectInfo.project_id` in `tools/project.py:102` is declared `str | int` for exactly this reason.

As written, Pydantic coerces `123` → `'123'`, and then `link_shared_bucket(source_project_id: str)` round-trips a stringified int back to the API. It works, but `str | int` matches the codebase's existing accommodation and avoids the implicit coercion.

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,201 @@
...
+        description='Total data size of the shared bucket in bytes.',
+        validation_alias=AliasChoices('dataSizeBytes', 'data_size_bytes'),
+    )
+
+    @model_validator(mode='before')
+    @classmethod
+    def set_project_fields(cls, values: dict) -> dict:
+        if project := cast(dict | None, get_nested(values, 'project')):
+            values.setdefault('project_id', project.get('id'))
+            values.setdefault('project_name', project.get('name'))
```
cjayyy, line 96: Minor, and consistent with `BucketDetail.set_source_project`, so take or leave: `setdefault` (rather than direct assignment as in `tools.py:242-245`) means a raw payload that already carries a snake_case `project_id` silently wins over `project.id`. Probably intentional, but it isn't obvious from the code.

Also note this mutates the caller's dict in place — harmless here since each raw dict is used once.

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,201 @@
...
+            get_shared_buckets,
+            annotations=ToolAnnotations(readOnlyHint=True),
+            serializer=toon_serializer_compact,
+            tags={STORAGE_TOOLS_TAG},
+        )
+    )
+    mcp.add_tool(
+        FunctionTool.from_function(
+            link_shared_bucket,
+            annotations=ToolAnnotations(destructiveHint=False),
```
cjayyy, line 38: `destructiveHint=False` is right (this creates, doesn't destroy), but the operation is **not** idempotent — a second call with the same `target_bucket_name` will 4xx. Consider making that explicit so clients don't auto-retry:

```python
annotations=ToolAnnotations(destructiveHint=False, idempotentHint=False),
```

src/keboola_mcp_server/tools/storage/shared_buckets.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Links a bucket shared with this project (found via `get_shared_buckets`) into this project
+    as a new local bucket, so its tables become directly queryable/joinable like any other
+    bucket in the project.
+    """
+    stage = target_stage or (source_bucket_id.split('.', 1)[0] if '.' in source_bucket_id else None)
+    if stage not in ('in', 'out'):
+        raise ToolError(
+            f'Could not determine stage from source_bucket_id={source_bucket_id!r}; pass target_stage explicitly.'
+        )
```
cjayyy, line 199: **Dev-branch semantics:** `get_buckets` does substantial prod/dev shading work (`has_storage_branches`, `_combine_buckets`, `shade_by`), but this returns a bare `BucketDetail` with `branch_id`/`prod_id` unset and no links — an object that doesn't participate in the shading model the rest of the storage tools assume.

The RFC explicitly rejects fitting shares into `BucketDetail`'s identity model for `get_shared_buckets` ("no `branch_id`/`prod_id` shading concept"), then returns `BucketDetail` from the link tool. Worth either setting `branch_id` from the client, or a sentence in the RFC on why linking is safe here.

tests/tools/storage/test_shared_buckets.py
```diff
@@ -0,0 +1,210 @@
...
+        self, mocker: MockerFixture, mcp_context_client: Context
+    ) -> None:
+        keboola_client = KeboolaClient.from_state(mcp_context_client.session.state)
+        keboola_client.storage_client.bucket_link = mocker.AsyncMock(
+            return_value={
+                'id': 'in.c-linked',
+                'name': 'c-linked',
+                'displayName': 'Linked bucket',
+                'stage': 'in',
+                'created': '2026-01-01T00:00:00+0000',
```
cjayyy: This mock is the crux of the blocker on `shared_buckets.py:201` — it supplies a full bucket payload (`name`, `displayName`, `stage`, `created`), so the test passes while the reference client suggests `POST /buckets` returns `{id}` only. The most valuable test in the file is currently asserting against an assumed shape.

Once the real response is confirmed, please narrow this mock to match it.

tests/tools/storage/test_shared_buckets.py
```diff
@@ -0,0 +1,210 @@
...
+        keboola_client = KeboolaClient.from_state(mcp_context_client.session.state)
+        keboola_client.storage_client.shared_bucket_list = mocker.AsyncMock(return_value=[_shared_bucket_raw('in.c-a')])
+
+        result = await get_shared_buckets(mcp_context_client, offset=-5)
+
+        assert len(result.shared_buckets) == 1
+
+
+@pytest.mark.asyncio
+class TestLinkSharedBucket:
```
cjayyy, line 168: **Missing the error-path test the RFC scoped:** "otherwise cover it with a unit test against a mocked 4xx/409 (already-linked) response."

That's the most likely real-world failure — agent lists shares, links one that's already linked — and nothing here verifies the agent gets an actionable message rather than a raw HTTP error through `tool_errors()`. A 403 (no access to the share) case would be worth adding alongside it.

Also not covered: a raw payload with `project` **absent**, i.e. the false branch of `if project := ...` in `set_project_fields`.

The rest of the coverage here is solid — 4 parametrized pagination cases including offset-beyond-total, limit clamping, negative offset, stage derivation from both prefixes, and explicit override.

src/keboola_mcp_server/clients/storage.py
```diff
@@ -462,6 +462,48 @@ async def bucket_metadata_delete(self, bucket_id: str, metadata_id: str) -> None
...
+        :return: The newly linked bucket's details as a dictionary
+        """
+        bid = branch_id or self._branch_id
+        data: JsonDict = {
+            'name': name,
+            'stage': stage,
+            'sourceProjectId': source_project_id,
+            'sourceBucketId': source_bucket_id,
+        }
+        if display_name is not None:
```
cjayyy, line 501: This diverges from the reference client, which uses truthy `if ($displayName)` and therefore **drops** an empty string:

```php
if ($displayName) {
    $options['displayName'] = $displayName;
}
```

`is not None` sends `displayName: ''` instead, which the API may well reject. `test_bucket_link_sends_explicit_empty_display_name` is a nice catch and pins the Python behavior — but it documents the divergence rather than establishing that the API accepts it. Worth either matching the PHP truthiness or confirming `''` is valid server-side.

src/keboola_mcp_server/tools/storage/__init__.py
```diff
@@ -1,3 +1,13 @@
...
+from keboola_mcp_server.mcp import KeboolaMcpServer
+from keboola_mcp_server.tools.storage.shared_buckets import add_shared_bucket_tools
+from keboola_mcp_server.tools.storage.tools import STORAGE_TOOLS_TAG
+from keboola_mcp_server.tools.storage.tools import add_storage_tools as _add_storage_tools
+
+
+def add_storage_tools(mcp: KeboolaMcpServer) -> None:
+    """Adds all storage tools (buckets, tables, shared buckets) to the MCP server."""
+    _add_storage_tools(mcp)
+    add_shared_bucket_tools(mcp)
```
cjayyy, line 14: Nit: shadowing an imported name with a same-named wrapper (`add_storage_tools as _add_storage_tools`, then redefining `add_storage_tools`) costs some readability. It does keep `server.py` untouched, which is a fair trade.

Alternative: call `add_shared_bucket_tools(mcp)` from within `tools.py`'s own `add_storage_tools`, dropping this module to a plain re-export.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Fixed** ✅ in commit eff9b4d0

Applied your suggested fix exactly: link_shared_bucket now fetches bucket_detail(id) explicitly instead of trusting bucket_link's response shape. One extra call, and the contract holds regardless of what POST /buckets actually returns.

tests/tools/storage/test_shared_buckets.py
Matovidlo: **Fixed** ✅ in commit eff9b4d0

Narrowed the mock to match: bucket_link now returns {'id': ...} only, and bucket_detail (also mocked) supplies the full payload -- pinned to the real two-call contract.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Fixed** ✅ in commit eff9b4d0

raw.get('id') or '' -- a missing id now fails with SharedBucketDetail's own Pydantic error naming the field, not a bare KeyError from the sort.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Fixed** ✅ in commit eff9b4d0

source_project_id (link_shared_bucket param) and SharedBucketDetail.project_id both now accept str | int, matching ProjectInfo.project_id's existing accommodation for the same API behavior.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Won't fix** ℹ️

Tried this and reverted -- it fails test_tool_annotations_and_tags, which enforces (across the whole codebase) that a destructiveHint=False tool leaves idempotentHint unset, matching create_config and every other create-style tool here. Adding it just for this tool would be a real, if narrow, inconsistency with that convention. Called out the actual retry hazard in the docstring instead ("Not idempotent: calling this again with the same target_bucket_name fails once that name already exists") so the information reaches the agent even without the annotation.

src/keboola_mcp_server/clients/storage.py
Matovidlo: **Won't fix** ℹ️

Leaving this as `is not None` for now -- it's the deliberate fix from an earlier round (see the commit that introduced it: a prior truthy check was silently dropping an intentionally-set empty string). Matching the PHP client's truthy-drop would reintroduce that bug for the (rare) case of an explicit empty display name, and I don't have a way to confirm live whether the API actually rejects `displayName: ''`. If you've seen it reject empty string in practice, happy to revisit -- otherwise this is a real disagreement between two plausible readings rather than a clear bug.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Already addressed** ✅ in commit eff9b4d0

Improved the doc to add the missing specific-users scope and distinguish organization vs organization-project more clearly, and explicitly flagged that the value set isn't fully pinned server-side. Didn't pin exact values since I don't have a way to confirm them live -- happy to tighten further once someone can check against a real project with each sharing type.

tests/tools/storage/test_shared_buckets.py
Matovidlo: **Fixed** ✅ in commit eff9b4d0

Added: already-linked (409) and no-access (403) errors from bucket_link now have tests confirming they propagate rather than get swallowed (they surface as the raw httpx.HTTPStatusError today -- no friendlier mapping added, since that'd be a behavior change beyond what you flagged). Also added a raw-payload-missing-project test for set_project_fields's false branch.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Intentional** ℹ️

Documented rather than fixed -- added a docstring note that the linked bucket doesn't participate in this server's prod/dev branch-shading. Looked into fixing it properly: BucketDetail.branch_id is derived from a legacy FAKE_DEVELOPMENT_BRANCH metadata marker (tools.py), not a plain API field, and populating it correctly here would mean pulling that whole mechanism into a code path that doesn't otherwise touch it. That felt like the wrong place to make that call without understanding the legacy branch-detection model in more depth than this PR should require -- flagging it as a known gap instead of guessing at a fix.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Intentional** ℹ️

Leaving as setdefault, per your own "take or leave" -- agreed it's intentional (a raw payload that already carries a flattened project_id should win over the nested project.id) and matches the spirit of BucketDetail.set_source_project elsewhere, even if not byte-identical.

src/keboola_mcp_server/tools/storage/__init__.py
Matovidlo: **Fixed** ✅ in commit eff9b4d0

Went with a module-qualified call (`from keboola_mcp_server.tools.storage import tools as _tools`, then `_tools.add_storage_tools(mcp)`) rather than your suggested alternative of calling add_shared_bucket_tools from inside tools.py's own add_storage_tools -- that would make tools.py import shared_buckets.py, which imports STORAGE_TOOLS_TAG/BucketDetail back from tools.py, i.e. a circular import. This __init__.py has to stay the composition point; added a comment explaining why so it doesn't look like an arbitrary shadow next time.

Matovidlo:
Thanks for the review, @vaclavnosek — replied inline to each point. Summary of what changed in eff9b4d0:

- **Blocker fixed**: `link_shared_bucket` now fetches `bucket_detail(id)` explicitly instead of trusting the link response's shape (also fixes the test mock that was hiding it).
- **Fixed**: `raw.get('id')` instead of `raw['id']` in the sort key; `source_project_id`/`project_id` widened to `str | int`; sharing-scope enum doc improved (added the missing `specific-users` scope); added the 409/403/missing-`project` test cases.
- **Left as-is, with rationale inline**: `idempotentHint=False` (conflicts with this codebase's `destructiveHint=False` → `idempotentHint` unset convention, enforced by an existing test — noted the retry hazard in the docstring instead), the `displayName` truthy-vs-`is not None` question (standing disagreement, no way to confirm live which the API actually wants), the `setdefault` project-field precedence (agreed intentional per your own "take or leave"), and dev-branch shading (documented as a known gap — fixing it properly means touching the legacy `FAKE_DEVELOPMENT_BRANCH` metadata mechanism, which felt like more than this PR should take on blind).

Full test suite (1699 tests) + black/isort/flake8/`check-tools-docs` all green on top of the fixes.

src/keboola_mcp_server/tools/storage/shared_buckets.py
cjayyy: **Upgrading this from a nit to a bug — verified locally.** Pydantic v2 does not coerce int→str here, so an integer `project.id` fails validation and `get_shared_buckets` returns **nothing**:

```
ToolError: Found 1 validation error(s) for SharedBucketDetail
- field: project_id
  message: Input should be a valid string
  extra: {type: string_type, input: '1234'}
```

Because every item in the page is validated, a single int `project.id` upstream takes down the entire listing.

Storage API project IDs are integers in JSON — and this repo's own fixture for the same nested `project` object confirms it: `tests/tools/storage/test_tools.py:325` uses `'project': {'id': 1234, 'name': 'A demo project'}`.

```suggestion
    project_id: str | int | None = Field(default=None, description='The ID of the project the shared bucket belongs to.')
```

Matches `ProjectInfo.project_id` (`tools/project.py:102`), which is already `str | int` for this reason. Verified: with this one-line change every pagination case passes against a 120-item stub.

tests/tools/storage/test_shared_buckets.py
```diff
@@ -0,0 +1,210 @@
...
+from keboola_mcp_server.tools.storage.tools import BucketDetail
+
+
+def _shared_bucket_raw(bucket_id: str, **overrides: Any) -> dict[str, Any]:
+    raw = {
+        'id': bucket_id,
+        'displayName': f'Display {bucket_id}',
+        'stage': 'in',
+        'description': 'A shared bucket.',
+        'project': {'id': 'proj-1', 'name': 'Source Project'},
```
cjayyy, line 38: This string `'proj-1'` is the only reason the int-coercion bug on `shared_buckets.py:57` isn't caught. Compare `tests/tools/storage/test_tools.py:325`, which fixtures the same nested object as `'project': {'id': 1234, 'name': 'A demo project'}`.

Switching to an int makes this fixture match the real payload — and it fails on the current model, which is the point:

```suggestion
        'project': {'id': 1234, 'name': 'Source Project'},
```

src/keboola_mcp_server/tools/storage/shared_buckets.py
cjayyy: **Confirmed by local test.** Stubbing `POST /buckets` to return `{"id": "in.c-linked"}` — the shape `Client::linkBucket()`'s `@var array{id: string}` implies — produces:

```
ToolError: Found 4 validation error(s) for BucketDetail
- name: Field required
- displayName: Field required
- stage: Field required
- created: Field required
```

And the server's own telemetry proves the side effect lands before the throw:

```
POST /v2/storage/branch/default/buckets   <- succeeds
POST /v2/storage/events  {"message": "MCP tool \"link_shared_bucket\" call failed. ToolError: Found 4 validation error(s)..."}
```

So on the PHP-implied shape the bucket **is linked** but the agent is told the call failed — and a retry then hits already-exists. The same stub returning a full bucket passes cleanly, which is exactly what the unit test mocks.

The `bucket_detail` follow-up I suggested makes this immune to the response shape either way.

cjayyy (commented):
**Follow-up: tested locally in Docker.** Built the image from this PR's head (`a42c8227`) and drove both tools over the real MCP protocol against a stubbed Storage API. Two things to report — one new bug, and confirmation of the blocker.

**Setup:** `docker build` from the PR head, then FastMCP client → server → HTTPS stub on `connection.localtest` (the client rewrites any storage URL to `https://connection.<suffix>` at `clients/client.py:171`, so the stub has to be HTTPS on :443).

**Test suite:** 1693 passed, including all 36 new tests. Both tools register correctly with the expected annotations.

---

### 1. NEW — `get_shared_buckets` fails outright when `project.id` is an integer

Not a style nit (I originally filed it as one) — it's a hard failure. Pydantic v2 does not coerce int→str in this position:

```
ToolError: Found 1 validation error(s) for SharedBucketDetail
- field: project_id
  message: Input should be a valid string
  extra: {type: string_type, input: '1234'}
```

The tool returns **zero** results — one int `project.id` anywhere in the response fails the whole page, since every item is validated.

This repo's own fixture for the same nested `project` object uses an int: `tests/tools/storage/test_tools.py:325` → `'project': {'id': 1234, 'name': 'A demo project'}`. The new test fixture uses `'proj-1'` (a string), which is the only reason the suite passes. Storage API project IDs are integers in JSON.

Confirmed boundary:

```
int id (real API + repo's existing fixture) -> FAIL: Input should be a valid string
str id (this PR's fixture)                  -> PASS: project_id = 'proj-1'
```

Fix: `project_id: str | int | None` (matching `ProjectInfo.project_id` at `tools/project.py:102`, which is `str | int` for exactly this reason). Please also change the test fixture to an int so it reflects the real payload. With that one-line change, all pagination cases below pass.

### 2. CONFIRMED — the `BucketDetail` blocker, with the side effect ordering proven

Stubbed `POST /buckets` to return `{"id": "in.c-linked"}` (what `Client::linkBucket()`'s `@var array{id: string}` implies):

```
ToolError: Found 4 validation error(s) for BucketDetail
- name: Field required
- displayName: Field required
- stage: Field required
- created: Field required
```

The telemetry event the server itself emitted confirms the damage ordering:

```
POST /v2/storage/branch/default/buckets  {"name":"linked","stage":"in","sourceProjectId":"1234","sourceBucketId":"in.c-shared-000"}
POST /v2/storage/events  {"message":"MCP tool \"link_shared_bucket\" call failed. ToolError: Found 4 validation error(s)..."}
```

The link **succeeds server-side**, then the tool throws. The agent sees only a failure and will likely retry — which then 4xx's on already-exists. Same stub returning a full bucket passes cleanly, so the only thing standing between this and a broken write path is the real response shape.

Worth settling with one manual call against a project with a real share (the RFC's step 5). If it does return the full bucket, disregard — but the reference client is the only evidence either way right now, and it points the other way.

### 3. Verified working

The pagination logic is solid — every case correct against 120 stubbed shares:

| Call | Rows | `message` |
|---|---|---|
| defaults | 50 | `Returning 50 of 120 shared buckets. Use offset=50 to see more.` |
| `limit=200` | 50 | clamped to default |
| `limit=-1` | 50 | clamped to default |
| `offset=115, limit=10` | 5 | `Returning 5 of 120 shared buckets.` |
| `offset=500` | 0 | `Returning 0 of 120 shared buckets.` |

Also verified: `POST` payload is exactly right (`name`/`stage`/`sourceProjectId`/`sourceBucketId`, `displayName` omitted when `None`); branch-scoped paths resolve to `/v2/storage/branch/default/{shared-buckets,buckets}`; stage derivation works from both prefixes; and the 400 already-exists path surfaces a readable `API error: The bucket already exists` through `tool_errors()` — good, though still worth a regression test.

---

**Note:** the PR description says `1.73.4 → 1.74.0`, but `pyproject.toml` at this head is `1.75.0` (base moved on). Just a stale description.

tests/tools/storage/test_shared_buckets.py
cjayyy, line 168: Good news on the error path: I stubbed a 400 already-exists and `tool_errors()` does surface something actionable —

```
ToolError: Error calling tool 'link_shared_bucket': Client error '400 Bad Request' for url '.../branch/default/buckets'
API error: The bucket already exists
```

So this isn't broken, just uncovered. Still worth pinning as a regression test, since it's the most likely real-world path and the message quality is what the agent depends on to recover.

tests/tools/storage/test_shared_buckets.py
```diff
@@ -0,0 +1,287 @@
...
+    return raw
+
+
+def _shared_bucket_raw(bucket_id: str, **overrides: Any) -> dict[str, Any]:
+    raw = {
+        'id': bucket_id,
+        'displayName': f'Display {bucket_id}',
+        'stage': 'in',
+        'description': 'A shared bucket.',
+        'project': {'id': 'proj-1', 'name': 'Source Project'},
```
cjayyy, line 38: Last loose end from the int-`project_id` bug: the model is fixed, but this fixture still uses a **string**, so the suite has no regression guard for the exact failure that took `get_shared_buckets` down at `a42c8227`.

I flipped it locally to confirm the model handles the real shape — it does, and only the assertion needs to follow:

```suggestion
        'project': {'id': 1234, 'name': 'Source Project'},
```

With that, line 52 needs `assert bucket.project_id == 1234`. Verified result: `1 failed, 20 passed`, and the failure is purely the stale string assertion — `SharedBucketDetail` itself validates the int cleanly (`project_id=1234`).

Worth doing so the int path is actually pinned; right now `str | int` is only exercised by the type annotation, not by a test.

cjayyy (commented):
**Re-tested at `eff9b4d0` — both fixes verified working.** Rebuilt from the new head and re-ran the same Docker harness against a stubbed Storage API.

### Blocker: resolved ✅

The `bucket_detail` follow-up does exactly what it needs to. With `POST /buckets` stubbed to return `{"id": "in.c-linked"}` only, the observed call sequence is now:

```
POST /v2/storage/branch/default/buckets
GET  /v2/storage/branch/default/buckets/in.c-linked
```

→ returns a valid `BucketDetail` (`id: in.c-linked | name: c-linked | display_name: Linked | stage: in | created: ...`). Previously this raised a 4-field `ValidationError` *after* the link had already committed. The tool is now immune to the response shape either way, which was the point.

### `project_id` int bug: resolved ✅

`str | int | None` fixes it. With an int `project.id` (the real API shape) the tool now returns all 50 rows and serializes `project_id: 1234` correctly — no validation error. Nice touch converting with `str(source_project_id)` at the client boundary so the API still receives a string.

### Regression check — all green

| Call | Rows | |
|---|---|---|
| defaults | 50 | OK |
| `limit=200` → default | 50 | OK |
| `limit=-1` → default | 50 | OK |
| `offset=115, limit=10` | 5 | OK |
| `offset=500` | 0 | OK |

Error paths still behave: 400 already-exists surfaces `API error: The bucket already exists`; unparseable stage raises the explicit `ToolError` before any API call.

**Full suite: 1699 passed** (up from 1693 — the six new tests).

### One loose end (inline)

The test fixture still uses a *string* `project.id`, so nothing in the suite would catch a regression of the int bug. I verified that flipping it to `1234` makes the model validate fine — only the assertion on line 52 needs updating alongside it. Details inline.

Thanks for the quick turnaround on these — the `destructiveHint` and `is not None` push-backs both make sense.

tests/tools/storage/test_shared_buckets.py
cjayyy: cc @Matovidlo

Matovidlo: **Already addressed** ✅ in commit 19ccbdd65ad976e961aa0a19a9f57accd458c0d6

The int-coercion bug you found is regression-tested directly: `test_integer_project_id_is_accepted` (tests/tools/storage/test_shared_buckets.py:69-73) validates a `project={'id': 123, ...}` payload and asserts `bucket.project_id == 123`, so the exact failure mode you traced is pinned. Left the shared `_shared_bucket_raw` default fixture on the string `'proj-1'` rather than flipping it -- it's used by ~8 other pagination/sorting tests that don't care about `project_id`'s type, and flipping it would just mean updating unrelated assertions for no added coverage. Happy to flip it too if you'd still rather have every test exercise the int path, but the regression itself is covered either way.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Already addressed** ✅ in commit 19ccbdd65ad976e961aa0a19a9f57accd458c0d6

Confirmed and fixed: `project_id` is `str | int | None` (shared_buckets.py:57), matching `ProjectInfo.project_id`. Regression-tested by `test_integer_project_id_is_accepted`.

tests/tools/storage/test_shared_buckets.py
Matovidlo: **Already addressed** ✅ in commit 19ccbdd65ad976e961aa0a19a9f57accd458c0d6

Same fix as the sibling thread on this line -- `test_integer_project_id_is_accepted` (test_shared_buckets.py:69-73) pins the exact int-coercion bug you traced via a dedicated fixture override, rather than flipping the shared `_shared_bucket_raw` default. See my reply on the other thread on this line for the reasoning on why I left the shared fixture as a string.

src/keboola_mcp_server/tools/storage/shared_buckets.py
Matovidlo: **Already addressed** ✅ in commit 19ccbdd65ad976e961aa0a19a9f57accd458c0d6

Confirmed and fixed: link_shared_bucket now fetches bucket_detail(id) explicitly instead of trusting bucket_link's response shape, so this is immune to whatever POST /buckets actually returns. Thanks for verifying the side-effect-lands-before-throw failure mode locally -- that's exactly why the extra call is worth it despite the cost.

tests/tools/storage/test_shared_buckets.py
Matovidlo: **Fixed** ✅ in commit b01961faeda29d6e9669377cbff102dfe2471b56

Pinned as a regression test: merged the 409/403 tests into one parametrized test_bucket_link_http_error_propagates and added the 400 already-exists case you stubbed locally, so all three HTTP error paths (400/403/409) are now covered.

cjayyy (approved):

PR #652, opened 2026-08-03
feat(AI-3112): support keboola.duckdb-transformation via generic config tools

Matovidlo:
## Description

**Linear**: [AI-3112](https://linear.app/keboola/issue/AI-3112/kai-supporting-creation-of-duckdb-transformations)

### Change Type

### Summary

Adds `keboola.duckdb-transformation` as a supported transformation type, routed through the
existing generic `create_config`/`update_config` tools — the same path Python/R transformations
already use. `create_sql_transformation`/`update_sql_transformation` stay Snowflake/BigQuery-only,
since their component ID is derived from the project's workspace SQL dialect and DuckDB has no
cloud workspace backend.

Changes:
- `utils.py`: new `DUCKDB_TRANSFORMATION_ID` constant, added to `FOLDER_SUPPORTING_COMPONENT_IDS`
  so folder metadata management works the same as for Python/R.
- `tools.py`: docstrings for `create_config`, `update_config`, `create_sql_transformation`, and
  `update_sql_transformation`, plus the `update_sql_transformation` 404 error message, updated to
  mention DuckDB alongside Python/R so the model routes DuckDB requests correctly on both the
  create and update paths.
- No new tools, no workspace changes. DuckDB-specific sync actions (`syntax_check`,
  `lineage_visualization`, `execution_plan_visualization`, `expected_input_tables`) and config
  parameters (`backend_size`, `timeout`, `duckdb_version`, etc.) require no new code — they're
  already handled generically by `run_sync_action` and the opaque-JSON `parameters` field.

RFC: `feature_spec/duckdb_transformation_support/RFC.md`

### Design note (per @tomasfejfar's review)

DuckDB transformations are conceptually SQL transformations, and ideally would go through
`create_sql_transformation`/`update_sql_transformation` like Snowflake/BigQuery, not the
Python/R-oriented `create_config`/`update_config` path. That's a real limitation of this PR, not
a design preference — `create_sql_transformation`'s component ID is derived from
`WorkspaceManager.get_sql_dialect()`, which only maps Snowflake/BigQuery (DuckDB has no cloud
workspace to plug into), so making DuckDB fit there is a larger change than this PR's scope.
This is exactly what [AI-3662](https://linear.app/keboola/issue/AI-3662/consolidate-transformation-tools-and-move-guidance-into-project-system)
(linked above) is scoped to address — tracking it explicitly here so it isn't lost.

### Resolved: `get_config_examples`/`get_components` discovery gap

Both tools the system prompt directs an agent to consult before `create_config` initially returned
nothing usable for `keboola.duckdb-transformation` — the underlying example database lives in a
separate service (`keboola/kai-bot`, `component_configuration_examples.jsonl`) and had no entry
for this component. [keboola/kai-bot#236](https://github.com/keboola/kai-bot/pull/236) (merged
2026-08-20) added one, so `get_config_examples` now returns a real example. The inline DuckDB
stopgap example that had been added to `create_config`'s docstring for this gap has been removed
(per @tomasfejfar's review).

## Testing

### Optional testing

## Checklist

- [x] Self-review completed

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -245,6 +245,14 @@ WHEN NOT TO USE:
 - `keboola.data-apps` → use `modify_python_js_data_app` / `modify_streamlit_data_app` / `deploy_data_app`
 - `keboola.snowflake-transformation` / `keboola.google-bigquery-transformation` → use `create_sql_transformation`
 
+DUCKDB TRANSFORMATION EXAMPLE (`component_id='keboola.duckdb-transformation'`) — `get_config_examples`
+may not have real-world examples for this component yet, so `parameters` follows the same
+`blocks[].codes[].script` shape as `create_sql_transformation`, using DuckDB SQL:
+    {"blocks": [{"name": "Clean data", "codes": [{"name": "Dedupe", "script": [
+        'CREATE OR REPLACE TABLE "clean" AS SELECT * FROM "raw" '
+        'QUALIFY ROW_NUMBER() OVER (PARTITION BY "id" ORDER BY "updated_at" DESC) = 1;'
+    ]}]}]}
+
 EXAMPLES:
 - user_input: `Create a new configuration for component X with these settings`
     - set the component_id and configuration parameters accordingly
@@ -820,7 +828,9 @@ WHEN TO USE:
 
 Updates an existing root component configuration by modifying its parameters, storage mappings, name or description.
 Not for SQL transformations (`keboola.snowflake-transformation` / `keboola.google-bigquery-transformation`),
-data apps (`keboola.data-apps`) or flows — use the dedicated tools (see WHEN NOT TO USE).
+data apps (`keboola.data-apps`) or flows — use the dedicated tools (see WHEN NOT TO USE). This IS the tool for
+updating Python (`keboola.python-transformation-v2`), R (`keboola.r-transformation-v2`) and DuckDB
+(`keboola.duckdb-transformation`) transformations.
 
 This tool allows PARTIAL parameter updates - you only need to provide the fields you want to change.
 All other fields will remain unchanged.
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -185,7 +185,9 @@ EXAMPLES:
 
 Creates a root component configuration using the specified name, component ID, configuration JSON, and description.
 Not for SQL transformations (`keboola.snowflake-transformation` / `keboola.google-bigquery-transformation`),
-data apps (`keboola.data-apps`) or flows — use the dedicated tools (see WHEN NOT TO USE).
+data apps (`keboola.data-apps`) or flows — use the dedicated tools (see WHEN NOT TO USE). This IS the tool for
+Python (`keboola.python-transformation-v2`), R (`keboola.r-transformation-v2`) and DuckDB
+(`keboola.duckdb-transformation`) transformations.
 
 BEFORE CALLING - REQUIRED STEPS:
 1. Call `get_components([component_id])` to retrieve the component's `configuration_schema`.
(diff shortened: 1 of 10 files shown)


Matovidlo:
@claude review

Matovidlo:
Rebased onto the AI-3664 fix branch (`martinvasko-ai-3664-migrate-flow-integration-tests-off-legacy-orchestrator`) as an interim step to unblock Integration Tests, which were failing due to the now-disabled legacy Orchestrator flow API (see #653). This PR's diff will temporarily include #653's commits until that PR merges to `main`, at which point I'll rebase this branch onto `main` to drop the duplication.

tomasfejfar:
:eyes:

tomasfejfar (changes requested):
Doesn't do the same for update. 

Also, while it works, it's not great. DuckDB transformation is essentially a SQL transformation and should be treated as such. Limitations of our tool shouldn't dictate that it goes to code-like transformations. But I wouldn't block on that alone. The missing update is the blocker.

tomasfejfar:
Also:

> Friction summary: get_components and get_config_examples — the two tools the system prompt explicitly directs you to consult before calling create_config — return nothing usable for keboola.duckdb-transformation. The only way to actually learn the config shape was docs_query, and even that needed a second, more targeted question after the first answer resolved to the wrong (but similarly-named) component. A user following the documented discovery steps literally would hit two dead ends before finding a working path.

Matovidlo:
@tomasfejfar Fixed the blocker: \`update_config\`'s docstring now mirrors \`create_config\`'s — explicitly names Python/R/DuckDB as component IDs it handles, not just the \`update_sql_transformation\` 404 fallback message.

On the design point: agreed, and not blocking this alone as you said — added a "Design note" section to the PR description linking [AI-3662](https://linear.app/keboola/issue/AI-3662/consolidate-transformation-tools-and-move-guidance-into-project-system) explicitly as the tracked place to fix DuckDB properly (its own SQL-transformation-like tool) rather than the generic-config workaround this PR ships.

On the friction report: traced it all the way through — \`get_config_examples\`/\`get_components\` return nothing usable for this component because the underlying example database (a separate service, \`keboola/kai-bot\`) genuinely has no entry for \`keboola.duckdb-transformation\` yet, not a bug on this side. Filed keboola/kai-bot#236 (draft) adding one, and added a compact inline DuckDB example directly to \`create_config\`'s docstring as a stopgap until that lands.

Rebased onto main, version bumped, full suite + ruff + check-tools-docs green.

TOOLS.md
```diff
@@ -228,6 +230,14 @@ WHEN NOT TO USE:
 - `keboola.data-apps` → use `modify_python_js_data_app` / `modify_streamlit_data_app` / `deploy_data_app`
 - `keboola.snowflake-transformation` / `keboola.google-bigquery-transformation` → use `create_sql_transformation`
 
+DUCKDB TRANSFORMATION EXAMPLE (`component_id='keboola.duckdb-transformation'`) — `get_config_examples`
```
tomasfejfar: It doesn't feel right to put example of niche component into prompt like this. 

There is already https://github.com/keboola/kai-bot/pull/236/changes merged, no?

tomasfejfar (approved):
I still think duckdb should be sql transformation to prevent the logic tension of "do not use create_sql_transformation for creating SQL for duck db" 😁

TOOLS.md
Matovidlo: **Fixed** ✅ in commit 18ccb3472d9b4f2966a8af58811253f04bd62150

Agreed — removed the inline "DUCKDB TRANSFORMATION EXAMPLE" stopgap block from create_config's docstring and regenerated TOOLS.md. keboola/kai-bot#236 merged on 2026-08-20, so get_config_examples now returns a real example for keboola.duckdb-transformation and the inline stopgap is no longer needed.

tomasfejfar (approved):

PR #664, opened 2026-08-06
AI-3699: RFC for full-table export tool

Matovidlo:
## Description

**Linear**: [AI-3699](https://linear.app/keboola/issue/AI-3699/support-17247-improvement-to-mcp-server)

### Change Type

### Summary

RFC only — no implementation in this PR.

`query_data` is the only tool that returns table rows today, and it hard-caps results at
`MAX_ROWS = 1_000` (`src/keboola_mcp_server/tools/sql.py:27`). This blocks any automation
that needs to pull a full Storage table through the MCP server, and simply raising
`MAX_ROWS` doesn't fix it — `query_data` runs synchronously and returns rows inline,
which doesn't scale to full-table exports. Storage API already supports this via an
async export job (`POST /v2/storage/branch/{branchId}/tables/{tableId}/export-async`)
that produces a file in File Storage — the MCP server has no client method or tool
using it yet.

Adds `feature_spec/table_export/RFC.md` proposing a new `export_table` tool built on
that endpoint. It maps 1:1 to the real `ExportTableRequest2` request body (fileType,
columns, format, gzip, includeInternalTimestamp, limit, orderBy, whereFilters,
timezone, sourceBranchId), and correctly models the export as a Storage-API job
(`GET /v2/storage/jobs/{id}` via the existing `AsyncStorageClient.job_detail`) — a
different system from the Job Queue that `run_job`/`get_jobs` poll — so the tool
submits, polls, and returns in a single call, mirroring the existing
`_Workspace._wait_for_new_workspace` pattern.

The tool returns a **Storage file download URL** rather than inline content by default —
a full table dumped into LLM context is rarely what's needed, and this lets the agent
read only the parts of the file it actually wants. Timeout behavior is also spelled out:
on timeout the tool returns a typed error with the Storage job ID, and recovery is
re-calling `export_table` rather than polling a separate endpoint. A remaining open
question covers whether unrestricted full-table export needs a size/row guard.

Explicitly out of scope for the eventual implementation: changes to `query_data`/
`MAX_ROWS` (tracked separately in AI-2772), pushing the exported data anywhere on the
caller's behalf, the general "too many MCP tools" tool-management question raised
separately (candidate for its own RFC/spike), and Kai-side integration (tracked
separately, to be picked up once this RFC is settled).

Version bump is patch-only (docs/RFC change per CLAUDE.md versioning rules); no code
changes in this PR.

## Testing

### Optional testing

N/A — RFC-only change, no runtime behavior to test.

## Checklist

- [x] Self-review completed
- [ ] Unit tests added/updated (if applicable) — N/A, no code
- [ ] Integration tests added/updated (if applicable) — N/A, no code
- [x] Documentation updated (if applicable) — RFC added

### DIFF
--- a/feature_spec/table_export/RFC.md
+++ b/feature_spec/table_export/RFC.md
@@ -41,11 +41,13 @@ A new tool (working name `export_table`) must let an agent:
    The tool does not need to expose every one of these on day one, but whichever subset
    it exposes must use these exact names/shapes — no invented or guessed parameters.
 2. Have the tool wait for the export job to finish (bounded by a timeout, matching the
-   pattern already used for workspace creation — see Resolution Strategy) and return
-   the exported data to the agent. See the "How is the file delivered" open question
-   below — this is the main undecided point, not the timeout mechanics.
+   pattern already used for workspace creation — see Resolution Strategy) and return a
+   download URL for the exported file to the agent (see open question #1).
 3. Get a clear, typed error if the job fails or times out, consistent with how
-   `run_job`/`get_jobs` (`src/keboola_mcp_server/tools/jobs.py`) surface job failures.
+   `run_job`/`get_jobs` (`src/keboola_mcp_server/tools/jobs.py`) surface job failures. On
+   timeout the error includes the Storage job ID; recovery is calling the tool again
+   (the export submission is cheap to repeat), not polling a separate endpoint — see
+   open question #2.
 
 `query_data` and `MAX_ROWS` are **not** changed by this RFC — this is a new, separate
 tool for full-table retrieval, not a change to the SQL query path. (AI-2772 tracks
@@ -63,20 +65,18 @@ silently decided during implementation:
      The agent gets the data directly in context, no second fetch, no assumption about
      the agent's network access. Downside: response size is bounded by whatever the MCP
      transport/agent context can hold, so this only works up to some row/byte size —
-     ties directly into open question #3 below. To keep this tractable the tool should
-     force `fileType='csv'`, `gzip=false` when returning inline (parquet/gzip make no
-     sense to inline as text) and reserve those options for a URL-based path if one
-     exists.
+     ties directly into open question #3 below. It also unconditionally spends context on
+     the whole table even when the agent only needs part of it.
    - **Return a Storage file download URL** (from `GET /v2/storage/files/{id}` after the
      job completes) and let the agent fetch it. Cheaper on the MCP response, supports
-     `parquet`/`gzip`, but assumes the calling agent can make an outbound HTTP request to
-     wherever the file lives (S3/ABS/GCS, or Keboola's file-proxy) — not guaranteed for
-     every MCP client.
-   - **Recommendation:** default to inline content, since it works uniformly for every
-     MCP client and matches what the underlying request (AI-3699) actually needs — an
-     agent that can act on the data without a separate download step. A URL-based mode
-     could be added later behind the `fileType`/`gzip` params for callers who explicitly
-     want the cheaper path, but is not the default. Confirm in review.
+     `parquet`/`gzip`, and lets the agent read only the parts of the file it actually
+     needs instead of the whole table landing in context. Assumes the calling agent can
+     make an outbound HTTP request to wherever the file lives (S3/ABS/GCS, or Keboola's
+     file-proxy) — not guaranteed for every MCP client.
+   - **Recommendation:** default to the URL-based path. A full table dumped inline into
+     LLM context is rarely what the agent needs, and the two-step fetch is an acceptable
+     MVP tradeoff — it's straightforward to add an inline mode later behind the
+     `fileType`/`gzip` params if a client that can't fetch URLs turns out to need one.
 2. **Job polling model.** To be precise about what's actually happening: the
    `export-async` endpoint is a Storage API job (`GET /v2/storage/jobs/{id}`, via the
    existing `AsyncStorageClient.job_detail`), **not** a Job Queue job — it is a
@@ -87,6 +87,14 @@ silently decided during implementation:
    timeout — exactly the pattern `_Workspace._wait_for_new_workspace`
    (`workspace.py:767-794`) already uses for the same job system. No fire-and-poll split
    is needed or currently supported.
+   - **On timeout:** the export job keeps running in Storage after the tool call returns
+     (submitting it doesn't get cancelled just because the tool stopped waiting). The
+     tool returns a typed timeout error containing the Storage job ID, and the documented
+     recovery is for the agent to call `export_table` again with the same parameters —
+     there's no separate "resume polling this job" tool or job-ID input, since that would
+     duplicate the job-submission side-effect handling for a case (very slow exports)
+     that's expected to be rare. If this proves too costly in practice (re-submitting
+     large exports repeatedly), a resume-by-job-ID path can be added later.
 3. **Row limit removed entirely, or just raised?** Exports go through File Storage, so
    there's no inherent row cap the way `query_data` has one — but very large tables may
    still need caller-supplied `limit`/`whereFilters` to keep exports usable. Confirm
@@ -107,9 +115,7 @@ silently decided during implementation:
     part.
   - A `file_detail(file_id) -> JsonDict` wrapper for `GET /v2/storage/files/{id}`, needed
     to resolve the file referenced in the completed job's `results` into a download
-    URL/credentials — there is currently no Files API client method at all. Needed
-    regardless of delivery mode: the inline-content path still has to fetch the file
-    from wherever it landed before returning it to the agent.
+    URL/credentials — there is currently no Files API client method at all.
 - New tool in `src/keboola_mcp_server/tools/storage/tools.py` (or a new
   `tools/storage/export.py` if it grows large): `export_table`, following the
   `@tool_errors()` + `tool_errors`/`ToolAnnotations` conventions used by the other
@@ -118,13 +124,15 @@ silently decided during implementation:
 - Polling/timeout logic modeled directly on
   `_Workspace._wait_for_new_workspace` (`workspace.py:767-794`): loop on `job_detail`,
   check `status == 'success'`, bounded by a timeout, `asyncio.sleep` between polls.
-- Resolve open question #1 (inline content vs. URL) before implementing the final fetch
-  step, since it determines whether the tool needs an HTTP client call to download the
-  file server-side or just formats a URL from `file_detail`.
+- `export_table` returns a download URL (formatted from `file_detail`), not fetched file
+  content — no HTTP client call to download the file server-side is needed.
 
 ## Scope
 
-In scope:
+The following describes the scope of the **implementation PR** that follows this RFC —
+this RFC PR itself only adds this document.
+
+In scope (implementation PR):
 
 - `table_export_async` + `file_detail` client methods.
 - `export_table` tool (exact sync/async shape per the open questions above).
@@ -137,18 +145,20 @@ Out of scope:
 
 - Any change to `query_data` / `MAX_ROWS` (tracked separately in AI-2772).
 - Pushing the exported data anywhere on the caller's behalf — this tool only produces
-  the export and returns it (inline or via URL, per open question #1); what the calling
-  agent does with it afterward is out of scope.
+  the export and returns a download URL for it; what the calling agent does with it
+  afterward is out of scope.
 - General "too many tools" tool-management work raised in the same conversation — that's
   a separate concern from this specific export tool and should get its own RFC/spike if
   pursued.
+- Kai-side integration — tracked separately, to be picked up once this RFC is settled.
 
 ## Verification
 
 1. `tox` — pytest, black, flake8, check-tools-docs all exit 0.
 2. Manual end-to-end via local MCP (`.mcp.json` per project `CLAUDE.md`) against a real
-   project: export a table larger than 1,000 rows, confirm the tool returns the full
-   table content (or, if the URL path is chosen instead, that the returned
-   download URL/credentials actually retrieve it).
+   project: export a table larger than 1,000 rows, confirm the returned download
+   URL/credentials actually retrieve the full table content.
 3. Confirm behavior on export failure (bad table ID, no read access) surfaces a clear
    tool error rather than a raw job-status dict.
+4. Confirm timeout behavior: force a slow export, confirm the tool returns a typed
+   timeout error with the Storage job ID, and that re-calling `export_table` succeeds.
--- a/feature_spec/table_export/RFC.md
+++ b/feature_spec/table_export/RFC.md
@@ -1,0 +1,154 @@
+# RFC: Full-table export tool
+
+Linear: [AI-3699](https://linear.app/keboola/issue/AI-3699/support-17247-improvement-to-mcp-server)
+
+## Problem
+
+`query_data` (`src/keboola_mcp_server/tools/sql.py`) is the only tool that returns table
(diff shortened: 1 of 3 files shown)


Matovidlo:
@claude review

pyproject.toml
cjayyy, line 1: Does RFC require version change?

Matovidlo: Maybe not but it's in general guidance in claude.md, should I get rid of it?

cjayyy: 🤷 Probably leave it if it's in the guidance.

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,154 @@
...
+2. **Job polling model.** To be precise about what's actually happening: the
+   `export-async` endpoint is a Storage API job (`GET /v2/storage/jobs/{id}`, via the
+   existing `AsyncStorageClient.job_detail`), **not** a Job Queue job — it is a
+   different system from the one `run_job`/`get_jobs` poll
+   (`jobs_queue_client`/`src/keboola_mcp_server/tools/jobs.py`). There is no existing
+   generic tool for polling Storage API jobs, so this should be a single tool call that
+   internally submits the export and polls `job_detail` to completion, bounded by a
+   timeout — exactly the pattern `_Workspace._wait_for_new_workspace`
+   (`workspace.py:767-794`) already uses for the same job system. No fire-and-poll split
+   is needed or currently supported.
```
cjayyy, line 89: What happens when it times out before the file is ready? How does the agent continue to poll that job/file?

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,154 @@
...
+tool for full-table retrieval, not a change to the SQL query path. (AI-2772 tracks
+whether `query_data`'s truncation behavior itself should change; that is out of scope
+here.)
+
+## Open Design Questions (need agreement before implementation)
+
+These materially affect feasibility/shape and should be resolved in review, not
+silently decided during implementation:
+
+1. **How is the exported data delivered back to the agent?** Two options:
```
cjayyy, line 62: I don't like option 1 as it automatically pushes whole table into LLM context. With option 2, agent can fetch whole file but only read parts of it as needed. 

Also I don't think 2 step process is a problem here. IMO it is valid approach to provide file URL instead. Anyway I think it is sufficient for MVP and if it becomes an issue we can change it pretty easily, no?

cjayyy: I would ask others for opinions on this as well.

feature_spec/table_export/RFC.md
cjayyy, line 1: Does this also have Kai-side counterpart?

pyproject.toml
Matovidlo: **Already addressed** ✅

Leaving it in per the earlier thread — it's part of the general CLAUDE.md guidance, not an accidental action item for this RFC-only PR.

feature_spec/table_export/RFC.md
Matovidlo: **Fixed** ✅ in commit 54ebac7715301a7481524183bee9e2204ca388df

Agreed — switched the recommendation in open question #1 to the URL-based delivery path (default), for exactly the reason you raised: dumping the whole table into LLM context is rarely what's needed, and the agent can read only the parts of the file it wants. Inline content is now the fallback that could be added later behind fileType/gzip if a client that can't fetch URLs needs it. Updated Required Behavior #2, Resolution Strategy, and Verification to match.

feature_spec/table_export/RFC.md
Matovidlo: **Fixed** ✅ in commit 54ebac7715301a7481524183bee9e2204ca388df

Added explicit timeout/recovery behavior to open question #2 and Required Behavior #3: on timeout the tool returns a typed error containing the Storage job ID, and the documented recovery is calling export_table again (the export job keeps running server-side; resubmitting is cheap) rather than polling a separate endpoint. Also added a verification step for this. If repeated resubmission on very slow exports turns out to be costly in practice, a resume-by-job-ID path can be added later.

feature_spec/table_export/RFC.md
Matovidlo: **Question** ❓

No Kai-side counterpart planned as part of this RFC — I'll integrate it on the Kai side myself once this RFC is settled. Noted explicitly in the Scope section as tracked separately.

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,164 @@
+# RFC: Full-table export tool
+
+Linear: [AI-3699](https://linear.app/keboola/issue/AI-3699/support-17247-improvement-to-mcp-server)
+
+## Problem
+
+`query_data` (`src/keboola_mcp_server/tools/sql.py`) is the only tool that returns table
+rows today, and it hard-caps results at `MAX_ROWS = 1_000`
+(`src/keboola_mcp_server/tools/sql.py:27`). This makes it impossible to build an
+automation that pulls a full Storage table through the MCP server for any table larger
```
cjayyy, line 10: **1. Stale `MAX_ROWS` — the problem statement's central premise is 10x off.** (correctness, confirmed)

The RFC states `query_data` "hard-caps results at `MAX_ROWS = 1_000` (`src/keboola_mcp_server/tools/sql.py:27`)". That's no longer true on main: commit `04c6f75` ("fix: raise query_data max rows limit to 10k", merged via #665) changed it to `10_000`.

```
$ git show origin/main:src/keboola_mcp_server/tools/sql.py | grep -n 'MAX_ROWS = '
28:MAX_ROWS = 10_000
```

This 1,000-row threshold is the RFC's main justification, and it recurs in Verification step 2 ("export a table larger than 1,000 rows"). The 10x raise materially weakens the "impossible to build an automation" framing — the gap this new tool must close is much smaller than described. Please rebase and re-verify the motivation still holds.

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,164 @@
+# RFC: Full-table export tool
+
+Linear: [AI-3699](https://linear.app/keboola/issue/AI-3699/support-17247-improvement-to-mcp-server)
+
+## Problem
+
+`query_data` (`src/keboola_mcp_server/tools/sql.py`) is the only tool that returns table
+rows today, and it hard-caps results at `MAX_ROWS = 1_000`
```
cjayyy, line 8: **7. `MAX_CHARS` is omitted — there are two independent caps, not one.** (completeness, confirmed)

The RFC frames `MAX_ROWS` as the single blocker and argues raising it "does not fix it in general" purely on payload/scalability grounds. But `sql.py:29` also defines `MAX_CHARS = 50_000`, passed to `execute_query` alongside `max_rows` (`sql.py:313`), and the tool docstring says results truncate when *either* limit is exceeded.

On a wide table the character cap binds well before the row cap, so the practical ceiling is often far below whatever `MAX_ROWS` is set to. Omitting this understates the problem, and it leaves open question #3 framed purely in rows when byte size is the actual binding constraint — which the RFC itself gestures at ("some row/byte size").

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,164 @@
...
+  - A `file_detail(file_id) -> JsonDict` wrapper for `GET /v2/storage/files/{id}`, needed
+    to resolve the file referenced in the completed job's `results` into a download
+    URL/credentials — there is currently no Files API client method at all.
+- New tool in `src/keboola_mcp_server/tools/storage/tools.py` (or a new
+  `tools/storage/export.py` if it grows large): `export_table`, following the
+  `@tool_errors()` + `tool_errors`/`ToolAnnotations` conventions used by the other
+  storage tools in that file. A single tool call: submit → poll → fetch → return —
+  no job-ID handoff to a separate poll tool (see open question #2).
+- Polling/timeout logic modeled directly on
+  `_Workspace._wait_for_new_workspace` (`workspace.py:767-794`): loop on `job_detail`,
```
cjayyy, line 125: **2. `_Workspace._wait_for_new_workspace` does not exist.** (correctness, confirmed)

This is the one section telling the implementer how to build the polling, and the reference is wrong in three ways:

```
$ grep -rn '_wait_for_new_workspace' src/
(no results)
```

The real polling loop is **inline inside `WorkspaceManager._create_ws`** (`workspace.py:702-794`) — a different class (`WorkspaceManager`, not `_Workspace`), and not a separately callable method at all. The cited range `767-794` covers only the tail of that loop.

Same bad reference appears at lines 87 and 114. An implementer can't find this and has to reverse-engineer which loop was meant, which defeats the point of citing a canonical pattern.

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,164 @@
...
+  check `status == 'success'`, bounded by a timeout, `asyncio.sleep` between polls.
+- `export_table` returns a download URL (formatted from `file_detail`), not fetched file
```
cjayyy, line 127: **3. The cited polling loop has no error branch — copying it verbatim breaks RFC requirement #3.** (correctness, confirmed)

The RFC says to model polling on the workspace loop and "check `status == 'success'`, bounded by a timeout". The actual loop (`workspace.py:766-794`) branches on **only** those two conditions — there is no `status == 'error'` / terminal-failure branch.

Consequence: an export job that fails immediately (bad table ID, revoked read access) returns status `error` on the first poll. The copied loop reads that as "not done yet", sleeps, re-polls for the full timeout, then reports a **timeout** error.

That directly contradicts requirement #3 ("clear, typed error if the job fails or times out") and Verification step 3 ("bad table ID, no read access surfaces a clear tool error rather than a raw job-status dict"). The RFC should tell the implementer to *add* the missing error branch, not mirror the loop as-is.

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,164 @@
...
+## Resolution Strategy
+
+- `src/keboola_mcp_server/clients/storage.py`: add
+  - `table_export_async(table_id, branch_id=None, **export_params) -> JsonDict` — thin
+    POST wrapper for `branch/{branch_id}/tables/{table_id}/export-async`, taking the
+    `ExportTableRequest2` fields listed above, following the existing method style
+    (e.g. `table_detail`, `bucket_table_list`).
+  - Reuse existing `job_detail(job_id)` (`clients/storage.py:848`) to poll the resulting
+    job — it already hits the correct `jobs/{id}` Storage-API endpoint (the same one
+    `_Workspace._wait_for_new_workspace` polls), no new client method needed for that
```
cjayyy, line 114: **8. Branch-scoped export endpoint reuses a poller marked as having no branch support.** (correctness, **plausible** — not fully confirmed)

The Resolution Strategy has `table_export_async` POST to `branch/{branch_id}/tables/{table_id}/export-async`, then poll via existing `job_detail`. But `job_detail` (`clients/storage.py:848-856`) hits a bare `jobs/{job_id}` and carries an explicit marker:

```python
return cast(JsonDict, await self.get(endpoint=f'jobs/{job_id}'))  # TODO: no branch support
```

That's asymmetric with every branch-aware method nearby (`table_detail`, `bucket_detail`, `bucket_table_list`), which all thread `bid = branch_id or self._branch_id`.

If an export submitted against a non-default dev branch yields a job only resolvable within that branch's scope, the reused poller can't see it and the tool stalls to timeout.

**What I verified:** the TODO marker and the asymmetry. **What I could not verify from this repo:** whether Storage export jobs are actually branch-scoped in a way that breaks the bare `jobs/{id}` poll — that needs a check against the Storage API itself. The RFC currently asserts "no new client method needed for that part" without acknowledging the gap; please either confirm branch-scoped polling works or flag it as a risk.

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,164 @@
...
+
+The following describes the scope of the **implementation PR** that follows this RFC —
+this RFC PR itself only adds this document.
+
+In scope (implementation PR):
+
+- `table_export_async` + `file_detail` client methods.
+- `export_table` tool (exact sync/async shape per the open questions above).
+- Unit tests for the new client methods and tool (mocking the Storage API), following
+  existing patterns in `tests/tools/test_storage.py` / `tests/clients/test_storage.py`.
```
cjayyy, line 140: **6. `tests/tools/test_storage.py` does not exist.** (correctness, confirmed)

`tests/clients/test_storage.py` exists, but the tool-test path does not. Storage *tool* tests actually live at `tests/tools/storage/test_tools.py`, alongside `tests/tools/storage/test_usage.py` — mirroring the `src/keboola_mcp_server/tools/storage/` package layout.

An implementer following this creates a new top-level test file that diverges from the established package-mirroring convention.

feature_spec/table_export/RFC.md
```diff
@@ -0,0 +1,164 @@
...
+- `TOOLS.md` regen via `tox -e check-tools-docs`.
+- Version bump (minor — new tool) and `uv lock`.
+
+Out of scope:
+
+- Any change to `query_data` / `MAX_ROWS` (tracked separately in AI-2772).
```
cjayyy, line 146: **5. The out-of-scope AI-2772 change has already merged to main.** (correctness, confirmed)

This bullet, and line 52-54 ("`query_data` and `MAX_ROWS` are **not** changed by this RFC ... AI-2772 tracks whether `query_data`'s truncation behavior itself should change"), defer the limit question to a future ticket. But commit `04c6f75` on main already raised it to 10k.

So the RFC defers as an open future question a decision that has been made and shipped. Reviewers weighing open question #3 do so without knowing the adjacent limit just moved 10x — which is exactly the context needed to judge whether this separate export tool is still warranted in its proposed form.

cjayyy (changes requested):
## Code review: 8 findings

This PR adds only an RFC document plus a version bump, so these are documentation-accuracy defects rather than runtime bugs. They're still worth blocking on: an RFC's whole function is to be the spec an implementer builds against, and a wrong premise, an unfindable pattern reference, and a pattern that's actively wrong to copy all propagate straight into the implementation PR.

**The most important issue is #1 + #5 together.** PR #665 raised `query_data`'s cap 10x (`MAX_ROWS` 1_000 -> 10_000, commit 04c6f75) *while this RFC sat open* — and the RFC explicitly defers that exact change to a future ticket as out of scope. That isn't just a stale number; it undercuts the stated motivation. Open question #3 ("row limit removed entirely, or just raised?") deserves a fresh read now that the adjacent limit has already moved. Recommend rebasing onto main and re-validating the premise before the implementation PR starts.

Finding #8 is flagged as plausible rather than confirmed — see the comment for what I could and couldn't verify.

All file/line claims in the RFC were checked against the PR head (54ebac7) and current main (e5fbd72).

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "keboola-mcp-server"
-version = "1.74.7"
+version = "1.74.8"
```
cjayyy: **4. Version bumped backwards — 1.74.8 is lower than main's 1.75.4.** (correctness, confirmed)

This PR bumps `1.74.7` -> `1.74.8` (same in `uv.lock`), but main is already well past that:

```
$ git show origin/main:pyproject.toml | grep -n "^version"
7:version = "1.75.4"
```

The branch is off `4d763df`, before #604 (1.75.3) and #665 (1.75.4) landed. Merging as-is either regresses the declared package version below main or produces a conflict, depending on how the merge resolves.

Also worth noting: the RFC's own Scope section asks for a **minor** bump for the implementation PR ("Version bump (minor — new tool)"), while this doc-only PR takes a patch bump for adding a markdown file. Either way this needs rebasing onto main's 1.75.4 line.

PR #665, opened 2026-08-10
fix(sql): raise query_data max rows limit to 10k

Matovidlo:
## Description

**Linear**: N/A

### Change Type  

### Summary

Raises the `query_data` tool's `MAX_ROWS` cap from 1,000 to 10,000 so larger result sets can be returned in one call without truncation. `MAX_CHARS` (50,000) is unchanged and still truncates before 10k rows are reached for wide/large result sets.

## Testing

### Optional testing

## Checklist

- [x] Self-review completed
- [x] Unit tests added/updated (if applicable) — existing `tests/tools/test_sql.py` suite passes (61 passed)

### DIFF
--- a/TOOLS.md
+++ b/TOOLS.md
@@ -4287,7 +4287,13 @@ ERROR PREVENTION:
 * Check for division by zero using NULLIF(denominator, 0)
 * Always use the LIMIT clause in your SELECT statements when fetching data. There are hard limits imposed
   by this tool on the maximum number of rows that can be fetched and the maximum number of characters.
-  The tool will truncate the data if those limits are exceeded.
+  The tool will truncate the data if those limits are exceeded. Pick the LIMIT to match the actual need:
+  - Use a small LIMIT (<=100) for inspection, sampling, or checking distinct values.
+  - Compute aggregates (COUNT, GROUP BY, SUM, AVG, etc.) in SQL rather than pulling raw rows to aggregate
+    them yourself — a large LIMIT does not make row-level aggregation in context correct or efficient.
+  - Only use a large LIMIT when the user explicitly asked for row-level data.
+  - A truncated result is a contiguous prefix of the query's output, not a random sample: any aggregate
+    computed over truncated data is wrong, not just incomplete.
 
 DATA VALIDATION:
 * When querying columns with categorical values, use query_data tool to inspect distinct values beforehand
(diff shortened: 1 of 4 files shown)


Matovidlo:
@claude review

tomasfejfar (approved):
Approving. `MAX_CHARS` (50k) still binds first, so for wide tables the cap that actually matters is unchanged; narrow results are the ones that stop getting cut at 1k rows.

One thing for a follow-up: the docstring says "Always use the LIMIT clause" (`sql.py:296`) but nothing about magnitude, so now the agent has no reason not to reach for the whole 10k. It can't pick `MAX_ROWS` either, so the LIMIT it writes is the only lever - which makes the docstring the only place to put this. Worth spelling out: small LIMIT (≤100) for inspection and distinct-value checks, `COUNT`/`GROUP BY`/`SUM` in SQL instead of pulling raw rows to aggregate in context, big LIMIT only when the user explicitly asked for row-level data. Plus that a truncated result is a contiguous prefix (`workspace.py:385-396`), so aggregates computed over it are wrong.

Also, the old kai-assistant doesn't spill oversized tool results to disk, IIRC — only the newer agent does, so there the CSV goes straight into the context window. Can you check that before we release? 50k chars is survivable, I'd just like to know it's the binding limit everywhere.

Matovidlo:
Checked the kai-assistant disk-spill question.

Confirmed: two separate agent implementations exist. `kai-assistant-backend` (the legacy one, explicitly marked as sunsetting in code comments) uses the Vercel AI SDK directly with no disk-spill for oversized MCP tool results — a large `query_data` CSV goes straight into the LLM context, bounded only by this tool's 50k-char truncation. `kai-agent` (the newer one) runs on the Claude Agent SDK/Claude Code CLI, which does spill oversized tool output to disk and keeps only a reference in context.

So 50k chars is the real binding limit on the legacy path specifically, not just in theory. Given legacy is on its way out, I think that's an acceptable tradeoff rather than a blocker — flagging so it's a known, accepted risk rather than a surprise.

Also pushed a docstring update per your other point (`sql.py:296`): now spells out small LIMIT for inspection, SQL-side aggregation instead of pulling raw rows, big LIMIT only when explicitly needed, and a note that truncated results are a contiguous prefix so aggregates over them are wrong.

tomasfejfar (approved):

PR #674, opened 2026-08-19
feat(AI-3755): accept programmatic bearer tokens for tools/list, scoped by X-KBC-ProjectId

Matovidlo:
## Description

**Linear**: [AI-3755](https://linear.app/keboola/issue/AI-3755/accept-programmatic-bearer-tokens-for-toolslist-scoped-by-x-kbc)

### Change Type

### Summary

A one-line cache-key fix in Kai's tool-inventory cache (adding `principal.projectId` to the key)
stopped a shared cache entry from silently answering one project's `tools/list` with another
project's data. Fixing that surfaced a second, previously-masked issue: some Kai sessions carry a
Keboola **bearer session token** (`kbc_at_...`) instead of a legacy Storage API token — decided by
kbc-ui's auth strategy, not by Kai — and this server has never been able to accept one. Every call,
including `tools/list`, 401s with no path to make it work (there is no client-side exchange: kbc-ui's
`getSapiTokenString()` deliberately returns empty under a bearer session).

This PR accepts a programmatic bearer token (`kbc_at_*`/`kbc_pat_*`) together with `X-KBC-ProjectId`
and exchanges it at Connection's auth-bridge resolver for that project's legacy Storage token —
pinned to exactly one project, with none of the multi-project scoping machinery (no scope object,
no ask-first gate). It is a minimal, standalone cherry-pick out of "Part A" of the much larger,
still-unmerged [PSGO-261 branch (PR #605)](https://github.com/keboola/mcp-server/pull/605), plus one
fix from later in that branch that Part A needs to be useful end-to-end (jobs-queue/ai-service/
sync-actions wired onto the resolved token, so `get_jobs`/`run_job` work too), plus one new
robustness fix specific to this narrower scope (`tools/list` skips the exchange, since the resolver
call has up to a ~35s timeout and a client's initial capability-discovery fetch must stay fast).

Full design, scope, and what's deliberately excluded (local PKCE login, multi-project scope, OAuth
session exchange — all still only on PR #605) is in
[`feature_spec/programmatic_token_project_scope/RFC.md`](https://github.com/keboola/mcp-server/blob/martinvasko-ai-3755-accept-programmatic-bearer-tokens-for-toolslist-scoped-by-x/feature_spec/programmatic_token_project_scope/RFC.md).

Files: `clients/auth_bridge.py` (new — `is_programmatic_token()`, `StorageTokenResolver`),
`config.py` (`project_id`), `mcp.py` (exchange wiring + `/list` skip), `clients/client.py`
(jobs-queue/ai-service/sync-actions bearer wiring).

## Testing

`tox` (pytest, ruff, check-tools-docs) green — except the pre-existing, environment-specific
`test_server.py::test_json_logging` flake, which fails identically on `origin/main` and is unrelated
to this change.

Unit tests: `tests/clients/test_auth_bridge.py` (detection, resolver success/error mapping,
no-token-material-logged), `tests/test_mcp.py` (`TestProgrammaticTokenExchange`; `/list`-skip
threading through `on_request`/`create_session_state`), `tests/clients/test_client.py`
(parametrized token-selection across all six bearer-capable clients).

No dev-stack access from this environment — live verification of the resolver exchange against a
real stack is still pending.

### Optional testing

## Checklist

- [x] Self-review completed
- [ ] Integration tests added/updated (if applicable) — no dev-stack access to wire up live PAT integtests from this environment
- [x] Project version bumped according to the change type (→ 1.76.0, minor) + `uv.lock`
- [x] Documentation updated (if applicable) — RFC at `feature_spec/programmatic_token_project_scope/RFC.md`

### DIFF
--- a/feature_spec/programmatic_token_project_scope/RFC.md
+++ b/feature_spec/programmatic_token_project_scope/RFC.md
@@ -52,17 +52,27 @@ narrower scope specifically calls for:
    `5xx`/timeout/network map to `502`. No token material is ever logged.
 2. **`config.py`**: new `project_id` field, mapping `X-KBC-ProjectId` (via alias) and
    `KBC_PROJECT_ID`.
-3. **`mcp.py`**: `SessionStateMiddleware.create_session_state` detects a programmatic token and
+3. **`mcp.py`**: the token itself arrives only as `Authorization: Bearer kbc_at_...`/`kbc_pat_...`
+   — never as a Storage-token header or alias, so `Config.replace_by`'s generic header-to-field
+   mapping cannot route it into `storage_token` (caught in review: an earlier version of this
+   change read `config.storage_token` in the exchange path without ever populating it from
+   `Authorization`, so the exchange never actually triggered for the traffic this RFC exists to
+   fix). `SessionStateMiddleware.apply_request_config` now reads the `Authorization` header
+   directly, and — only when it's empty and the header value looks like a programmatic token —
+   populates `storage_token` from it (stripping the `Bearer` scheme). An explicit
+   `X-Storage-(Api-)Token` always wins; a non-programmatic `Authorization` value (e.g. this
+   server's own OAuth bearer) is never forwarded as a Storage token.
+4. **`mcp.py`**: `SessionStateMiddleware.create_session_state` detects a programmatic token and
    exchanges it before building `KeboolaClient`, using `config.project_id` to pin the exchange to
    one project. No scope object, no multi-project fan-out, no ask-first gate — a session either has
    a legacy token (used as today) or a bearer token + project id (exchanged once, used as today
    from that point on).
-4. **`clients/client.py`**: `jobs_queue`, `ai_service`, and `sync_actions` are switched onto the
+5. **`clients/client.py`**: `jobs_queue`, `ai_service`, and `sync_actions` are switched onto the
    already-existing `bearer_or_sapi_token` (storage/scheduler/data-science/metastore already use
    it) instead of the raw `self._token`. Without this, a resolved token still 401s on
    `get_jobs`/`run_job` because those three clients were wired to send the pre-exchange raw token.
    Cherry-picked from later in the PSGO-261 branch, where this exact gap was found and fixed.
-5. **New for this narrower scope — skip the exchange on `/list`.** The resolver call has up to a
+6. **New for this narrower scope — skip the exchange on `/list`.** The resolver call has up to a
    ~35s timeout (`connect=5s, read=30s`); a client's initial `tools/list` fetch must be fast, and
    this server already has a precedent for this exact trade-off (branch-id validation is likewise
    skipped for `/list`). `create_session_state` gains a `skip_token_exchange` flag, set from
@@ -72,6 +82,32 @@ narrower scope specifically calls for:
    `project_has_semantic_models` fails closed on any exception. This directly forecloses the
    `tools/list`-hangs-under-a-bad-credential failure mode this incident is about, for the new
    bearer-token path specifically.
+
+## Security review notes
+
+- **Token leakage**: `apply_request_config`'s pre-existing debug log used to dump the full inbound
+  headers mapping (`headers={http_rq.headers}`), which already covered `X-Storage-(Api-)Token`.
+  Adding `Authorization` as a second live-credential-bearing header this server actually consumes
+  widened that log line's blast radius, so it now logs header **names** only, never values.
+  `StorageTokenResolver`/`StorageTokenExchangeError` never log or place token material in exception
+  messages (existing behavior, re-verified); `Config.__repr__` already redacts any field whose name
+  contains `token`/`password`/`secret`, so `LOG.info(f'...{config}.')` in `create_session_state`
+  stays safe.
+- **SSRF / stack pinning**: `StorageTokenResolver`'s own `connection.`-prefix hostname check is a
+  weak allowlist (matches `connection.attacker.tld` too) — the same known gap as the parent RFC's
+  increment-7 finding #4, deliberately not re-fixed here (that fix is a separate, larger increment).
+  It is not independently reachable in the deployed case this feature runs in: `apply_request_config`
+  already pins `storage_api_url` back to the server's own configured stack before the resolver ever
+  sees it, whenever the server has one configured (the normal, expected shape of a deployed
+  server). A deployed server started with no stack of its own would lose that pinning — a
+  misconfiguration, not a state this change introduces.
+- **Authorization scope**: only a value that already looks like `kbc_at_`/`kbc_pat_` is ever copied
+  out of `Authorization` into `storage_token`; this server's own OAuth bearer (or any other scheme)
+  is left untouched, and an explicit `X-Storage-(Api-)Token` always takes precedence.
+- **Access control**: this server does not itself decide whether `subject_token`'s owner may access
+  `X-KBC-ProjectId` — that authorization boundary is Connection's resolver endpoint, exactly as the
+  parent RFC's contract already specifies. A caller supplying a project id it has no access to gets
+  the resolver's own 400/401/403, passed through unchanged.
 
 ## Scope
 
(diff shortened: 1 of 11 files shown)


Matovidlo:
@claude review

AdamVyborny (changes requested):
Thanks for cutting this down — the shape is right, and it closes both of the questions I was left with after reading #605.

Two things I want to call out as genuinely good before the problem:

- **The `/list` skip is the right call and does more than save time.** Taking the resolver's ~35 s window off capability discovery removes the exact failure mode the incident was heading toward: a client's first `tools/list` hanging on a credential that can never resolve. Reusing the existing branch-validation precedent for it is nicer than inventing a new flag semantics.
- **Pulling the `jobs_queue` / `ai_service` / `sync_actions` wiring forward settles a real ambiguity.** #605's body contradicted itself on those three — "Known limitations" said they'd still 401 under a resolved token, while Decision §6 and the service table said they were already bearer-first. Having the fix in the same PR means `get_jobs`/`run_job` work instead of failing one layer deeper once `tools/list` starts succeeding.

## Blocking: the token never reaches the slot the exchange reads

The RFC states the requirement as:

> A request carrying `Authorization: Bearer kbc_at_...` (or `kbc_pat_...`) **and** an `X-KBC-ProjectId` header must be accepted and resolved to that project's legacy Storage token

The implementation reads the token from `config.storage_token` (`create_session_state` → `is_programmatic_token(storage_token)`, and `_exchange_programmatic_token` passes `subject_token=config.storage_token`). But `config.storage_token` can only be populated two ways, and an inbound `Authorization` header is neither:

1. `_amend_config`'s `config.replace_by(http_rq.headers)`. `Config._read_options` matches a field only on `<name>`, `KBC_<name>` or `X-<name>`, and `_normalize` just lowercases and strips `_`/`-`. So the reachable keys are `storagetoken` / `kbcstoragetoken` / `xstoragetoken` (and `storageapitoken` / … via the alias). `Authorization` normalizes to `authorization`, which matches no field — `bearer_token` carries no alias, so it would need `X-Bearer-Token`.
2. `http_rq.scope['user']`, which requires an `AuthenticatedUser` + `ProxyAccessToken` — this server's *own* OAuth session. Kai is not an OAuth client of the MCP server, so this is never populated for its requests.

Kai sends the credential **only** in `Authorization`. Its bearer branch sets `Authorization: Bearer <token>` and `X-KBC-ProjectId`, and deliberately sends no storage-token header (the sapi branch sends `X-Storage-Token`, which is why legacy traffic works today). So for the requests this PR exists to fix, `config.storage_token` stays `None`, `is_programmatic_token(None)` is `False`, the exchange never runs, and the pre-existing "Storage API token is not provided." check fires exactly as it does now.

I verified this by reading `config.py`/`mcp.py` on `main` plus this diff — I have no dev stack, so this is a code reading, not a live repro. Happy to be shown I've misread the header path.

## Two ways to close it

**Server side (my preference, and what the RFC already promises):** read an inbound `Authorization: Bearer <token>` into the slot the exchange consumes when the value is a programmatic token. No client changes, and nothing else that talks to this server has to learn the quirk.

**Client side (one line in kai-agent):** also send the token as `X-Storage-Token` in the bearer branch. That satisfies `is_programmatic_token()`, `project_id` still comes from `X-KBC-ProjectId`, and `/list` skips the exchange as designed. This is consistent with how the code actually models things — the new `project_id` docstring says it is "only consulted when the inbound Storage token is a Keboola programmatic token", i.e. a programmatic token arriving in the storage-token slot is the expected shape. But it contradicts the RFC's prose, so if you'd rather go this way, please update the RFC to say the token must arrive in `X-Storage-Token` — otherwise the next client repeats this week.

Either is fine by me and I'm glad to ship the kai-agent one-liner immediately if that unblocks you faster. Requesting changes only so the RFC and the wiring end up agreeing, whichever direction you pick.

Matovidlo:
@AdamVyborny confirmed — verified independently and you're right, this was a real bug: `Authorization` matched no field/alias in `Config.replace_by`, so the exchange never triggered for the bearer traffic this PR exists to fix.

Went with the server-side fix (your preference, matches the RFC's stated contract): `apply_request_config` now reads `Authorization` directly and routes a `kbc_at_`/`kbc_pat_`-shaped value into `storage_token` when that slot is otherwise empty. An explicit `X-Storage-(Api-)Token` still wins, and a non-programmatic bearer (e.g. this server's own OAuth token) is never forwarded as a Storage token. RFC updated (`feature_spec/programmatic_token_project_scope/RFC.md`, item 3) with the exact mechanism and the review finding that caught it.

While re-checking the diff for token leakage (ran a full OWASP/token-exfiltration pass alongside this), found and fixed a related issue in the same function: its pre-existing debug log dumped the full inbound headers mapping, which already covered `X-Storage-(Api-)Token` and now also covers the new `Authorization` bearer. Switched to logging header names only.

New regression tests cover both: the Authorization→storage_token routing (including the explicit-header-wins and non-programmatic-bearer-ignored cases), the full Authorization + X-KBC-ProjectId → resolver chain end-to-end, and a test asserting no header value ever reaches the debug log. Full `tox` green.

Also ran ponytail + thermo-nuclear structural passes on the whole diff — no blockers, one minor `frozenset`→tuple nit applied.

AdamVyborny (approved):
Confirmed fixed — thanks for taking the server-side route.

```python
if not config.storage_token and is_programmatic_token(auth_header := http_rq.headers.get('Authorization')):
    config = dataclasses.replace(config, storage_token=strip_bearer(auth_header))
```

That's the right shape on all three counts I cared about: gated on an empty slot so an explicit `X-Storage-(Api-)Token` still wins, gated on the programmatic prefix so this server's own OAuth bearer is never forwarded downstream as a Storage token, and `is_programmatic_token` already tolerates the `Bearer ` prefix and a `None`. Traced it against Kai's actual request (`Authorization: Bearer kbc_at_…` + `X-KBC-ProjectId`, no storage-token header): `storage_token` now gets populated, `project_id` comes off the alias, non-`/list` calls exchange, and `/list` short-circuits with the raw token so the pre-existing "not provided" check passes and discovery stays fast. That closes the incident with no kai-agent change, which is the outcome I was hoping for.

Also worth saying: the logging fix is a good independent catch. The old line interpolated the whole headers mapping, so it was already dumping live `X-Storage-(Api-)Token` values at debug — pre-existing, and your change to header names only fixes more than it was strictly required to.

Approving. Three non-blocking notes, the first two verified rather than relayed from the bot.

**1. The resolver's `status_code` never shapes the response** (Copilot's suppressed comment — real). `StorageTokenExchangeError` is a `RuntimeError`, and `cli.py`'s `_exception_handlers` maps `HTTPException` specially with `Exception: _create_exception_handler(status_code=500, ...)` as the catch-all. Nothing catches `StorageTokenExchangeError` between `resolve()` and that handler, so every resolver failure lands as a 500 and the carefully-computed status is discarded — the RFC's "resolver 400/401/403 pass through unchanged; 5xx/timeout/network → 502" isn't delivered yet. This one has a concrete consequence on our side: a bearer caller passing a project id they can't access should read as a 403 policy refusal, but will arrive in Kai's ops channel as an unhandled 500 — which is the exact alert noise that started this incident. Re-raising as `starlette.exceptions.HTTPException` would do it.

**2. A 200 with a non-JSON body escapes the same way** (also real). `body = cast(dict, response.json())` sits outside the `try` that catches `httpx.HTTPError`, and `.json()` raises `json.JSONDecodeError` (a `ValueError`), so a proxy error page returned as 200 becomes an unhandled 500 instead of the intended 502. Same root as #1 and fixable in the same pass.

**3. Worth confirming before we call the incident closed, not before merge.** #605's Decision §6 removed `StorageTokenResolver` outright because live testing against a real dev stack hit a **403 on this exact resolver call**, attributed to `SCOPE_INTERNAL_AUTH_BRIDGE_RESOLVE_STORAGE_TOKEN` being a separately-provisioned Manage scope from `exchange-oauth-token`'s. This PR reinstates that call, and its own testing section notes live verification is still pending. So the code can be right and the deployment still fail closed. If that scope grant isn't in place on the target stack yet, it'd be good to know before we tell ops the errors are going away — and per #1 it would surface as another 500 rather than a legible 403.

Nothing here blocks. Happy to take the kai-agent side of anything if it's easier for you.

PR #676, opened 2026-08-20
AJDA-3161: generate a unique default draft branch instead of the literal `init`

MiroCillik:
## Description

**Linear**: [AJDA-3161](https://linear.app/keboola/issue/AJDA-3161/support-17342-draft-creation-for-existing-apps-uses-stale-branch) (SUPPORT-17342, reported by BONAMI.CZ)

### Change Type

### Summary

Creating a draft for an **existing** python-js data app could silently serve outdated code. The preview looked fine; it just wasn't built on current `main`. If the user approved it and the agent edited from that stale tree, promoting the draft could regress production. Nothing errored anywhere.

**Root cause — two halves, both required:**

1. `_DEFAULT_DRAFT_BRANCH` was the fixed literal `init` (introduced in v1.64 by AI-3286, replacing the random `iter-<6-hex>`). Every default-branch draft of a given prod app therefore asked for the same branch name. Note the asymmetry it created: the draft *slug* gets a unique `-draft-<hex>` suffix (`_derive_slug_from_name`), the branch did not.
2. Git's checkout DWIM did the rest. In a full clone where `origin/init` exists, `git checkout init` creates a local branch **tracking that stale remote tip** — it does not branch off `main`, and it does not warn. The v1.64 docstring said `git checkout <branch> (creating it from main)`; the parenthetical is advice, the DWIM is behavior. **No buggy agent code is needed to reproduce this — the tool docstring alone is enough.**

`init` survives promotion only when step 5's `git push origin --delete <branch>` was skipped or failed — an abandoned draft, a manual promote, an interrupted session. That is exactly the "apps that were previously iterated on" in the report.

**Fix:**

- `_generate_default_draft_branch()` returns `draft-<6 hex>`, unique per call, mirroring the slug suffix. The default path can no longer collide. The `draft-` prefix keeps AI-3286's readability goal — only the fixed-literal half is reverted, not the move away from `iter-`.
- A branch the **agent** supplies can still collide, and the server must not silently rename what the caller asked for. So the draft-create response now carries `_draft_checkout_hint()`: branch off `origin/main` explicitly (`git checkout -B <branch> origin/main`), never a bare `git checkout`, plus a `git rev-list --count <branch>..origin/main` behind-check for the deliberate-resume case. The three scenario docstrings now say the same thing.

**Known gap, deliberately not closed here:** the data-science API exposes only `GET /apps/{id}/git-repo` and `POST /apps/{id}/git-repo/credentials` — there is **no branch-list endpoint**, so MCP cannot detect a collision server-side and reject or auto-suffix a colliding agent-supplied branch. It can only instruct. If that endpoint lands, the check belongs on the draft create path. Documented in `docs/python-js-data-apps.md`.

A related doc fix is in flight on `keboola/ai-kit`: the `keboola-git` skill tells agents "the pre-receive hook declines branch deletes", which is false (stock Forgejo, no branch protection, no hooks, credential has collaborator `write`) — and believing it makes agents skip the delete step, which manufactures the lingering-branch precondition.

## Testing

E2E scenario (wrong → right):

| | before | after |
|---|---|---|
| First default draft on PROD | branch `init` | branch `draft-a1b2c3` |
| Second default draft on same PROD, `origin/init` still present | branch `init` → `git checkout init` resolves to stale tip → preview serves old code | branch `draft-9f4e21` → no collision possible |
| Agent-supplied branch reused across sessions | bare `git checkout <branch>` → stale tip, silent | `change_summary` instructs `git checkout -B <branch> origin/main` + behind-check |

Automated:

- `tests/tools/test_data_apps.py::test_modify_python_js_data_app_create_draft_defaults_branch_to_unique_name` — asserts two consecutive default creates return **different** branches. Verified it fails on the pre-fix code (temporarily restoring `return 'init'`) and passes on the fix.
- Extended `..._create_draft_uses_external_git` to assert the checkout hint is surfaced on the agent-supplied-branch path too.
- Integration test updated to accept the generated pattern.
- Full suite: 1744 passed. (`tests/test_server.py::test_json_logging` fails identically on unmodified `main` in my environment — pre-existing, unrelated.)
- `ruff format --check` + `ruff check` clean; `TOOLS.md` regenerated and idempotent.

## Checklist

- [x] Self-review completed
- [x] Unit tests added/updated (regression test included, per CONTRIBUTING § Bug fixes)
- [x] Integration tests added/updated
- [x] Project version bumped according to the change type (1.76.1 → 1.76.2; `uv.lock` version line synced)
- [x] Documentation updated (`docs/python-js-data-apps.md` — new "The stale-branch failure mode" section, parameter reference, flows, migration table; `TOOLS.md` regenerated)

No RFC: bug fix (CONTRIBUTING § RFC Requirement).

https://claude.ai/code/session_018r835YHKwuTTa8TYgNqewd


Matovidlo (approved):
LGTM did you test it with Kai and without Kai ?

docs/python-js-data-apps.md
```diff
@@ -108,7 +108,7 @@ Step 2: modify_python_js_data_app(
             slug='demo-draft',
             parent_configuration_id=PROD,
         )                    ──► { configuration_id: DRAFT, repo_url: R,
-                                   git_clone_url: U, branch: 'init' }
+                                   git_clone_url: U, branch: 'draft-<hex>' }
```
MiroCillik, line 111: **Stale `init` in the flow diagram** — Step 2 here (and Step 4) were updated to `draft-<hex>`, but Steps 3 and 5 further down the same code block were not:

- L115-116: `git checkout init (create if empty)` / `git push origin init`
- L128-129: `git merge init` / `git push origin --delete init`

So the diagram now contradicts itself inside one block, and — worse — the project's own reference doc still teaches the bare-`git checkout` DWIM trap that the new "stale-branch failure mode" section ~300 lines below exists to warn against. Those four occurrences should become `<branch>` / `draft-<hex>` too.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -988,14 +996,16 @@ async def modify_python_js_data_app(
...
-       → `(configuration_id=DRAFT, repo_url=R, git_clone_url=U, branch='init')`.
-       Default draft branch is `'init'`. Override with `branch=<name>` for a descriptive name.
-    3. YOU: `git clone U`; `git checkout init` (creating it if the repo is empty); write source;
-       `git push origin init`.
+       → `(configuration_id=DRAFT, repo_url=R, git_clone_url=U, branch='draft-<hex>')`.
+       The default branch is generated fresh per draft (`draft-<hex>`) so it can never collide
+       with a branch left behind by an earlier draft. Override with `branch=<name>` for a
+       descriptive name — if you do, YOU own uniqueness (see step 3).
+    3. YOU: `git clone U`; `git checkout -b <branch>` (the repo of a brand-new prod app is empty,
+       so there is no `main` to branch from yet); write source; `git push origin <branch>`.
```
MiroCillik, line 1009: **Scenario A steps 3 and 5 now contradict each other.** Step 3 asserts "the repo of a brand-new prod app is empty, so there is no `main` to branch from yet" and does `git checkout -b <branch>` — but step 5 still opens with `git checkout main`.

An agent following this docstring verbatim on a brand-new app never creates `main`, so step 5 fails with `pathspec 'main' did not match` and the approved draft cannot be promoted. `integtests/tools/test_data_apps.py:363-384` only passes because it explicitly creates and pushes `main` first — that instruction is missing from step 3.

Either add "create and push an empty `main` first" to step 3, or make step 5 use `git checkout -B main <branch>` for the brand-new-app case.

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -2107,6 +2132,46 @@ def _get_data_app_slug(name: str) -> str:
...
+    promote) outdated code with no error anywhere (AJDA-3161). ``checkout -B ... origin/main``
+    states the intended base explicitly; the ``rev-list`` check covers the case where the agent
+    deliberately resumes an existing branch.
+
+    :param draft_branch: The branch the draft was pinned to
+    :return: A hint string for the tool's `change_summary`
+    """
+    return (
+        f"Draft pinned to branch '{draft_branch}'. Base it on current production explicitly: "
+        f'`git fetch origin && git checkout -B {draft_branch} origin/main` '
```
MiroCillik: **`checkout -B <branch> origin/main` sets the draft branch's upstream to `origin/main`.** With a remote-tracking start point and the default `branch.autoSetupMerge=true`, `-B` implies `--track`. Consequences on the draft branch:

- `git status` reports "up to date with 'origin/main'" — actively misleading while iterating.
- A bare `git push` targets `main`. Under the default `push.default=simple` that's a hard error, but under `push.default=upstream` it **actually pushes unreviewed draft commits to `main`** — exactly what the absolute rule at L988 ("NEVER push directly to `main`") forbids.

Suggest `git checkout -B <branch> --no-track origin/main` here, and in the two other places carrying the same recipe: Scenario B step 3 (L1025) and the `branch` output-field description (L430).

src/keboola_mcp_server/tools/data_apps.py
```diff
@@ -1012,8 +1022,11 @@ async def modify_python_js_data_app(
             parent_configuration_id=PROD,
             branch='<describes-the-change>',   # e.g. 'add-revenue-filter'
        )` → `(DRAFT, R, U2, branch)`. Use U2 (it has its own fresh token).
-    3. YOU: `git clone U2`; `git checkout <branch>` (creating it from `main`); edit source;
-       `git push origin <branch>`.
+    3. YOU: `git clone U2`; `git fetch origin && git checkout -B <branch> origin/main` — state the
+       base explicitly. NEVER a bare `git checkout <branch>`: when that branch already exists on
+       the remote (a descriptive name reused from an earlier session), git silently checks out its
+       stale tip instead of branching from `main`, and the preview then serves outdated code with
+       no error. Edit source; `git push origin <branch>`.
```
MiroCillik: **No remedy offered for the non-fast-forward case this guidance walks into.** `git checkout -B <branch> origin/main` followed by `git push origin <branch>` is rejected as non-fast-forward whenever the remote branch holds commits that are not in `main` — i.e. an *abandoned* earlier draft whose descriptive name the agent reuses, which is precisely the collision this paragraph targets.

(The case in the ticket — `init` lingering after a *successful* merge — fast-forwards fine, since its tip is an ancestor of `main`. This is the abandoned-draft variant.)

Neither the docstring nor `_draft_checkout_hint` mentions a fix, so the likeliest agent recovery from the push rejection is the bare `git checkout <branch>` the same paragraph forbids. Worth adding: on rejection, either `git push --force-with-lease origin <branch>` or delete the stale remote branch first.

MiroCillik (commented):
Automated code review (Claude Code). The core fix is correct: `secrets` is already imported, `draft_branch` is `None` on the prod-create path so `checkout_hint` is correctly skipped, `branch` on prod create is still rejected upstream, the `'\n'.join(...) or None` empty case still collapses to `None`, and no `_DEFAULT_DRAFT_BRANCH` / `'init'` references remain in code.

4 findings below, most severe first. Note also two things judged below the reporting bar: the new `assert branches[0] != branches[1]` is a ~1-in-16.7M flake given `token_hex(3)`, and the "can never collide" / "collision-free" wording slightly overstates those 24 bits of entropy.

docs/python-js-data-apps.md
MiroCillik: Fixed in 24b41b34. All four `init` occurrences are gone from the diagram — step 3 is now `git checkout -b <branch>` / `git push origin <branch>`, and step 5 is `git checkout -b main <branch>` (with the existing-app `checkout main && merge` variant noted inline) / `git push origin --delete <branch>`. The remaining `init` mentions in this file are all historical prose describing the v1.64 behaviour that AJDA-3161 reverts, which is correct to keep.

src/keboola_mcp_server/tools/data_apps.py
MiroCillik: Fixed in 24b41b34. Step 5 no longer assumes `main` exists: it now reads `git checkout main && git merge <branch>`, followed by an explicit brand-new-app branch — "On a brand-new app (step 3 above) \`main\` does not exist yet, so create it from the approved draft instead: \`git checkout -b main <branch>\`". Both paths then converge on `git push origin main`. The same correction was applied to step 5 of the flow diagram in docs/python-js-data-apps.md.

docs/python-js-data-apps.md
```diff
@@ -413,14 +417,47 @@ Against `data-science.canary-orion.keboola.dev`:
 
 ---
 
+## The stale-branch failure mode (AJDA-3161)
```
MiroCillik: Is this whole section really necessary? Don't mention linear tasks and Support tasks

MiroCillik: Done in 55fa63a2.

**Ticket references** — removed all 11 `AJDA-3161` / `SUPPORT-17342` mentions this branch introduced, across `data_apps.py` (module comment, draft-create comment, `_draft_checkout_hint` docstring, `_generate_default_draft_branch` docstring), the unit tests, and this doc. The surrounding prose already explains the mechanism, so nothing is lost by dropping the identifiers. The pre-existing `AI-3286` / `AI-3005` references at the top of this doc and in the integtests predate the branch, so I left them alone — say the word if you want those gone too.

**The section** — I kept it but cut it down by roughly a third and retitled it "Why the draft branch is generated, not fixed". The Symptom/Cause/Fix incident framing and the v1.63/v1.64 archaeology are gone; what remains is what justifies the design and would otherwise get re-litigated: that `git checkout <branch>` silently resolves to a stale `origin/<branch>` rather than branching off `main`, when a stale branch actually lingers (promote-time delete skipped or failed), and the note that the data-science API has no branch-list endpoint so the collision cannot be caught server-side.

Happy to drop it entirely if you would rather the doc not carry it — it is a single self-contained section, so removing it is a clean one-line revert.

MiroCillik:
@Matovidlo plz, nie som si isty ci toto releasnut samostatne alebo cakat na dalsie veci?

Matovidlo:
Uz je to releasnute, takze to vali v produkci, ked tak to otestuj

MiroCillik:
nvm, uz sa to releaslo s tymto https://github.com/keboola/mcp-server/releases/tag/v1.81.1 :)

PR #453, opened 2026-04-01
feat(oauth): dynamic OAuth client registration via Connection [AI-2883]

Matovidlo:
## Description

**Linear**: [AI-2883](https://linear.app/keboola/issue/AI-2883/mcp-server-implement-dynamic-oauth-client-registration-via-connection) | Relates to: [RISK-76](https://linear.app/keboola/issue/RISK-76)

### Change Type

- [ ] Major (breaking changes, significant new features)
- [x] Minor (new features, enhancements, backward compatible)
- [ ] Patch (bug fixes, small improvements, no new features)

### Summary

Addresses the security issue RISK-76 where the MCP server bypassed Connection's OAuth client registry entirely — accepting any `client_id` without validation and using a hardcoded domain whitelist for redirect URI validation.

**What changed:**

- Removed the hardcoded `_ALLOWED_DOMAINS` / `_RE_LOCALHOST` whitelist constants
- Added `_validate_client()` method that calls Connection's `POST /oauth/clients/validate` endpoint before authorizing any client
- `authorize()` now enforces two flows:
  - **Flow A — Pre-registered clients** (Claude.ai, ChatGPT, Make.com, Cursor, n8n, etc.): redirect URI is validated against Connection's registered URI list for that specific client
  - **Flow B — Dynamic registration**: unknown clients receive `pending` status; their `client_id`, `redirect_uri`, and `client_name` are forwarded as URL params to Connection, which shows an approval screen during the OAuth login flow
- `validate_redirect_uri()` is now minimal — only rejects dangerous scripting schemes (`javascript`, `data`, `vbscript`); per-client domain validation is fully delegated to Connection
- Fails open (treats as `approved`) when Connection is unreachable, to avoid breaking existing clients during migration rollout

**Connection side (prerequisite — must be deployed first):**
- `POST /oauth/clients/validate` endpoint
- Client approval screen in OAuth login flow
- Pre-registered known clients via `php bin/console league:oauth2-server:create-client`

## Testing

- [ ] Tested with Cursor AI desktop (`Streamable-HTTP` transports)

### Optional testing
- [ ] Tested with Cursor AI desktop (all transports)
- [ ] Tested with claude.ai web and `canary-orion` MCP (`SSE` and `Streamable-HTTP`)
- [ ] Tested with In Platform Agent on `canary-orion`
- [ ] Tested with RO chat on `canary-orion`

## Checklist

- [x] Self-review completed
- [x] Unit tests added/updated (if applicable)
[...]

PR #683, opened 2026-08-25
AI-3773: fix(oauth): allow agnes.keboola.systems as a redirect URI

AdamVyborny:
## Description

**Linear**: [AI-3773](https://linear.app/keboola/issue/AI-3773/mcp-oauth-redirect-allowlist-rejects-agneskeboolasystems)

### Change Type

### Summary

Agnes' "Connect with `keboola_mcp_aws_eu`" button fails against the eu-central-1 MCP server:

```json
{"error": "invalid_request", "error_description": "Invalid redirect_uri: https://agnes.keboola.systems/"}
```

The URI reads as a bare root only because `_strip_redirect_uri()` reduces it to scheme+host+port before logging. Agnes actually sends `https://agnes.keboola.systems/api/mcp/oauth-client/callback` — and the MCP server never checks the path, only the host.

The cause is the host allowlist in `_ALLOWED_DOMAINS`, whose Keboola pattern is `^(?!(?:.*\.)?hub\.).+\.keboola\.(com|dev)$`. Agnes used to live at `agnes.canary-orion.keboola.dev`, which matched; after the move to `agnes.keboola.systems` it does not, because `.systems` is not in the TLD alternation. This adds an exact-host entry alongside the existing `librechat.glami-ml.com` / `cloud.onyx.app` / `n8n.*.groupondev.com` ones.

**Why exact host and not `*.keboola.systems`:** AI-3591 is retiring this list in favour of Connection's League OAuth2 client registry and specifically names the list's wildcard patterns as a subdomain-takeover risk. An exact host adds no new wildcard, and this entry disappears with the list when AI-3591 lands.

### E2E scenario

Before → the "Connect" button on the Agnes MCP source card bounces to the error above and the source stays `Not connected`.
After → the consent screen renders and the OAuth callback completes.

Deploying this needs a release plus an `mcp-server/tag.yaml` bump in `kbc-stacks/kbc-eu-central-1` (eu-central-1 currently runs `production-92fcccfb828b50d82e5b0a6dccd51c5fdeb7b469`), and the same for any other stack Agnes connects to.

## Testing

Not run — this change touches only the redirect-URI host allowlist, which is covered by the unit test below. The real-world verification is the Agnes connect flow against a stack running this build.

## Checklist

- [x] Self-review completed

### Regression test

Five cases added to the existing `test_validate_redirect_uri` parametrize block. The two positive cases fail on the unfixed code with exactly the production error, and pass after:

```
FAILED tests/test_oauth.py::TestSimpleOAuthProvider::test_validate_redirect_uri[uri27-True]
FAILED tests/test_oauth.py::TestSimpleOAuthProvider::test_validate_redirect_uri[uri28-True]
E   mcp.shared.auth.InvalidRedirectUriError: Invalid redirect_uri: https://agnes.keboola.systems/
```

The three negative cases pin the boundary: no subdomains, no bare `keboola.systems`, no sibling hosts.

### Local checks

`ruff format --check`, `ruff check` and `check-tools-docs` all pass. `pytest tests` is green apart from `tests/session_store/*` (24 errors) and `test_server.py::test_json_logging` — both reproduce identically on a clean `origin/main` checkout here, they need a local Postgres.

Note on `uv.lock`: only the project's own version line is updated. A full `uv lock` on my machine re-resolved ~200 unrelated packages because uv 0.7.12 can't parse `exclude-newer = "7 days"` and silently drops the cooldown, so I left the rest of the lock untouched.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,8 +76,8 @@ jobs:
           tags: |
             type=sha,format=long
             # Production: clean v* / agent-v* tags -> production stacks.
-            type=raw,value=production-${{ github.sha }},enable=${{ (startsWith(github.ref, 'refs/tags/v') || startsWith(github.ref, 'refs/tags/agent-v')) && !contains(github.ref, '-dev.') }}
-            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-dev.') }}
+            type=raw,value=production-${{ github.sha }},enable=${{ (startsWith(github.ref, 'refs/tags/v') || startsWith(github.ref, 'refs/tags/agent-v')) && !contains(github.ref, '-dev') }}
+            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-dev') }}
             # Canary: canary-orion-v* / canary-orion-agent* tags -> canary-orion stacks.
             type=raw,value=canary-orion-${{ github.sha }},enable=${{ startsWith(github.ref, 'refs/tags/canary-orion-') }}
             # Dev: dev-v* / dev-agent* tags -> testing stacks.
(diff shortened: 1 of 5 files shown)


AdamVyborny:
@Matovidlo not sure how far you are with #453 but until then...

Matovidlo (approved):

PR #685, opened 2026-08-25
fix(PSGO-280): resolve legacy Storage token for jobs-queue/AI-service/sync-actions

Matovidlo:
## Description

**Linear**: [PSGO-280](https://linear.app/keboola/issue/PSGO-280/run-job-queue-api-401s-for-oauth-session-programmatic-token-multi)

### Change Type

### Summary

`run_job` (and any jobs-queue/AI-service/sync-actions call) has been 401ing with "Invalid access
token" for deployed, project-scoped, programmatic-token (`kbc_at_`/`kbc_pat_`) OAuth-session
users — e.g. the multi-project AWS US MCP session.

Root cause, found by tracing two prior fixes against each other:
- `fix(PSGO-261): drop the auth-bridge resolve-storage-token exchange for programmatic tokens`
  (78a0bf6e) removed the only code path that minted a legacy per-project Storage token for a
  programmatic session, because the exchange endpoint hit a 403 in live testing (missing Manage
  API scope grant on that dev stack at the time).
- `fix(INC-02580): keep the Storage token on jobs-queue/AI-service/sync-actions` (d2dfbd41),
  three weeks later, routed those three clients back onto the raw session token, on the
  assumption that `create_session_state` still put an exchanged legacy token into
  `storage_token` for programmatic sessions. That assumption was stale — the exchange had
  already been removed — so the "raw Storage token" these three clients got was actually the
  raw `kbc_at_/kbc_pat_` string, which Queue's `NewJobFactory` forwards to Storage as a legacy
  `X-StorageApi-Token` and gets rejected.

Fix: re-wire the dormant `StorageTokenResolver` (already written in `clients/auth_bridge.py`,
never called) back into `SessionStateMiddleware.create_session_state` and
`MultiProjectMiddleware.client_for_project`. Once a project id is known on a deployed session,
resolve the real per-project legacy Storage token via the auth-bridge and use it — instead of
the raw token — for jobs-queue/AI-service/sync-actions, leaving the Bearer form used by
Storage/data-science/scheduler/metastore untouched. On resolver failure (e.g. that Manage-scope
grant still isn't in place on some stack), falls back to the raw token with a logged warning
instead of breaking session creation.

## Testing

Not tested end-to-end against a live deployed OAuth-session stack (would need the auth-bridge's
Manage scope grant confirmed there first, and a real multi-project programmatic session). Added
unit test coverage instead: `tests/test_mcp.py::TestLegacyTokenResolutionForQueueBackedClients`
and `tests/test_multiproject.py::TestClientForProject` cover the resolve-success, resolve-failure
(fallback), and not-deployed (no-op) cases for both `create_session_state` and
`client_for_project`. Full suite + ruff + check-tools-docs run clean (one pre-existing, unrelated
flake in `tests/test_server.py::test_json_logging`, reproduced identically on `main` before this
change).

### Optional testing

## Checklist

- [x] Self-review completed

[PSGO-280]: https://keboola.atlassian.net/browse/PSGO-280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/keboola_mcp_server/mcp.py
+++ b/src/keboola_mcp_server/mcp.py
@@ -29,7 +29,12 @@ from starlette.requests import Request
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from keboola_mcp_server.auth_login import exchange_scoped_token, get_access_token, introspect_token, load_tokens
-from keboola_mcp_server.clients.auth_bridge import is_programmatic_token, strip_bearer
+from keboola_mcp_server.clients.auth_bridge import (
+    StorageTokenExchangeError,
+    StorageTokenResolver,
+    is_programmatic_token,
+    strip_bearer,
+)
 from keboola_mcp_server.clients.base import JsonDict
 from keboola_mcp_server.clients.client import KeboolaClient
 from keboola_mcp_server.config import (
@@ -827,16 +832,37 @@ class SessionStateMiddleware(fmw.Middleware):
             bearer_token = config.bearer_token
             extra_headers: dict[str, Any] = {}
             if is_programmatic_token(storage_token):
-                # A programmatic token (kbc_at_/kbc_pat_) is forwarded downstream as a Bearer --
-                # KeboolaClient already sends it that way to every service it wraps (Storage, Queue,
-                # AI, etc.), so no legacy per-project Storage token needs to be minted for it. Strip
-                # any inbound `Bearer ` scheme so the client's own `Bearer ` prefixing can't produce
-                # `Bearer Bearer …`. Narrow to a specific project via X-KBC-ProjectId when known
-                # (header, or a prior scope selection) -- unset (whole-stack) is exactly what
-                # get_accessible_projects/set_project_scope need before a project is chosen.
+                # A programmatic token (kbc_at_/kbc_pat_) is forwarded downstream as a Bearer for
+                # Storage/data-science/scheduler/metastore. Strip any inbound `Bearer ` scheme so
+                # the client's own `Bearer ` prefixing can't produce `Bearer Bearer …`. Narrow to a
+                # specific project via X-KBC-ProjectId when known (header, or a prior scope
+                # selection) -- unset (whole-stack) is exactly what get_accessible_projects/
+                # set_project_scope need before a project is chosen.
                 bearer_token = strip_bearer(storage_token)
                 if config.project_id:
                     extra_headers['X-KBC-ProjectId'] = config.project_id
+                    # jobs-queue/AI-service/sync-actions forward whatever they receive to Storage
+                    # as a legacy X-StorageApi-Token (see client.py) -- a kbc_at_/kbc_pat_ token
+                    # 401s there (INC-02580's fix was correct for that; it just can't help this
+                    # token type). Resolve the real per-project Storage token via the auth-bridge
+                    # for those three clients. Only possible on the deployed server (needs the SA
+                    # JWT to call the resolver); on failure (e.g. the resolver's Manage scope isn't
+                    # granted on this stack yet, see 78a0bf6e) keep the raw token so this degrades
+                    # to the pre-existing 401 instead of breaking session creation outright.
+                    if kubernetes_token_path := deployed_sa_token_path():
+                        try:
+                            storage_token = await StorageTokenResolver(
+                                storage_api_url=config.storage_api_url,
+                                kubernetes_token_path=kubernetes_token_path,
+                            ).resolve(subject_token=bearer_token, project_id=int(config.project_id))
+                        except (StorageTokenExchangeError, OSError, ValueError) as e:
+                            # OSError/ValueError: the projected SA JWT file is missing/empty (should
+                            # never happen when KBC_KUBERNETES_TOKEN_PATH is set on a real deployment,
+                            # but must not crash session creation if it does).
+                            LOG.warning(
+                                'Could not resolve a legacy Storage token for jobs-queue/AI-service/'
+                                f'sync-actions calls; those calls will keep 401ing: {e}'
+                            )
 
             client = await KeboolaClient(
                 storage_api_url=config.storage_api_url,
--- a/src/keboola_mcp_server/mcp.py
+++ b/src/keboola_mcp_server/mcp.py
@@ -29,12 +29,7 @@ from starlette.requests import Request
 from starlette.types import ASGIApp, Receive, Scope, Send
 
 from keboola_mcp_server.auth_login import exchange_scoped_token, get_access_token, introspect_token, load_tokens
-from keboola_mcp_server.clients.auth_bridge import (
-    StorageTokenExchangeError,
-    StorageTokenResolver,
-    is_programmatic_token,
-    strip_bearer,
-)
+from keboola_mcp_server.clients.auth_bridge import StorageTokenResolver, is_programmatic_token, strip_bearer
 from keboola_mcp_server.clients.base import JsonDict
 from keboola_mcp_server.clients.client import KeboolaClient
 from keboola_mcp_server.config import (
@@ -128,6 +123,38 @@ class ServerState:
     runtime_info: ServerRuntimeInfo
     session_store: SessionStore | None = None
     kai_scope_store: KaiScopeStore | None = None
+    _token_resolvers: dict[str, StorageTokenResolver] = dataclasses.field(
+        default_factory=dict, compare=False, repr=False
+    )
+
+    def storage_token_resolver(self, storage_api_url: str | None) -> StorageTokenResolver | None:
+        """The auth-bridge resolver for `storage_api_url`, or None when this process is not the
+        deployed server.
+
+        One instance per stack URL, reused for the life of the server -- each holds its own pooled
+        `httpx.AsyncClient` (see `StorageTokenResolver`), so a fanned-out multi-project call, and
+        every request after it, reuse one connection instead of a TLS handshake per resolve. This
+        is also the single place that answers "am I the deployed server?" for token exchange --
+        `KeboolaClient.create` just uses the resolver it is handed, or skips the exchange when
+        handed None.
+        """
+        path = deployed_sa_token_path()
+        if not path or not storage_api_url:
+            return None
+        # `path` is not part of the cache key: it comes from the process environment only
+        # (KBC_KUBERNETES_TOKEN_PATH), so it is a constant for the life of this ServerState.
+        if (resolver := self._token_resolvers.get(storage_api_url)) is None:
+            resolver = self._token_resolvers[storage_api_url] = StorageTokenResolver(
+                storage_api_url=storage_api_url, kubernetes_token_path=path
+            )
+        return resolver
+
+    async def aclose(self) -> None:
+        """Closes every pooled connection held by `storage_token_resolver`. Call once, from the
+        server's lifespan teardown."""
+        for resolver in self._token_resolvers.values():
+            await resolver.aclose()
+        self._token_resolvers.clear()
 
     @property
     def own_stack_storage_api_url(self) -> str | None:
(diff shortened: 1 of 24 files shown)


src/keboola_mcp_server/multiproject.py
tomasfejfar: What changed: client.token used to be a key that worked for the whole stack. This PR makes it a key for one specific project.

What broke: Two lines still treat it as the whole-stack key:

```
# multiproject.py:139 and :236
base_token = scope.scoped_token or (original_client.token if is_real_client else '')
```

That's .token (now project-bound) where it should be .bearer_token (still stack-wide).

What happens:

1. Session sets up for project A → client.token = A's key.
2. User asks for something in project B.
3. That code grabs A's key as the "base" key.
4. client_for_project tries to swap it for B's key — but A's key is the wrong type of key to hand the resolver, so the resolver says no.
5. The failure is swallowed with a log warning, and A's key is kept.
6. The client for B is built carrying A's key, with a header that says "project B".

Why the header doesn't save it: jobs-queue, AI-service and sync-actions decide which project you're in by looking at the key, not the header. Header says B, key says A. The key wins.

The damage: ask to run a job in B, it runs in A. Fan out a read over A and B, you get A's data twice — the second copy labelled B. Both silent, just a log warning.

How often: every Kai multi-project session on a working stack. Not an edge case.

Not a security hole: A is always a project the user already confirmed and can already read and write. Nothing crosses a tenant or scope boundary. It's the wrong one of two projects you already had access to — a correctness bug.

The fix: use bearer_token instead of token at those two lines, and/or make client_for_project check the token type before calling the resolver (like mcp.py:834 already does) and fail instead of quietly reusing the old project's key.

Root cause: the resolve logic was copy-pasted into two files. One copy got the type check, the other didn't. Merging them into one place makes this class of bug impossible.

One unknown: if Connection's resolver actually accepts the legacy key at step 4, the swap succeeds and there's no bug. Worth confirming server-side before deciding how urgent this is.

src/keboola_mcp_server/multiproject.py
```diff
@@ -324,9 +324,25 @@ async def client_for_project(
...
         # so a pre-prefixed value would otherwise become `Authorization: Bearer Bearer …`.
         token = strip_bearer(token)
+        # jobs-queue/AI-service/sync-actions reject this programmatic token the same way the
+        # primary session client's did before it was resolved in `create_session_state` (INC-02580 /
+        # PSGO-261) -- resolve the legacy per-project Storage token here too, so a fanned-out
+        # read/write against a non-active project doesn't 401 on those three clients either.
+        storage_api_token = token
+        if kubernetes_token_path := deployed_sa_token_path():
+            try:
+                storage_api_token = await StorageTokenResolver(
```
tomasfejfar: What it is: the scope_token is an encrypted blob that remembers which projects the user confirmed. To stop someone replaying a stolen one, it's locked to the caller's own token — the token gets hashed into the encryption's "AAD" (extra authenticated data). Lock it with one value, and you can only open it with the same value.

What broke: the lock and the key are now two different strings.

- Locking (minting) uses client.token — tools/project.py:537 and :635
- Unlocking (verifying) uses config.storage_token — mcp.py:535

Those used to be the same string. This PR makes client.token the resolved legacy token, but never writes it back into config — create_session_state assigns the resolved value to a local variable only. So config.storage_token stays the original programmatic token.

Result: you lock with the legacy token and try to open with the programmatic token. It won't open. The scope_token fails to decrypt on the very next call, and the confirmed scope is silently thrown away — logged as "Ignoring invalid or expired scope_token".

When: exactly on deployed programmatic sessions with a project id — because resolve_scope_binding_aad returns None (no locking at all) unless the server is deployed (scope.py:82), and those deployed-with-project-id sessions are the ones where the resolve fires. The two conditions line up perfectly, so it's not intermittent.

Damage: the user confirms a project scope, and it evaporates on the next request — they get asked again, or fall back to no scope. Annoying, not dangerous: it fails closed, so nothing unauthorized gets through. There's also a smaller conceptual problem — the replay lock is now bound to a server-minted token rather than the caller's own credential, which is not what the lock was designed to prove.

The fix: pick one value and use it on both sides. Binding both to config.storage_token (or equivalently client.bearer_token) keeps the original intent — the lock proves the caller presented that credential.

Relationship to R1: same root cause. client.token quietly changed meaning, and every existing reader of it inherited the new meaning without anyone updating them. R1 is one such reader, R2 is another. Neither shows up in a typecheck or the existing tests, because the type never changed — only what the string means.

tomasfejfar (commented):
Duplicated logic, leaked layer boundary. The programmatic-token check, SA-JWT check, resolve, three-exception catch, and raw-token fallback are duplicated near-verbatim in mcp.py and multiproject.py. This leaks token-exchange details into HTTP middleware. Recommendation: collapse into a single KeboolaClient.create() async factory that owns credential resolution. The first comment is direct consequence of this. 

Nesting. create_session_state is now four levels deep (if programmatic → if project_id → if deployed → try/except) with long narrative comments compensating for the structure. Delegating to the factory flattens both call sites to a handful of lines. 

implicit two-credential contract. KeboolaClient now receives two credentials whose service split is documented only in comments. Model it explicitly: rename to bearer_token (Storage/data-science/scheduler/metastore) and legacy_storage_token (jobs-queue/AI-service/sync-actions).

src/keboola_mcp_server/multiproject.py
Matovidlo: **Fixed** ✅ in commit b7655637

Confirmed and fixed: the two fan-out fallback sites now use `original_client.bearer_token` (the un-narrowed whole-stack subject token) instead of `.token`. Added `TestClientForProject`-style coverage (`tests/test_multiproject.py::test_fan_out_uses_bearer_token_not_active_project_resolved_token`) asserting the fallback picks `.bearer_token` and never the project-narrowed legacy token.

src/keboola_mcp_server/multiproject.py
Matovidlo: **Fixed** ✅ in commit 8ba11c44

Confirmed — and it's worse than the seal/unseal mismatch alone: this whole class of bug (this one plus your R1 above, plus a third one I found while chasing it — `on_list_tools`'s programmatic-token guard was also reading the renamed property and would have stopped firing post-resolution) all came from the same root cause, which is exactly what your review called out. Fixed by sealing with `client.bearer_token` (matches what `_read_scope_from_request` unseals with, `config.storage_token`) instead of the resolved per-project token, and by normalizing a `Bearer `-prefixed inbound token inside `resolve_scope_binding_aad` itself so the two sides can't drift on the scheme alone. Added a seal-through-`set_project_scope` / unseal-through-`_read_scope_from_request` round-trip test — the existing tests asserted each side in isolation, which is exactly why this broke unnoticed.

src/keboola_mcp_server/multiproject.py
```diff
@@ -324,9 +324,25 @@ async def client_for_project(
...
+                    storage_api_url=storage_api_url,
+                    kubernetes_token_path=kubernetes_token_path,
+                ).resolve(subject_token=token, project_id=project_id)
+            except (StorageTokenExchangeError, OSError, ValueError) as e:
+                LOG.warning(
+                    'Could not resolve a legacy Storage token for jobs-queue/AI-service/'
+                    f'sync-actions calls on project {project_id}; those calls will keep 401ing: {e}'
+                )
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit df45af69

Addressed — the resolve-and-fallback logic this warning lived in moved into `KeboolaClient.create()` (see the top-level reply to Tomas's review for the full rewrite), and the single remaining warning site now logs with `exc_info=True`.

src/keboola_mcp_server/mcp.py
```diff
@@ -827,16 +832,37 @@ async def create_session_state(
...
+                                kubernetes_token_path=kubernetes_token_path,
+                            ).resolve(subject_token=bearer_token, project_id=int(config.project_id))
+                        except (StorageTokenExchangeError, OSError, ValueError) as e:
+                            # OSError/ValueError: the projected SA JWT file is missing/empty (should
+                            # never happen when KBC_KUBERNETES_TOKEN_PATH is set on a real deployment,
+                            # but must not crash session creation if it does).
+                            LOG.warning(
+                                'Could not resolve a legacy Storage token for jobs-queue/AI-service/'
+                                f'sync-actions calls; those calls will keep 401ing: {e}'
+                            )
```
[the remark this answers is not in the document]
Matovidlo: **Fixed** ✅ in commit df45af69

Same fix as the multiproject.py copy of this warning — both collapsed into `KeboolaClient.create()`'s single warning, now with `exc_info=True`.

Matovidlo:
@tomasfejfar — addressed all three points, in `df45af69`:

**Duplicated logic / leaked layer boundary.** Added `KeboolaClient.create()` (`clients/client.py`) as the single classmethod that classifies a caller's raw credential and performs the auth-bridge exchange. `create_session_state` (mcp.py) and `client_for_project` (multiproject.py) both now collapse to a single `await KeboolaClient.create(...)` call each — the resolve/catch/fallback logic exists exactly once.

**Nesting.** `create_session_state` is down to one `if` for the header/project_id decision; the four-level nest you flagged moved into `create()`, which is one `if` + one `try` (programmatic-token check → resolve → catch).

**Implicit two-credential contract.** Renamed `storage_api_token`/`.token` to `legacy_storage_token`/`.legacy_storage_token` throughout, so the split with `.bearer_token` is explicit. Left `Config.storage_token`'s wire aliases (`storage_api_token` / `KBC_STORAGE_API_TOKEN` / `X-Storage-Api-Token` header) untouched — that's the wire protocol, a different thing from this internal name.

One thing this surfaced beyond your three comments: the rename made visible that three existing readers of the old `.token` were silently relying on it meaning "the caller's own credential" — which stopped being true the moment `create_session_state` started swapping it for a resolved per-project token. Two of those are the ones you already caught inline (replied there with the fixes + regression tests). The third was `on_list_tools`'s programmatic-token guard (`mcp.py`), which would have stopped firing post-resolution and re-enabled the `verify_token()` round-trip that `/list` exists specifically to skip — fixed the same way, checking `.bearer_token` instead.

Also gated the resolve out of `/list` requests entirely and gave the resolver a pooled, reused `httpx.AsyncClient` (per-stack-URL, cached on `ServerState`, closed on server shutdown) — a multi-project fan-out was paying a TLS handshake per resolve before this.

Ran a structural review, an over-engineering pass, and an adversarial correctness pass (3-lens plan + evidence-grounded review + verification) against the diff; nothing else came back.

tomasfejfar (approved):
Security review (two passes, scoped to `git diff origin/main...HEAD`, reviewed at d44430cf):

**Pass 1 (initial revision 45b262be):** two candidates were identified and both were adversarially filtered out as non-security; two real functional regressions were confirmed:
- cross-project confused-deputy in multi-project dispatch (`base_token` reused the active project's resolved legacy token for other targets)
- `scope_token` AAD sealed with `client.token` but unsealed with `config.storage_token`

**Pass 2 (rework 9e2445a9):** both regressions verified fixed (b7655637, 8ba11c44), and the `KeboolaClient.create()` consolidation (df45af69) was re-reviewed with no findings ≥ threshold. Ruled out explicitly: credential-slot confusion in `create()`, header-controlled `project_id` reaching the resolver (Connection-side authoritative), SA JWT egress to caller-influenced URLs (pinning in `apply_request_config` + `normalize_storage_api_url` fullmatch before the JWT is read), credential bleed in the pooled resolver client (per-request headers), token material in `exc_info=True` tracebacks (chain-free raises), and stale readers of the renamed `.legacy_storage_token`.

Two residual fail-closed edges, non-blocking: a legacy-token session with `bearer_token=None` in multi-project dispatch now errors instead of proceeding, and a deployed non-persisted legacy-token scope would seal AAD as `None` — both look unreachable in practice.

Delta 9e2445a9..d44430cf is a two-line integtest rename fix, no security relevance.

PR #687, opened 2026-08-26
fix(PSGO-280): auto-confirm OAuth project scope and guide toward it on a real 401

Matovidlo:
## Description

**Linear**: [PSGO-280](https://linear.app/keboola/issue/PSGO-280)

### Change Type

### Summary

Users authorizing via the league `/oauth/consent` screen can now pick "all projects" or freeze
access to a specific subset (checkboxes per org/project) before this server ever sees the
resulting token. That click *is* the scoping decision — but `_auto_confirm_single_project_scope`
only ever auto-confirmed a session when introspection returned exactly one project, on the old
assumption that this server's OAuth grant was always whole-stack. Every other session (including a
deliberate 2-of-N pick, e.g. `demo1` + `demo2`) was left `confirmed=False`, so the ask-first gate
in `multiproject.py` blocked every data tool call (`list configurations`, etc.) until the agent
called `set_project_scope` again for a choice the user had already made once at login.

- `oauth.py`: renamed `_auto_confirm_single_project_scope` → `_auto_confirm_project_scope`;
  drops the "exactly one project" restriction and auto-confirms the scope to whatever the
  freshly-exchanged token can reach, at any count — mirroring `set_project_scope`'s own
  "omit `project_ids` → introspect and scope to everything returned" behavior. Still best-effort
  (introspection/exchange failures just leave the session unconfirmed, same as before).
- `clients/base.py`: `RawKeboolaClient._raise_for_status` now appends a hint to any real
  `401 Unauthorized` it raises, pointing the agent at `get_accessible_projects`/
  `set_project_scope` — a stale/unset scope is a common cause of a 401 here and shouldn't be
  mistaken by the agent for an invalid/expired credential (reproduced live: an agent hit exactly
  this and told the user to "reconnect the connector" instead).

**Follow-up, found while testing the above against a real dev-stack build:** once a session
actually reaches multi-project mode, the first `get_configs` call arrived with
`component_types=""`/`component_ids=""` (some tool-calling clients send `""` instead of
omitting/`[]` an unset array-typed argument) — pydantic-core always rejects a bare string for a
`Sequence[...]` field, and `MultiProjectMiddleware`'s fan-out loop then duplicated that same
validation failure once per scoped project into a confusing "The tool failed for all N scoped
project(s)" aggregate instead of one clean error.

- `errors.py`: `ValidationErrorMiddleware` now coerces an empty-string argument to `[]` before
  `call_next` whenever the tool's own schema declares that parameter as an `array` — a bare string
  is never a legitimate value there, so this can't clobber real input. Fixes the failure itself
  rather than just prettifying its message.
- `multiproject.py`: the fan-out loop's "abort and surface once, don't duplicate per project"
  special case only matched a bare `FastMCPValidationError`/`PydanticValidationError`. In practice
  `ValidationErrorMiddleware` (registered inner of `MultiProjectMiddleware`) already converts that
  into a `ToolError` before it reaches this except clause, so the special case never actually fired
  for this exact scenario. Now also recognizes a `ToolError` whose `__cause__` is a
  `FastMCPValidationError`.
- `feature_spec/pat_token_support/RFC.md`: documents all of the above as increments 16–17.

## Testing

Not yet manually tested end-to-end against a real OAuth consent flow from this environment. The
OAuth auto-confirm + 401-guidance fix (commit `a56a5e42`) was verified via a `dev-v1.78.4-dev.1`
tag build deployed to a testing stack, where the reported multi-project OAuth session did reach
`get_configs` — which is what surfaced the `component_types=""`/`component_ids=""` follow-up bug
fixed in commit `a1783d44` (not separately re-verified against a fresh dev build; covered by unit
tests only). Full `tox` (pytest 2112 passed / 24 skipped, ruff, check-tools-docs) is green;
`tests/test_server.py::test_json_logging` fails the same way on `main` (pre-existing, unrelated
subprocess/env flake).

### Optional testing

## Checklist

- [x] Self-review completed

[PSGO-280]: https://keboola.atlassian.net/browse/PSGO-280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature_spec/pat_token_support/RFC.md
+++ b/feature_spec/pat_token_support/RFC.md
@@ -1154,3 +1154,47 @@ against both a minimal repro and FastMCP's actual app-construction shape):
 **Explicitly not fixed here:** the deployed `KeboolaClient`'s own request path has nothing to do
 with Postgres (Storage API calls don't touch this server's session database), so it needs no
 equivalent guard.
+
+## Extension: multi-project OAuth sessions never auto-confirmed (increment 16)
+
+Increment 8's premise -- "this server's OAuth grant is always whole-stack (`claudai projectless`
+scope), so introspection's count there is just the user's real total org membership, not evidence
+of a prior scoping choice" -- predates the league's `/oauth/consent` screen letting the user
+themselves pick "all projects" or freeze access to a specific subset (checkboxes per org/project,
+screenshot reviewed in the reporting conversation). That screen *is* the scoping decision; by the
+time `exchange_authorization_code` runs, the user has already made it, whether they kept "All
+projects" or picked a subset. Auto-confirming only when introspection returned exactly one project
+left every other session -- including a deliberate 2-of-N pick -- stuck at `confirmed=False`, so
+the ask-first gate in `multiproject.py` 401'd every data-tool call in `main` until the agent called
+`set_project_scope` again for a choice the user had already made once.
+
+**Fix:** `oauth.py`'s `_auto_confirm_single_project_scope` (renamed
+`_auto_confirm_project_scope`) drops the `len(introspection.projects) != 1` guard and confirms the
+scope to whatever the freshly-exchanged token can reach, at any count -- mirroring
+`set_project_scope`'s own "omit/null project_ids -> introspect and scope to everything returned"
+behavior in `tools/project.py`. Still best-effort and still a no-op on zero reachable projects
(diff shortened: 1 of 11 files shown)


tomasfejfar (approved):
LGTM

PR #688, opened 2026-08-26
chore: speed up integtest CI (single Python version + micro-fixes)

Matovidlo:
## Description

**Linear**: N/A — CI/test-infra cleanup, no Linear issue

### Change Type

### Summary

First, smaller step toward cutting the ~16-minute integtest CI job — bigger
parallelization work (matrix concurrency, pool sizing) follows in a separate
PR once this lands.

- **Drop the integtest job's Python matrix to 3.12 only.** It previously ran
  the full suite three times sequentially (`max-parallel: 1` across
  3.10/3.11/3.12), which is the actual source of the 16 minutes. Real Storage
  API behavior isn't Python-version-sensitive; the unit-test matrix in the
  `build` job already exercises all three interpreters. This alone removes
  ~2/3 of the job's wall-clock time.
- **Poll instead of sleep** for the MCP server subprocess to start listening
  in `test_mcp_server.py` (was a fixed 5s sleep, hit 3x per run) — now polls
  the port with a 10s timeout, typically resolving in well under a second.
- **Move `test_project_lock.py` into `tests/`.** It's a fully-mocked
  unit-test suite for the locking logic (no real Storage API calls, no
  `INTEGTEST_*` secrets needed) — it belongs with the fast, already-parallel
  unit-test matrix, not the real-API integtest suite. Added
  `pythonpath = ["."]` to pytest config so it can still import
  `integtests.project_lock` under `--import-mode=importlib` from its new
  location.

Verified the locking mechanism itself (`integtests/project_lock.py`) is
already safe for concurrent CI runs from different branches — locks are
keyed off the Keboola project's own branch metadata (identified by the pool
token), not the git ref/workflow run, so this needed no change.

## Testing

### Optional testing

## Checklist

- [x] Self-review completed
- [x] Unit tests added/updated (if applicable) — `tests/test_project_lock.py` moved and passes (38 passed)
- [ ] Integration tests added/updated (if applicable) — no test content changed, only CI matrix/sleep/location; ran `pytest integtests/ --collect-only` to confirm collection is unaffected (125 tests)
- [x] Project version bumped according to the change type — 1.78.3 → 1.78.4 (patch)
- [ ] Documentation updated (if applicable) — N/A


integtests/test_mcp_server.py
```diff
@@ -253,6 +254,18 @@ async def _assert_get_component_details_tool_call(client: Client, config: Config
     assert component_config.configuration_rows is None
 
 
+def _wait_for_port(port: int, timeout: float = 10.0) -> None:
```
cjayyy: **A crashed subprocess is reported as a port timeout.**

`_wait_for_port` polls only the socket and never checks `p.poll()`, so if the server process dies during startup the loop still burns the full 10s and raises `MCP server did not start listening on port N within 10.0s`.

That message points at startup latency when the real cause was a config/import failure. Reproduced with an already-dead subprocess: 10.1s wait, misleading timeout, while the true cause (`BOOM: config error`) sat in the subprocess stderr.

The `finally` block does still log stderr, so the cause is recoverable from the log — but the raised error actively misdirects, and the old `sleep(5)` failed twice as fast. Passing `p` in and bailing out early would fail fast with the real reason:

```python
def _wait_for_port(port: int, p: subprocess.Popen, timeout: float = 10.0) -> None:
    deadline = time.monotonic() + timeout
    while time.monotonic() < deadline:
        if p.poll() is not None:
            _, stderr = p.communicate()
            raise RuntimeError(f'MCP server exited with {p.returncode} before listening:\n{stderr}')
        try:
            with socket.create_connection(('127.0.0.1', port), timeout=0.5):
                return
        except OSError:
            time.sleep(0.1)
    raise TimeoutError(f'MCP server did not start listening on port {port} within {timeout}s')
```

integtests/test_mcp_server.py
```diff
@@ -253,6 +254,18 @@ async def _assert_get_component_details_tool_call(client: Client, config: Config
...
+    """Poll until something is listening on 127.0.0.1:port, or raise TimeoutError."""
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        try:
+            with socket.create_connection(('127.0.0.1', port), timeout=0.5):
```
cjayyy, line 265: **A pre-existing listener on the random port is a false positive.** (low severity — pre-existing, not introduced here)

Any successful connect counts as "our server is up". If `random.randint(8000, 9000)` picks a port already bound by something else — a leaked server from an earlier run, or a concurrent job on the same runner — this returns in ~1ms and the test then drives its MCP client against a foreign service, failing with a confusing protocol/HTTP error instead of a clear startup error. Confirmed by binding a dummy listener: returns `CONNECTED` in 0.001s with no real server.

The old `sleep(5)` was equally blind to this, so it's not a regression — but this new helper is the natural place to close it, e.g. bind-probe the port before launching the subprocess, or handshake `/mcp` rather than accepting a bare TCP connect. Worth folding in only if you're touching this function for the first comment anyway.

cjayyy (commented):
Two bugs in `_wait_for_port`. The CI matrix change and the `test_project_lock.py` move both look correct — I verified `pythonpath = ["."]` works even under tox's `package = "wheel"` / `--no-editable` env (38 passed), that the repo root doesn't shadow the installed package (code is under `src/`), and that collection is unaffected (2175 unit / 125 integtests).

integtests/test_mcp_server.py
Matovidlo: **Fixed** ✅ in commit e2aa5177

_wait_for_port now checks p.poll() each iteration and raises immediately with the subprocess's actual stderr if it died, instead of burning the full timeout. Used your suggested implementation almost verbatim.

integtests/test_mcp_server.py
Matovidlo: **Won't fix** ℹ️

Agreed this is a real gap, but leaving it as-is for now — the fix (bind-probing before subprocess launch, or handshaking `/mcp` instead of a bare TCP connect) is more invasive than this PR's scope, and as you noted the old `sleep(5)` had the same blind spot, so it's not a regression. Happy to file a follow-up if you'd rather track it explicitly.

cjayyy:
### `tests/test_project_lock.py:837` — real production request in the moved test

**`test_pool_timeout_raises` makes a real request to production `connection.keboola.com` — flaky ~40% of runs.** Confirmed the `_wait_for_port` fix in e2aa5177 is good; found this while re-verifying.

This test is the one in the file with no HTTP mock. `_make_pool()` → `_make_endpoint()` defaults to `https://connection.keboola.com` with token `'test-token'`, and `ProjectPool.acquire()` re-checks its deadline at the *top of the loop*, so with `max_wait_minutes=0` whether it raises `TimeoutError` before reaching `self._make_lock(endpoint)._try_acquire_once()` is a `datetime.now()` race — not a guarantee.

When it loses that race it issues a real request. Stress-running the file locally, 5/12 runs failed (failing runs 0.6s vs 0.03s):

```
INFO integtests.project_lock: [project_lock] Writing candidate lock 8a60d89f-... (runner: <host>/39903)
INFO httpx: HTTP Request: POST https://connection.keboola.com/v2/storage/branch/default/metadata "HTTP/1.1 401 Unauthorized"
E   httpx.HTTPStatusError: Client error '401 Unauthorized' for url
    'https://connection.keboola.com/v2/storage/branch/default/metadata'
```

So it attempts a lock-metadata **write** against production, and only the invalid token stops it.

Why this matters more after the move: in `integtests/` this ran in a push-only, same-repo-gated job. In `tests/` it's in the unit matrix that runs on **every push and every fork PR across 3 interpreters** — so CI now sends unauthenticated writes to production Keboola from forks, and the fast suite carries a network-dependent flake. It also contradicts the file's own docstring ("All HTTP calls are mocked via pytest-mock — no real Storage API or Keboola project is ever touched").

One mock closes it — 12/12 green afterwards, and an `httpx.Client.request` spy over the whole file reports zero outbound attempts:

```python
def test_pool_timeout_raises(mocker):
    pool = _make_pool(max_wait_minutes=0)
    mocker.patch('time.sleep')
    mocker.patch.object(pool, '_make_lock', return_value=mocker.MagicMock(**{'_try_acquire_once.return_value': None}))
```

Worth pointing `_make_endpoint`'s default URL at something non-routable (e.g. `https://connection.invalid`) as defence in depth, so a future unmocked test fails fast instead of reaching production. The move itself still looks right — the suite just isn't as hermetic as advertised.

<sub>Posting at PR level: `tests/test_project_lock.py` is a pure rename with no diff hunk, so GitHub can't anchor an inline comment to line 837.</sub>

Matovidlo:
**Fixed** ✅ in commit c9f7ed22

Mocked `_make_lock` in `test_pool_timeout_raises` so a lost deadline race can't reach `connection.keboola.com` — used your snippet almost verbatim. Also pointed `_make_endpoint`'s default URL at `https://connection.invalid` as the defense-in-depth you suggested, so any future unmocked test in the file fails fast instead of reaching production.

Verified: 20/20 stress-runs of the test pass (previously \~40% failure rate), full `tests/` suite is green (2151 passed, 24 skipped), ruff clean.

cjayyy (approved):

PR #693, opened 2026-09-01
AI-3790: surface semantic-object scope/visibility metadata in read tools

Matovidlo:
## Description

**Linear**: [AI-3790](https://linear.app/keboola/issue/AI-3790/verify-that-the-new-semantic-models-and-datasets)

### Change Type

### Summary

`get_semantic_context`, `search_semantic_context` and `validate_semantic_query` dropped `scope`,
`sourceProjectId`, `targetProjectIds` and `scopeElevationRequestedAt` before they ever reached tool
output — the metastore has returned all four in the JSON:API `meta` envelope since PSGO-140/PSGO-282
landed on the Go side (go-monorepo), but the Python client's `MetaObjectMeta` never modeled them and
none of the compact/full semantic-object output views forwarded them.

Consumers (Kai, an MCP client, a human via chat) had no way to tell, from the read tools alone,
whether a semantic object was `project`/`organization`/`targeted` scope, or which project an
`organization`-scope object was originally created in or promoted from.

Adds the four fields to `MetaObjectMeta`, plus a shared `_meta_fields` helper threaded through every
`CompactSemanticObject` subclass and the full `SemanticObject` view — so every existing read tool
now reports them with **no new parameters and no behavior change** for existing callers.

Purely additive/read-side — no RFC required per `CONTRIBUTING.md`'s table ("Enhancement to existing
tool / function (new param, extended response)").

## Testing

- [x] `tox` (pytest, ruff, check-tools-docs) — all green
- New unit tests in `tests/clients/test_metastore.py` (meta field parsing) and
  `tests/tools/semantic/test_tools.py` (compact/full object output carries the new fields, and stays
  `None` when the metastore doesn't return them)

## Checklist

- [x] Self-review completed
- [ ] Integration tests added/updated (not applicable — no new endpoint/behavior, purely field
      parsing)
- [x] Project version bumped (1.79.1 → 1.80.0)
- [x] Documentation updated — `TOOLS.md` regenerated (no diff: it doesn't document individual
      response fields)


src/keboola_mcp_server/tools/semantic/tools.py
```diff
@@ -38,6 +38,40 @@ class ConstraintValidationFinding(BaseModel):
 class CompactSemanticObject(BaseModel):
     id: str
     name: str | None = None
+    scope: str | None = Field(
+        default=None, description='Who can see this object: "project", "organization", or "targeted".'
+    )
+    project_id: int | None = Field(
+        default=None, description='Owning project id. Absent once promoted to "organization" scope.'
+    )
```
cjayyy, line 52: `project_id` is documented as *"Absent once promoted to `organization` scope"*, but nothing in the code produces that absence:

- `_meta_fields` (below) copies `meta.project_id` unconditionally.
- The metastore appears to return `projectId` on every object — the `_jsonapi_object` helper in `tests/clients/test_metastore.py` hardcodes `'projectId': 123` for all cases, including the new scope tests.

So an org-scope model emits **both** `scope: organization` and `project_id: 123`. A consumer that trusts the documented contract — treating a present `project_id` as proof the object is project-scoped, or reading it as "the one project that can see this" — concludes the object is private to project 123 when it's actually visible org-wide.

`test_full_semantic_object_also_carries_scope` builds exactly this state (`scope: 'organization'`, and the helper's default `projectId`) but asserts nothing about `project_id`, so the contradiction isn't caught.

Either drop `project_id` when `scope == 'organization'`, or fix the description to say it's always the owning/originating project.

tests/tools/semantic/test_tools.py
```diff
@@ -0,0 +1,62 @@
...
+        SemanticObjectType.SEMANTIC_MODEL,
+        'm1',
+        name='Shared Revenue Model',
+        meta={
+            'scope': 'targeted',
+            'projectId': 123,
+            'sourceProjectId': 456,
+            'targetProjectIds': [999999999],
+            'scopeElevationRequestedAt': '2026-01-03T00:00:00Z',
+        },
```
cjayyy, line 71: This fixture asserts two states that the field descriptions added in this same PR declare impossible:

1. **`sourceProjectId` with `scope: 'targeted'`** — `MetaObjectMeta.source_project_id` says *"Only ever present for `organization` scope"*, and `CompactSemanticObject.source_project_id` repeats it.
2. **`scopeElevationRequestedAt` with `scope: 'targeted'`** — the description says it's *"Set while a **`project`**-scope object has a pending request to promote it to `organization`"*.

Both can't be right at once. Either:

- the metastore really does return these for `targeted` scope → the descriptions are wrong, and a consumer that filters `source_project_id` by `scope == 'organization'` (as documented) silently drops real provenance data; or
- it doesn't → this fixture encodes a state that can never occur, and the test gives false confidence about a combination that will never be exercised.

Worth pinning the actual invariants against the Go side (PSGO-140 / PSGO-282) and then making the descriptions, `TOOLS.md`, and these fixtures agree. Right now three of the five new fields have fixtures contradicting their own documented contract — which is the signal that the invariants weren't nailed down before being written into the tool contract that Kai reads.

If the fields genuinely are independent of `scope`, the cleanest fix is to describe them as independent and split this into per-scope fixtures (`project` + pending elevation, `organization` + source, `targeted` + targets) so each documented combination is actually covered.

src/keboola_mcp_server/tools/semantic/tools.py
```diff
@@ -38,6 +38,40 @@ class ConstraintValidationFinding(BaseModel):
 class CompactSemanticObject(BaseModel):
     id: str
     name: str | None = None
+    scope: str | None = Field(
+        default=None, description='Who can see this object: "project", "organization", or "targeted".'
+    )
```
cjayyy: `scope` is typed `str | None` even though every description in this PR enumerates exactly three legal values. Nothing constrains it — unlike `SemanticObjectType` in `model.py`, which is a proper `str`-Enum.

Consider `Literal['project', 'organization', 'targeted'] | None`. Two concrete wins:

1. If the Go side adds or renames a value (`org`, or a new `restricted` tier), parsing fails loudly at the boundary instead of the value flowing through to clients that branch on the three known strings and silently fall through every case.
2. The generated MCP output schema then advertises the allowed values to callers, rather than an opaque `string`.

Same applies to `MetaObjectMeta.scope` in `clients/metastore.py`.

src/keboola_mcp_server/tools/semantic/tools.py
```diff
@@ -631,6 +674,19 @@ async def get_semantic_context(
...
     - `semantic_model_ids` optionally narrows the lookup to specific semantic models.
+    - Every object carries `scope` ("project", "organization", or "targeted"), `project_id`,
+      `source_project_id` (which project an "organization"-scope object came from, if known), and
+      `target_project_ids` (the sibling projects granted read access, for "targeted" scope).
+      `scope` describes visibility of the metastore object itself, not whether its underlying
+      Keboola Storage table is actually reachable from every project that can see it -- a
+      "targeted"/"organization" object's data may still need its bucket separately shared and
+      linked (`get_shared_buckets`/`link_shared_bucket`) before a query against it will work
+      outside the owning project.
+    - A `semantic-model`'s `scope_elevation_requested_at` being set means a project has asked an
```
cjayyy: *"Every object carries `scope` …"* likely overstates what the list path can actually deliver.

`_list_semantic_type_objects` in `service.py` calls `metastore.list_objects(object_type, limit=..., offset=...)` — `organization_scope` defaults to `False`, so it always hits `api/v1/repository/{type}`, never `.../organization`. Grepping the repo, `organization_scope=True` is passed only from `tests/clients/test_metastore.py` and `integtests/` — **no semantic tool ever sets it**.

If the project-scoped endpoint doesn't include org-scope objects, then from `get_semantic_context`/`search_semantic_context` those objects are missing entirely rather than returned with `scope: organization` — meaning the new field can never report a non-`project` scope through the listing paths, only through the by-ID detail path (`get_object`).

That's a pre-existing limitation, not something this PR introduced. The concern is that the new docstring reads as if it were solved: a user asking Kai *"which semantic models can I use?"* against a project that can read an org-scope model owned elsewhere still won't see it, and now the docs imply they would.

Two things worth doing:

1. Add a test where the listing returns a non-`project` scope — all three new tests go through `_metastore_object` directly, so nothing covers scope metadata surviving the `list_objects` → `_list_semantic_type_objects` path.
2. Either wire up `organization_scope=True` (separate PR is fine), or soften this bullet to say the fields are reported *when the object is visible in the current project's listing*.

src/keboola_mcp_server/tools/semantic/tools.py
```diff
@@ -38,6 +38,40 @@ class ConstraintValidationFinding(BaseModel):
...
+        ),
+    )
+    target_project_ids: tuple[int, ...] | None = Field(
+        default=None, description='Sibling project ids granted read access. Only present for "targeted" scope.'
+    )
+    scope_elevation_requested_at: str | None = Field(
+        default=None,
+        description='Set while a "project"-scope object has a pending request to promote it to "organization".',
+    )
+
```
cjayyy, line 67: `scope_elevation_requested_at` is inherited by all six `*Compact` subclasses and by the full `SemanticObject` view, but `TOOLS.md` documents it as a `semantic-model` concept only (*"A `semantic-model`'s `scope_elevation_requested_at` being set means…"*).

So it's emitted on datasets, metrics, relationships, glossary terms and constraints too. If the metastore only ever populates it for models, that's harmless in practice (nulls get filtered — see below) but the field still shows up in the output schema for all six types, which is misleading to a client reading the schema.

If elevation is genuinely model-only, it belongs on `SemanticModelCompact` rather than the shared base. If it can appear on any object type, `TOOLS.md` should drop the `semantic-model`-specific framing.

cjayyy (commented):
Reviewed for correctness. The plumbing is solid — my findings are all about the *semantics being promised* rather than the mechanics.

### What checked out

I verified these rather than assuming, since they're where this kind of change usually breaks:

- **Alias parsing is robust.** Ran the new `MetaObjectMeta` fields against pydantic 2.13: camelCase, snake_case, explicit nulls, and string-typed ids (`'456'`, `['999','1000']` — plausible from a JSON:API producer, especially given `organizationId` is already `str` here) all coerce correctly.
- **No union regression.** This was my main structural worry: `SemanticCompactObject` is a non-discriminated union, so adding five shared optional fields to the base could have caused smart-union mis-coercion. I instantiated all six subclasses plus `SemanticObject` through `SemanticObjectTypeContext` — every one round-trips as its exact type, including when only base fields are set.
- **"No behavior change for existing callers" holds.** `toon_serializer_compact` → `_filter_toon_nulls` drops all-`None` keys, so responses gain no new fields when the metastore omits them.
- **No field-name collisions** between the five new base fields and any subclass's fields.
- The new `tests/tools/semantic/__init__.py` is a real fix — that package was missing one.

### Findings

Five comments inline. The theme: **three of the five new fields have test fixtures that contradict their own documented contract.**

- `project_id` is documented as absent for `organization` scope, but `_meta_fields` copies it unconditionally and the metastore looks to always return it.
- `source_project_id` is documented as `organization`-only, yet the new test asserts it on a `targeted` object.
- `scope_elevation_requested_at` is documented as implying current scope is `project`, yet the same test sets it alongside `scope: 'targeted'`.

Nothing here crashes — these are contract issues. But these descriptions *are* the contract Kai and MCP clients consume, and right now a consumer implementing them as written would draw wrong conclusions about visibility. The pattern suggests the invariants weren't pinned down before being written into the tool docs.

Cheapest resolution: confirm the real invariants against PSGO-140 / PSGO-282 on the Go side, then make descriptions, `TOOLS.md`, and fixtures agree. Splitting the one combined fixture into per-scope cases would cover each documented combination and make any future drift obvious.

Two smaller items: `scope` would be better as a `Literal` of its three values than free-form `str`, and the "Every object carries `scope`" claim may overstate the list path, which never queries the `/organization` endpoint.

### Note

`mergeable_state` is `dirty` — needs a rebase on `main` before it can merge.

src/keboola_mcp_server/tools/semantic/tools.py
Matovidlo: **Fixed** ✅ in commit 3e161590

project_id's description no longer claims it's absent once promoted to organization scope -- it's now documented as present regardless of scope (copied unconditionally, as it always was), and test_full_semantic_object_also_carries_scope pins that an organization-scope object still carries its originating project_id.

tests/tools/semantic/test_tools.py
Matovidlo: **Fixed** ✅ in commit 3e161590

Split into three per-scope tests (project + pending elevation, organization + source_project_id, targeted + target_project_ids), each asserting the other scope-exclusive fields stay absent. No more impossible combined state.

src/keboola_mcp_server/tools/semantic/tools.py
Matovidlo: **Fixed** ✅ in commit 3e161590

scope is now Literal['project', 'organization', 'targeted'] via a shared ObjectScope type alias, applied to both MetaObjectMeta and CompactSemanticObject.

src/keboola_mcp_server/tools/semantic/tools.py
Matovidlo: **Fixed** ✅ in commit 3e161590

Reworded to state scope is only reported for an object visible in the current project's listing or loaded by id, and added test_list_semantic_type_objects_carries_scope_metadata_through_the_list_path covering that path directly (the existing tests all bypassed it via _to_semantic_service_data).

src/keboola_mcp_server/tools/semantic/tools.py
Matovidlo: **Fixed** ✅ in commit 3e161590

Dropped the semantic-model-specific framing -- the docstring now says "An object's scope_elevation_requested_at" rather than "A semantic-model's", matching that it's emitted on all six types via the shared base.

cjayyy (approved):

PR #694, opened 2026-09-01
AI-3790: RFC for resolving semantic-dataset physical data location

Matovidlo:
## Description

**Linear**: [AI-3790](https://linear.app/keboola/issue/AI-3790/verify-that-the-new-semantic-models-and-datasets)

### Change Type

- [ ] Major (breaking changes, significant new features)
- [ ] Minor (new features, enhancements, backward compatible)
- [x] Patch (bug fixes, small improvements, no new features) — docs only, no behavior change

### Summary

Design-only RFC, no implementation. Follows on from #693 (surfacing `scope`/`sourceProjectId`/
`targetProjectIds` on the read path): once a consumer can see that a semantic-dataset's parent
model is `organization`/`targeted` scope, there's still no way to tell whether the dataset's
`tableId` is actually queryable from the current project — scope on the metastore object and
bucket-sharing in Storage are two independent, unenforced mechanisms today.

Proposes a new `tools/semantic/data_location.py` that joins the `MetastoreClient` and
`StorageClient` already available on the same request-scoped `KeboolaClient` (mirroring the
pattern `tools/storage/shared_buckets.py` already uses) to resolve each dataset to one of
`local`/`linked`/`shared_not_linked`/`unreachable`, exposed as an opt-in enrichment on
`get_semantic_context` and as an extra pre-execution finding in `validate_semantic_query`.

RFC required per `CONTRIBUTING.md`'s table ("New end-to-end behavior or data flow", "new
architectural concept" — joining two clients that don't talk to each other today). No code in
this PR; implementation follows once the design is agreed.

## Testing

Not applicable — documentation only.

## Checklist

- [x] Self-review completed
- [ ] Unit tests added/updated — n/a, no code
- [ ] Integration tests added/updated — n/a, no code
- [x] Project version bumped (1.79.1 → 1.79.2)
- [x] Documentation updated — this PR *is* the documentation (new RFC)

PR #695, opened 2026-09-01
AI-3790: resolve where a semantic-dataset's physical data actually lives

Matovidlo:
## Description

**Linear**: [AI-3790](https://linear.app/keboola/issue/AI-3790/verify-that-the-new-semantic-models-and-datasets)

**Base branch note**: targets `AI-3790-surface-semantic-layer-scope` (#693), not `main` — this
depends directly on the `MetaObjectMeta.scope`/`.source_project_id` fields that PR adds. Rebase
onto `main` once #693 merges.

### Change Type

### Summary

Implements the design agreed in [#694](https://github.com/keboola/mcp-server/pull/694) (merged
RFC): `feature_spec/semantic_dataset_data_location/RFC.md`.

New `tools/semantic/data_location.py` joins the metastore's scope metadata with Storage
bucket/sharing lookups (`bucket_list`, `shared_bucket_list`) to resolve each semantic-dataset to
`local` / `linked` / `shared_not_linked` / `unreachable` — a semantic object's declared `scope`
only says the metastore record is visible elsewhere; it says nothing about whether the underlying
bucket was ever actually shared. Disambiguates a `bucket_id` match against the dataset's own
`sourceProjectId` when the metastore tracked one, falling back to a `bucket_id`-only match
otherwise (targeted-scope datasets never carry `sourceProjectId` — the common real case, not an
edge case).

Wired in as two **opt-in** enrichments, matching the RFC's scope (a routine call shouldn't pay for
extra Storage round trips it didn't ask for):
- `get_semantic_context(..., resolve_data_location=True)` attaches `data_location` to each
  returned `semantic-dataset`.
- `validate_semantic_query(..., resolve_data_location=True)` adds a warning-severity violation for
  each used dataset that isn't actually reachable here.

Also moved the semantic test package's shared object-builder/API-mock fixtures
(`_metastore_object`, `semantic_api_objects`, `mock_semantic_api`) from `test_service.py` into a
new `tests/tools/semantic/conftest.py` so the new test files can reuse them without a cross-file
import — matches the sibling `tools/components`/`tools/flow` test packages' own `conftest.py`
convention.

## Testing

- [x] `tox` (pytest, ruff, check-tools-docs) — all green; full suite 2169 passed, 24 skipped
      (pre-existing, unrelated)
- New unit tests: `tests/tools/semantic/test_data_location.py` (all four statuses, malformed/missing
  `tableId`, and the source-project disambiguation — including the case where a same-`bucket_id`
  shared bucket belongs to an unrelated project and must NOT be treated as a match), plus
  `get_semantic_context`/`validate_semantic_query` on/off coverage in `test_tools.py`

## Checklist

- [x] Self-review completed
- [ ] Integration tests added/updated (not applicable — no live-environment-only behavior)
- [x] Project version bumped (1.80.0 → 1.81.0)
- [x] Documentation updated — `TOOLS.md` regenerated


cjayyy (commented):
Reviewed the diff for correctness. The design is sound and the RFC/docstrings are unusually thorough — the substantive issues are all in `resolve_dataset_location`'s Storage lookup, which bypasses two conventions the rest of the repo follows. Both silently corrupt the status the feature reports, and the unit tests can't catch either because they hand-craft the payloads.

Also worth noting: I could not run the suite locally — `uv sync` fails on a pre-existing `pyproject.toml` parse error (`exclude-newer = "7 days"`, line 66), unrelated to this PR. Everything below comes from reading the code and its in-repo callers.

Checked and fine: warnings correctly don't flip `valid` (the "valid=True *and* violations empty" contract holds), and the docstring claim that `search_semantic_context` matches carry the scope fields is accurate — its output union is built from the compact types `_meta_fields` populates.

src/keboola_mcp_server/tools/semantic/data_location.py
```diff
@@ -0,0 +1,105 @@
...
+    return table_id.rsplit('.', 1)[0]
+
+
+async def resolve_dataset_location(client: KeboolaClient, dataset: SemanticDatasetData) -> DatasetLocation:
+    """Resolves where a semantic-dataset's tableId physically lives, relative to the calling project."""
+    bucket_id = _bucket_id_from_table_id(dataset.table_id)
+    if bucket_id is None:
+        return DatasetLocation(status=DatasetLocationStatus.UNREACHABLE)
+
+    for raw_bucket in await client.storage_client.bucket_list():
```
cjayyy, line 66: **`bucket_list()` bypasses the branch-aware helper — production-backed datasets get reported `unreachable`.**

This calls `client.storage_client.bucket_list()` directly rather than `merged_bucket_list` from `tools/storage_helpers.py`, whose own module docstring says:

> All callers (storage tools, search, etc.) should use these helpers instead of calling the storage client bucket/table methods directly.

Every other in-repo caller (`tools/search.py:117,142`, `tools/storage/tools.py:611`) does. Concrete failure: on a dev branch of a project with the `storage-branches` feature, `KeboolaClient` propagates `branch_id` into the storage client, so `bucket_list()` with no override resolves to `GET branch/<devBranchId>/buckets` — which returns *only* buckets created on that branch. A dataset with `tableId = out.c-RGP-Global.checkins` backed by an ordinary production bucket the project owns therefore matches nothing here, falls through to the shared-bucket lookup (absent there too, since it's locally owned), and is reported `status: unreachable` — "the underlying bucket is neither owned, linked, nor shared here. The query will likely fail." — for a bucket the caller can query fine.

The tests mock `bucket_list` with a bare `AsyncMock`, so no branch context is exercised.

src/keboola_mcp_server/tools/semantic/data_location.py
```diff
@@ -0,0 +1,105 @@
...
+        if raw_bucket.get('id') != bucket_id:
+            continue
+        source_project = get_nested(raw_bucket, 'sourceBucket.project')
```
cjayyy, line 69: **`sourceBucket` isn't requested, so `LINKED` is never detected.**

The only other place in the repo that reads this field — `BucketDetail.set_source_project`, via the same `sourceBucket.project` path — is fed by `_list_buckets`, which requests it explicitly:

```python
raw_bucket_data = await merged_bucket_list(client, include=['metadata', 'linkedBuckets'])
```

Here `bucket_list()` is called with `include=None`, so `params` stays empty and the `linkedBuckets` expansion isn't requested. For a bucket actually linked in from another project the response then carries no `sourceBucket` key, `get_nested` returns `None`, and line 74 returns `status: local` with `source_project_id=None` instead of `status: linked` with the source project id — losing exactly the cross-project provenance this feature exists to surface.

`test_linked_when_bucket_has_a_source_project` hand-injects a `sourceBucket` dict into the mocked list, so it passes regardless of whether the real API would have returned that field.

src/keboola_mcp_server/tools/semantic/data_location.py
```diff
@@ -0,0 +1,105 @@
...
+    # The dataset's own recorded source project, when the metastore tracked one (organization
+    # scope only -- absent for project/targeted scope, or when the creator opted out via
+    # dropSourceProject). Used to disambiguate a bucket_id that could plausibly exist in more
+    # than one project; when absent, fall back to a bucket_id-only match rather than refusing to
+    # resolve anything, matching this tool's existing "best-effort" framing elsewhere.
+    expected_source_project_id = dataset.data.meta.source_project_id if dataset.data.meta else None
+
+    fallback_match: SharedBucketDetail | None = None
+    for raw_shared in await client.storage_client.shared_bucket_list():
+        shared = SharedBucketDetail.model_validate(raw_shared)
```
cjayyy, line 85: **One unvalidatable shared bucket aborts the entire tool call.**

`SharedBucketDetail` declares `id`, `display_name` and `stage` as required (no defaults), and this validates *every* raw entry before the `shared.id != bucket_id` filter on the next line. So an unrelated shared bucket the caller never asked about is enough: any entry lacking `displayName` or `stage` (a `specific-users` share, or a shape variation on one stack) raises `ValidationError`, which `process_concurrently` captures and `unwrap_results` re-raises as `AggregateError('Failed to resolve one or more dataset locations.')` — failing the whole `get_semantic_context` call, which then returns no semantic context at all. That's a harsh failure mode for an opt-in *enrichment* layered on data already fetched successfully.

Filtering on the raw `raw_shared.get('id')` before validating would confine the blast radius to the one dataset.

(`get_shared_buckets` has the same validation exposure, but only over the requested page — and there the shared-bucket list is the tool's actual product rather than a side enrichment.)

src/keboola_mcp_server/tools/semantic/data_location.py
```diff
@@ -0,0 +1,105 @@
...
+                status=DatasetLocationStatus.LINKED, bucket_id=bucket_id, source_project_id=source_project['id']
+            )
+        return DatasetLocation(status=DatasetLocationStatus.LOCAL, bucket_id=bucket_id)
+
+    # The dataset's own recorded source project, when the metastore tracked one (organization
+    # scope only -- absent for project/targeted scope, or when the creator opted out via
+    # dropSourceProject). Used to disambiguate a bucket_id that could plausibly exist in more
+    # than one project; when absent, fall back to a bucket_id-only match rather than refusing to
+    # resolve anything, matching this tool's existing "best-effort" framing elsewhere.
+    expected_source_project_id = dataset.data.meta.source_project_id if dataset.data.meta else None
```
cjayyy, line 81: **The disambiguation reads the dataset's own meta, not the parent model's — so it's inert for targeted scope.**

The RFC specifies step 2 as matching "against the dataset's parent **model's** `sourceProjectId`", and its Problem section places `scope`/`sourceProjectId` on the parent model. This reads `dataset.data.meta.source_project_id` instead.

Per that field's own docstring in `clients/metastore.py` it is "Only ever present for 'organization' scope, and optional even there", and the PR description confirms targeted-scope datasets never carry it. So for every targeted-scope dataset — the cross-project sharing case this feature targets — `expected_source_project_id` is `None`, and the code silently takes the `bucket_id`-only fallback at line 97, returning the *first* shared bucket with a matching id. Two sibling projects in one organization that both expose e.g. `out.c-RGP-Global` resolve to whichever the API happened to list first, yielding a `source_project_id`/`source_bucket_id` pair a caller may then pass to `link_shared_bucket` — linking the wrong project's bucket.

I see the deviation is deliberate and documented in the PR description; the part that concerns me is that the resulting ambiguity is silent. Reading the parent model's meta (reachable via the dataset's `model_uuid`) would restore the intended disambiguation; failing that, it seems worth reporting the ambiguity to the caller rather than resolving it arbitrarily.

src/keboola_mcp_server/tools/semantic/data_location.py
```diff
@@ -0,0 +1,105 @@
...
+        if shared.id != bucket_id:
+            continue
+        if expected_source_project_id is not None and str(shared.project_id) == str(expected_source_project_id):
+            return DatasetLocation(
+                status=DatasetLocationStatus.SHARED_NOT_LINKED,
+                bucket_id=bucket_id,
+                source_project_id=shared.project_id,
+                source_bucket_id=shared.id,
+            )
+        fallback_match = fallback_match or shared
```
cjayyy, line 95: **Dead assignment when `expected_source_project_id` is set.**

This runs for any same-id shared bucket that didn't match `expected_source_project_id`, regardless of whether that id is set — but the only reader (line 97) is guarded on `expected_source_project_id is None`. So whenever the expected id *is* set, every write here is dead and the function returns `unreachable`, which is the intended behaviour per `test_unreachable_when_shared_bucket_belongs_to_a_different_project`.

No wrong result, but it reads as though a same-id/different-project bucket could still surface as a fallback when the expected project is known, and it obscures the actual intent: a known-but-unmatched source project is deliberately conclusive. `elif expected_source_project_id is None:` would make the control flow say that.

src/keboola_mcp_server/tools/semantic/tools.py
```diff
@@ -325,6 +385,66 @@ def _compact_semantic_object(obj: semantic_service.SemanticServiceData) -> Compa
...
+async def _resolve_dataset_locations(
+    client: KeboolaClient, groups: Sequence[semantic_service.SemanticServiceDataTypeGroup]
+) -> dict[str, DatasetLocation]:
+    """Resolves data_location for every semantic-dataset object across the given groups, keyed by object id."""
+    dataset_objects = [
+        obj for group in groups if group.object_type == SemanticObjectType.SEMANTIC_DATASET for obj in group.objects
+    ]
+    if not dataset_objects:
+        return {}
+    results = await process_concurrently(
```
cjayyy, line 397: **Both Storage lists are refetched once per dataset.**

`bucket_list` and `shared_bucket_list` are plain uncached HTTP GETs (no memoization anywhere in `clients/storage.py`), and their results depend only on the project/branch — not on the dataset. But this fans `resolve_dataset_location` out per dataset, and each invocation independently awaits both lists.

`get_semantic_context(resolve_data_location=True)` with an empty `ids` selection returns *all* semantic datasets in the project, so 50 datasets means 50 `GET branch/<id>/buckets` plus up to 50 `GET branch/<id>/shared-buckets` calls fetching byte-identical payloads, 10 in flight at a time. The parameter docstring does warn of "extra Storage API calls per dataset", but the per-dataset multiplier is avoidable: fetch both lists once and pass them into the resolver. Same duplication in `_dataset_location_findings_for_results`.

src/keboola_mcp_server/tools/semantic/data_location.py
Matovidlo: **Fixed** ✅ in commit 38df614a

Now routed through merged_bucket_list (tools/storage_helpers.py) with include=['metadata', 'linkedBuckets'], same as search.py/storage/tools.py.

src/keboola_mcp_server/tools/semantic/data_location.py
Matovidlo: **Fixed** ✅ in commit 38df614a

Same fix as the comment above -- merged_bucket_list is now called with include=['metadata', 'linkedBuckets'], so sourceBucket is actually populated for a real linked bucket.

src/keboola_mcp_server/tools/semantic/data_location.py
Matovidlo: **Fixed** ✅ in commit 38df614a

Now filters raw shared-bucket entries by id == bucket_id before calling SharedBucketDetail.model_validate, so an unrelated malformed entry is never validated at all when resolving a given dataset. Added test_unrelated_malformed_shared_bucket_entry_does_not_abort_resolution.

src/keboola_mcp_server/tools/semantic/data_location.py
Matovidlo: **Fixed** ✅ in commit 38df614a

resolve_dataset_location now takes model_source_project_id (the parent model's sourceProjectId, batched-fetched once per outer call by the two call sites via dataset.model_uuid) instead of the dataset's own meta. When it's None and more than one shared bucket shares the id, the result now sets a new `ambiguous: bool` field rather than silently picking the first match -- source_project_id/source_bucket_id are left unset in that case. Added test_get_semantic_context_disambiguates_by_the_parent_models_source_project (integration-level, proving the wiring) and test_ambiguous_when_multiple_projects_share_the_same_bucket_id_and_model_source_unknown.

src/keboola_mcp_server/tools/semantic/data_location.py
Matovidlo: **Fixed** ✅ in commit 38df614a

Rewritten as explicit if/elif/elif branches (known-project match / single-match fallback / ambiguous-multiple-match), so there's no assignment that's only ever dead when the other branch already fired.

src/keboola_mcp_server/tools/semantic/tools.py
Matovidlo: **Fixed** ✅ in commit 38df614a

resolve_dataset_location is now a pure function taking pre-fetched local_buckets/shared_buckets lists. Both call sites (_resolve_dataset_locations, _dataset_location_findings_for_results) fetch each list exactly once per outer call via asyncio.gather and pass it to every dataset being resolved; the validate_semantic_query path also reuses the semantic-models it already fetched for semantic_model_ids instead of a redundant per-dataset model lookup. Covered by assert_awaited_once() assertions on both storage calls in the get_semantic_context/validate_semantic_query tests.

src/keboola_mcp_server/tools/semantic/tools.py
```diff
@@ -434,11 +560,41 @@ def _to_tool_finding(finding: semantic_service.ConstraintValidationFinding) -> C
...
 
+def _dataset_location_finding(
+    dataset: semantic_service.SemanticDatasetData, location: DatasetLocation
+) -> semantic_service.ConstraintValidationFinding | None:
+    """Builds a pre-execution finding for a used dataset whose data isn't actually reachable here."""
+    label = dataset.display_name or dataset.id
+    if location.status == DatasetLocationStatus.SHARED_NOT_LINKED:
+        message = (
+            f'Dataset "{label}" (tableId {dataset.table_id}) is shared with this project but not linked in '
+            f'yet. Use link_shared_bucket with source_project_id={location.source_project_id!r}, '
```
cjayyy, line 577: **The new `ambiguous` case renders a finding that tells the caller to pass `None`.**

`ambiguous=True` is still a `SHARED_NOT_LINKED`, and this branches only on `status` — so with `source_project_id`/`source_bucket_id` deliberately left unset, `validate_semantic_query(resolve_data_location=True)` emits:

> Dataset "Checkins" (tableId out.c-RGP-Global.checkins) is shared with this project but not linked in yet. Use link_shared_bucket with source_project_id=None, source_bucket_id=None before querying it.

That hands the agent literal `None`s as though they were actionable arguments — the exact guessing the flag was added to prevent, just relocated from the resolver to the message. Worth an `if location.ambiguous:` branch that names the ambiguity and says to resolve it with the user before calling `link_shared_bucket`.

`test_ambiguous_when_multiple_projects_share_the_same_bucket_id_and_model_source_unknown` covers the resolver, but nothing exercises `ambiguous` through the finding path, which is why this got through.

src/keboola_mcp_server/tools/semantic/tools.py
```diff
@@ -673,6 +835,17 @@ async def get_semantic_context(
             )
         ),
     ] = (),
+    resolve_data_location: Annotated[
+        bool,
+        Field(
+            description=(
+                'For semantic-dataset objects, resolve whether their underlying Storage table is actually '
+                'reachable from this project and attach it as `data_location`. Off by default: it costs extra '
+                'Storage API calls per dataset (bucket_list/shared_bucket_list), so only turn it on when you '
```
cjayyy: **Stale cost description after the batching fix.**

"costs extra Storage API calls **per dataset** (bucket_list/shared_bucket_list)" was accurate before `b25fc0c`; now both lists are fetched exactly once per outer call regardless of dataset count. `validate_semantic_query`'s equivalent at line 1032 says "per used dataset" and is stale the same way.

These two strings are the tool descriptions an agent reads when deciding whether to set the flag, so an overstated cost will make it under-use a feature that's now roughly two extra requests total. `TOOLS.md` carries the same text in both `resolve_data_location` schemas and needs regenerating alongside.

cjayyy (commented):
Re-reviewed after `b25fc0c`. All six earlier findings are genuinely addressed in the code, not just in the commit message — I checked each one and ran the suite (94 passed). The `ambiguous` flag is a better answer to finding 4 than what I suggested: refusing to guess beats picking a candidate.

Two small loose ends the fix itself introduced or left behind. Neither is a regression of the original six, and neither needs another round-trip from me — push whenever.

src/keboola_mcp_server/tools/semantic/tools.py
Matovidlo: **Fixed** ✅ in commit 9e4fc8109afd3d7250720fc86381dc1392366a53

_dataset_location_finding now special-cases SHARED_NOT_LINKED + ambiguous: it says the source is ambiguous and asks the caller to resolve it with the user before calling link_shared_bucket, instead of filling in source_project_id=None/source_bucket_id=None as if those were usable arguments. Added test_validate_semantic_query_surfaces_ambiguous_dataset_location to exercise this through the finding path via validate_semantic_query, as you noted nothing did.

src/keboola_mcp_server/tools/semantic/tools.py
Matovidlo: **Fixed** ✅ in commit 9e4fc8109afd3d7250720fc86381dc1392366a53

Reworded both resolve_data_location docstrings (get_semantic_context and validate_semantic_query's "per used dataset" at line 1032) to say the two Storage calls are fetched once per outer call, not once per dataset -- matching the batching fix in b25fc0c. Regenerated TOOLS.md so both schema copies match.

Matovidlo:
Thanks for the re-review. Both loose ends fixed in `9e4fc810` (pushed):

* The `ambiguous` finding now says the source is ambiguous and asks to resolve it with the user before calling `link_shared_bucket`, instead of handing back `source_project_id=None`/`source_bucket_id=None`. Added a test exercising this through the finding path (`test_validate_semantic_query_surfaces_ambiguous_dataset_location`).
* Reworded the stale "per dataset"/"per used dataset" cost description in both `resolve_data_location` docstrings (and regenerated `TOOLS.md`) to reflect the `b25fc0c` batching fix -- it's now two extra calls per outer call, not per dataset.

Replied inline on the two anchored comments as well.

cjayyy (approved):

PR #13, opened 2025-03-20
KAB-881: add set up instructions for Cursor AI

vita-stejskal:

PR #697, opened 2026-09-02
RLS: row-level security query tool (pilot, do not merge)

padak:
## Description

**Linear**: AI-XXX (pilot, no issue yet)

### Change Type

### Summary

Pilot of row-level security (RLS) for SQL queries. **Draft, not for merge as-is** — technical validation phase, being tested on-site with a client.

RFC: `feature_spec/rls_query_tool/RFC.md` · Plan: `feature_spec/rls_query_tool/PLAN.md` · Strategy write-up (CZ/EN): https://artifact-hub-1304628444.hub.keboola.com/a/g-R9YYqZvSaAKJenwN4C0PHi

- New tool `query_data_rls(sql_query, query_name, user)`: rewrites every table in a `SELECT` to `(SELECT * FROM <table> WHERE <predicate>) AS <alias>` using rules from a YAML file (`--rls-rules-path` / `KBC_RLS_RULES_PATH`, never settable from a header), and discloses the applied rules in `applied_rules`.
- Fail-closed: non-`SELECT`, multiple statements, unparseable SQL, unknown dialect, a table or user without a rule, CTE names colliding with protected tables, `SELECT ... INTO`, table modifiers (SAMPLE / AT / PIVOT / FOR SYSTEM_TIME / alias column lists) and any FROM source that is not a plain table, subquery, VALUES or UNNEST are refused. A re-parse of the generated SQL asserts the output is still a single `SELECT` with every table wrapped.
- When the rules path is set the server registers **only** `query_data_rls`; `query_data` is not registered at all, so it cannot be re-enabled via `X-Allowed-Tools`. Without the path nothing changes (`TOOLS.md` unchanged).
- Rules are validated at startup; a missing or invalid file refuses to start the server.
- Principal binding: `KBC_RLS_PRINCIPAL_SOURCE=header|argument` must be chosen when RLS is on. In `header` mode the identity comes only from the `X-RLS-Principal` request header asserted by the calling application (AD / Google / Okta / app session — the MCP does not validate it and must be reachable only from that application); the tool's `principal` argument is refused. `argument` mode keeps the model-supplied identity for local/stdio use.
- Review hardening (see the review thread): mandatory `dialect:` in the rules file, `<bucket>.<table>` keys only (BigQuery dataset normalisation), `applied_rules` discloses table keys only, FROM-less function allowlist, output WHERE check, query size cap + off-loop rewrite, `outcome=` logging, sqlglot pinned, and **RLS mode forces read-only tools** so no write tool can bypass the rewrite.
- `examples/rls-demo/`: a small Node manual test harness (users, editable rules grid/YAML with validate-and-reload, SQL editor with live rewritten SQL, run-as-user and run-for-all-users comparison in tabs). Env-only configuration, local use only; not part of the server package.
- `query_data`'s execution body was extracted into a shared helper (`_execute_and_serialize`, `_to_csv`) — behaviour-preserving, covered by the existing tests.

Documented pilot limitations (see README section "Row-Level Security (pilot)"): in `argument` mode the identity is model-supplied; predicates are authored in the workspace dialect (no transpilation); no groups, column masking, audit log or hot reload; other tool docstrings still mention `query_data`.

## Testing

### Optional testing

Local: `ruff` clean; `pytest tests` green except the pre-existing, environment-specific `test_json_logging`; `TOOLS.md` regeneration produces no diff. Unit coverage: `tests/test_rls.py` (rules loading, rewrite, adversarial fail-closed cases), `tests/tools/test_sql.py`, `tests/test_server.py`, `tests/test_config.py`, `tests/test_cli.py`.

## Checklist

- [x] Self-review completed

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,32 @@ docker run \
 > **Note**: The server will use the Streamable HTTP transport and listen on `localhost:8000` for incoming connections at `/mcp`.
 > You can change `-p` to map the container's port somewhere else.
 
+### Row-Level Security (pilot)
+
+Restrict what each user can read from a shared table by giving the server a YAML rules file and
+starting it with `--rls-rules-path <file>` (or `KBC_RLS_RULES_PATH=<file>`). When set, the server
+exposes `query_data_rls` **instead of** `query_data`; the unrestricted tool is not registered at all.
+
+```yaml
+# rls.yaml -- table -> user -> SQL predicate (workspace dialect, inserted into WHERE verbatim)
+tables:
+  invoices:
+    petr: "country = 'CZ'"
+    monika: "country = 'DE'"
+    admin: "TRUE"              # unrestricted access must be written explicitly
+  in.c-crm.orders:             # <bucket>.<table> key takes precedence over a bare table name
+    petr: "country = 'CZ' AND status <> 'draft'"
+```
+
+`query_data_rls(sql_query, query_name, user)` rewrites every table in the SELECT to
+`(SELECT * FROM <table> WHERE <predicate>)` for that user, runs it, and reports the applied rules in
+`applied_rules`. It is fail-closed: a table without a rule for the user, a non-SELECT statement, or an
+unparseable query is refused and nothing is executed. The file is validated at startup; a broken file
+stops the server.
+
+Limitation: the `user` argument is supplied by the MCP client / model and is not verified by the server
+(same trust level as the `X-*` request headers). Suitable for a pilot behind a trusted client.
+
 ### Do I Need to Start the Server Myself?
 
 | Scenario | Need to Run Manually? | Use This Setup |
--- a/README.md
+++ b/README.md
@@ -381,7 +381,9 @@ docker run \
 ### Row-Level Security (pilot)
 
 Restrict what each user can read from a shared table by giving the server a YAML rules file and
-starting it with `--rls-rules-path <file>` (or `KBC_RLS_RULES_PATH=<file>`). When set, the server
+starting it with `--rls-rules-path <file>` (or `KBC_RLS_RULES_PATH=<file>`), plus
+`--rls-principal-source header|argument` (or `KBC_RLS_PRINCIPAL_SOURCE`) to say where the identity
+comes from — see [Who is the principal?](#who-is-the-principal) below. When set, the server
 exposes `query_data_rls` **instead of** `query_data`; the unrestricted tool is not registered at all.
 
 ```yaml
@@ -396,8 +398,8 @@ tables:
     petr: "country = 'CZ' AND status <> 'draft'"
 ```
 
-`query_data_rls(sql_query, query_name, user)` rewrites every table in the SELECT to
-`(SELECT * FROM <table> WHERE <predicate>)` for that user, runs it, and reports which tables were
+`query_data_rls(sql_query, query_name, principal)` rewrites every table in the SELECT to
+`(SELECT * FROM <table> WHERE <predicate>)` for that principal, runs it, and reports which tables were
 filtered in `applied_rules`. That list holds table keys only — the predicates themselves are never
 returned to the caller, since the filter is the admin's policy and disclosing it tells the reader
 what was withheld; failures that involve a predicate say only which rule could not be applied, with
@@ -416,18 +418,54 @@ both backends. Key matching follows each backend's own name resolution: case-ins
 Snowflake, case-**sensitive** on BigQuery, where `in_c_crm.Invoices` really is a different table
 from `in_c_crm.invoices`.
 
-Limitations:
+`query_data_rls` refuses any query longer than 20,000 characters, and any query too deeply nested for
+the rewriter to parse, before it reaches the workspace. The rewrite itself runs off the event loop, so
+a pathological query slows down only the session that sent it.
 
-- The `user` argument is supplied by the MCP client / model and is not verified by the server
-  (same trust level as the `X-*` request headers). Suitable for a pilot behind a trusted client.
-- Other tool docstrings and the bundled project prompt still refer to `query_data`; in RLS mode the
-  model must use `query_data_rls` instead. This is a known pilot limitation.
+A server started with `--rls-rules-path` is read-only as a whole: it exposes, and accepts calls to,
+only tools annotated `readOnlyHint=True`. Write tools such as `run_job`, `create_config`,
+`create_sql_transformation` or `deploy_data_app` are absent from `tools/list` and are refused if
+called, regardless of `X-Allowed-Tools` or the token role.
+
+#### Who is the principal?
+
+The principal is the identity the rules are keyed by. Where it comes from is a deployment decision,
+made once with `--rls-principal-source` / `KBC_RLS_PRINCIPAL_SOURCE`; a server with a rules file and
+no source refuses to start, because neither mode is a safe default for the other's deployment.
+
+- **`header`** (production shape) — the principal is the `X-RLS-Principal` request header. The
+  `principal` tool argument must be omitted; passing it is refused rather than ignored, and a request
+  without the header is refused too. The model can neither choose nor override the identity.
+- **`argument`** (pilot/demo) — the principal is the `principal` tool argument, supplied by the MCP
+  client or the model and not verified by anything. An `X-RLS-Principal` header is ignored in this
+  mode, with a warning in the log.
+
+The wrapper architecture behind `header` mode: an application authenticates its own end users
+(Active Directory, Google, Okta, its own session cookie — the MCP server has no opinion). That
+application is the only client of this MCP server, and it asserts the authenticated user on every
+request in the `X-RLS-Principal` header. The server does **not** verify the header — it trusts the
+wrapper — so the MCP endpoint must be reachable only from that application, never directly from an
+LLM, an end user's MCP client, or the public internet.
+
+Rule keys are therefore whatever strings the calling application asserts: e-mail addresses
+(`petr@example.com`), IdP subject ids (`a1b2c3-...`), or group names (`sales-cz`) all work equally
+well, as long as the rules file uses the very same strings the wrapper sends. Every query is logged
+with `outcome=`, `principal=` and `source=header|argument`, so a deployment that is meant to be
+header-bound can be audited for calls that were not.
 
 To try the feature by hand, [`examples/rls-demo/`](examples/rls-demo/) is a small Node.js app that
-starts the server in RLS mode and lets you pick a user, type SQL, see the rewritten statement live
-and compare the result across users — plus edit the rules file and reload it. It is an example only:
-it is not part of the server and nothing in `src/` depends on it. See its
-[README](examples/rls-demo/README.md) for the prerequisites and configuration.
+plays the wrapper role: it runs the server in RLS header mode, offers a mock sign-in, and lets you
+type SQL, see the rewritten statement live and compare the result across sign-ins — plus edit the
+rules file and reload it. It is an example only: it is not part of the server and nothing in `src/`
+depends on it. See its [README](examples/rls-demo/README.md) for the prerequisites and configuration.
+
+Limitations:
+
+- The server never verifies the principal itself. In `header` mode that trust is placed in the
+  calling application (which must be the only client able to reach the server); in `argument` mode it
+  is placed in the MCP client / model, which is suitable only for a pilot behind a trusted client.
+- Other tool docstrings and the bundled project prompt still refer to `query_data`; in RLS mode the
+  model must use `query_data_rls` instead. This is a known pilot limitation.
 
 ### Do I Need to Start the Server Myself?
 
--- a/README.md
+++ b/README.md
@@ -386,6 +386,7 @@ exposes `query_data_rls` **instead of** `query_data`; the unrestricted tool is n
 
 ```yaml
 # rls.yaml -- table -> user -> SQL predicate (workspace dialect, inserted into WHERE verbatim)
+dialect: snowflake             # required: snowflake or bigquery; predicates are never transpiled
 tables:
   invoices:
     petr: "country = 'CZ'"
@@ -399,7 +400,8 @@ tables:
 `(SELECT * FROM <table> WHERE <predicate>)` for that user, runs it, and reports the applied rules in
 `applied_rules`. It is fail-closed: a table without a rule for the user, a non-SELECT statement, or an
 unparseable query is refused and nothing is executed. The file is validated at startup; a broken file
-stops the server.
+stops the server. `dialect` pins the workspace backend the predicates are written for: every predicate
+is parsed in that dialect at startup, and a query against a workspace of any other dialect is refused.
 
 A bare key such as `invoices` matches a table of that name in every schema/bucket; use the
 `<bucket>.<table>` form to scope a rule.
--- a/README.md
+++ b/README.md
@@ -401,8 +401,15 @@ tables:
 unparseable query is refused and nothing is executed. The file is validated at startup; a broken file
 stops the server.
 
-Limitation: the `user` argument is supplied by the MCP client / model and is not verified by the server
-(same trust level as the `X-*` request headers). Suitable for a pilot behind a trusted client.
+A bare key such as `invoices` matches a table of that name in every schema/bucket; use the
+`<bucket>.<table>` form to scope a rule.
+
+Limitations:
+
+- The `user` argument is supplied by the MCP client / model and is not verified by the server
+  (same trust level as the `X-*` request headers). Suitable for a pilot behind a trusted client.
+- Other tool docstrings and the bundled project prompt still refer to `query_data`; in RLS mode the
+  model must use `query_data_rls` instead. This is a known pilot limitation.
 
 ### Do I Need to Start the Server Myself?
 
--- a/README.md
+++ b/README.md
@@ -397,8 +397,11 @@ tables:
 ```
 
 `query_data_rls(sql_query, query_name, user)` rewrites every table in the SELECT to
-`(SELECT * FROM <table> WHERE <predicate>)` for that user, runs it, and reports the applied rules in
-`applied_rules`. It is fail-closed: a table without a rule for the user, a non-SELECT statement, or an
+`(SELECT * FROM <table> WHERE <predicate>)` for that user, runs it, and reports which tables were
+filtered in `applied_rules`. That list holds table keys only — the predicates themselves are never
+returned to the caller, since the filter is the admin's policy and disclosing it tells the reader
+what was withheld; failures that involve a predicate say only which rule could not be applied, with
+the detail in the server log. It is fail-closed: a table without a rule for the user, a non-SELECT statement, or an
 unparseable query is refused and nothing is executed. The file is validated at startup; a broken file
 stops the server. `dialect` pins the workspace backend the predicates are written for: every predicate
 is parsed in that dialect at startup, and a query against a workspace of any other dialect is refused.
--- a/README.md
+++ b/README.md
@@ -388,11 +388,11 @@ exposes `query_data_rls` **instead of** `query_data`; the unrestricted tool is n
 # rls.yaml -- table -> user -> SQL predicate (workspace dialect, inserted into WHERE verbatim)
 dialect: snowflake             # required: snowflake or bigquery; predicates are never transpiled
 tables:
-  invoices:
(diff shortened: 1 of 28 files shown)


jordanrburger:
Review of `feat/rls-query-tool` @ `500a1a4c` (sqlglot 30.17.0 as locked). Everything below was run against the actual `rewrite_query` / `RlsRules.load` on this head, not the design doc. All four cases from your Slack reply, my whole bucket-2 list, and every CTE-shadowing shape I could think of are refused now — nothing below repeats those. `tests/test_rls.py` is green (104 tests); the 32 new params at the end all pass against this head as written.

### High

**1. RLS wraps `query_data`, but the same credential still runs unfiltered SQL through the other registered tools.**
With `--rls-rules-path` set, `create_sql_transformation` / `update_sql_transformation` (`tools/components/tools.py:183-196`), `run_job` (`tools/jobs.py:35-40`), `create_config` / `update_config` (writers), and `modify_streamlit_data_app` / `deploy_data_app` (`tools/data_apps.py:47-81`) stay registered; none carries `readOnlyHint`. Three calls bypass the rewrite: `create_sql_transformation` with `SELECT * FROM "in.c-crm"."orders"` → `run_job` → a writer config or a data app reads the output with the project token. A transformation can also `UPDATE` the protected table so every row satisfies petr's predicate. The only server-side gate today is `token_role == 'readonly'` (`mcp.py:1023`) or the infra-injected `X-Read-Only-Mode` header, and neither is required by the PR or README. Fix: when `rls_rules is not None`, force the read-only tool filter in `on_list_tools`/`authorize_tool_call` (or refuse to start unless the token role is readonly). The README line "the unrestricted tool is not registered at all" is true of the tool, not of the capability.

### Medium

**2. Predicates are parsed in the workspace's dialect, but the rules file has no dialect pin.** (High if one server serves more than one backend.)
`RlsRules.load()` is dialect-agnostic (`rls.py:98`); `rewrite_query` parses the predicate under whatever `get_sql_dialect()` returns (`sql.py:440`, `rls.py:383`). Verified: rule `"status" <> 'deleted'` on a BigQuery workspace becomes `WHERE 'status' <> 'deleted'` — left operand is a `Literal`, so the filter is constant TRUE and the table is unrestricted. The module docstring names this hazard but nothing enforces "one file, one backend"; multi-project/OAuth deployments hit it silently. Fix: top-level `dialect: snowflake|bigquery` in the YAML, refuse the rewrite when the workspace dialect differs, and parse predicates at load in that dialect (which also makes the load-time check real).

**3. Rule-key matching is looser than "qualified takes precedence" suggests.** Verified on head:
- bare `invoices` matches `"in.c-hr"."invoices"`, `"INVOICES"`, `"invoices"` (three distinct Snowflake objects) and BigQuery `Invoices` (BQ table names are case-sensitive by default): `SELECT * FROM "in.c-hr"."invoices"` gets petr's CRM predicate and reports `applied_rules=["invoices: ..."]`.
- catalog is ignored (`rls.py:113-116`): `OTHER_DB."in.c-crm"."orders"` and `` `otherproj`.`ds`.`invoices` `` match `in.c-crm.orders` / `ds.invoices`.
- on BigQuery the README's `in.c-crm.orders` key can never match — the dataset is `in_c_crm` (`workspace.py:534`) — so lookup silently falls back to the bare key. With `{in.c-crm.orders: strict, orders: TRUE}` the permissive bare rule wins (`` SELECT * FROM `in.c-crm`.`orders` `` → `rules=['orders: TRUE']`).
- a single quoted identifier `"in.c-crm.orders"` matches the *qualified* key.
Fix: fully-qualified keys only (`<schema>.<table>`, optionally `<db>.<schema>.<table>`), validate at load that a key is a form the matcher can produce (no quotes, no `*`), and make bare keys an explicit opt-in or drop them.

**4. Predicate text is disclosed to the model, on the happy path and on errors.**
`applied_rules` carries `"<key>: <predicate>"` (`rls.py:371`) straight into the tool output (`sql.py:451`) — `country = 'CZ'` invites `user: monika` / `'DE'`. Error paths leak too: `rls.py:385` echoes the predicate (`Failed to parse 'country = = 1'`), `rls.py:303` names a table that appears only in an admin predicate (`rewrite left an unwrapped table reference: hr_salaries_2026`), `sql.py:297` passes backend errors through (`invalid identifier 'COUNTRY'`). Fix: `applied_rules` = table keys only (`["invoices", "in.c-crm.orders"]`); predicate-side failures → one generic message, details to the server log.

**5. Load-time validation lets misconfiguration widen access or fail only at query time.** Verified with `RlsRules.load()`:
- duplicate keys after case-fold silently overwrite (`rls.py:101-102`): `invoices: {petr: "FALSE"}` + `Invoices: {petr: "TRUE"}` loads as `{'invoices': {'petr': 'TRUE'}}`; same for `petr`/`Petr`. Reject duplicates after normalization.
- `id IN (SELECT id FROM secret)`, `country`, `1` all load. The first is refused on every query by `_check_output`; the others surface as backend errors. At load: reject any `exp.Table`/`exp.Subquery` inside the predicate and require a boolean root (`Predicate`/`And`/`Or`/`Not`/`Boolean`), per the RFC's "broken file stops the server".
- YAML `yes:`/`on:` become user `true`; `"": "TRUE"` (empty user) loads; keys containing quotes (`'"in.c-crm"."orders"'`) or `*` load and never match, so every query is refused. Require `isinstance(key, str)`, non-empty, `^[A-Za-z0-9_.\-]+$`.

**6. FROM-less SELECTs are unrestricted, and `test_rls.py:231` enshrines it.**
Any function call passes when no table is present: `SELECT GET_DDL('table','secret')` (columns of a table the user has no rule for), `GET_DDL('view', ...)` (view SQL = business logic), `SYSTEM$EXPLAIN_PLAN_JSON('SELECT * FROM secret')`, `SYSTEM$CLUSTERING_INFORMATION('secret')`, and `SNOWFLAKE.CORTEX.SEARCH_PREVIEW('svc', '{"query":"x","limit":100}')`, which returns rows from the indexed table. Fix: refuse `exp.Anonymous`/`exp.Func` named `GET_DDL`, prefixed `SYSTEM$`, or under `SNOWFLAKE.CORTEX` — or, simpler and closed-by-default, refuse a FROM-less SELECT that calls anything outside a small allowlist (`CURRENT_DATE`, `CURRENT_TIMESTAMP`, ...). Drop or invert the `GET_DDL` happy-path test.

**7. The CTE collision guard refuses the most natural LLM query shape.**
`WITH orders AS (SELECT * FROM "in.c-crm"."orders" WHERE amount > 0) SELECT COUNT(*) FROM orders` → `CTE name(s) collide with protected table(s): orders` (`rls.py:342-344`). Any CTE named like the bare table of *any* rule key is refused even when its body is correctly wrapped. `_is_cte_reference` now resolves scope precisely and fails closed on its own (every shape I tried is in the params below), so the blanket guard is redundant belt-and-braces with a real cost: this will be the first day-one complaint. Narrow it to "CTE named X whose body references bare X without RECURSIVE" (already caught) or drop it.

### Low

**8.** `_check_output` never checks for a WHERE: `_check_output("SELECT * FROM (SELECT * FROM invoices) AS invoices", dialect='snowflake')` is accepted (`rls.py:297-301`). The safety net should require `select.args['where']`, ideally equal to the predicate it just inserted.

**9.** `_identifier_key` (`rls.py:156-166`) encodes Snowflake's identifier model: quoting decides case-sensitivity. On BigQuery backticks are semantically null and table names are case-sensitive by default. Effects on head: `` WITH `secret` AS (SELECT 1) SELECT * FROM secret `` is refused as "ambiguous" on BQ (false positive, harmless); `WITH secret AS (SELECT 1) SELECT * FROM SECRET` on BQ is accepted as a CTE reference — correct only if BQ folds CTE names case-insensitively, which I could not confirm from the docs. Please verify that one on a real BQ workspace before the pilot touches BQ. Snowflake side is right.

**10.** `sqlglot ~= 30.0` (`pyproject.toml:25`) admits any 30.x; the allowlist depends on node types (`TableFromRows`, `Explode`, the `with`→`with_` rename the code already dances around). Pin `==30.17.0` (or `~=30.17`) and let the adversarial suite be the upgrade tripwire.

**11.** The rewrite is sync CPU on the event loop: 20k-term `OR` chain → 3.0 s, 2000 JOINs → 1.4 s, 200 nested parens → `RecursionError` (not `RlsError`; `tool_errors` still catches it, so fail-closed). Cap `len(sql_query)` and/or run `rewrite_query` in `asyncio.to_thread`.

**12.** `(SELECT * FROM invoices)` is refused as `got Subquery` — unwrap a top-level parenthesised SELECT. Parse errors return sqlglot's message with ANSI escapes (`\x1b[4mAND\x1b[0m`) to the model (`rls.py:322`); strip them or use a generic message. Same for the unquoted `in.c_crm.orders` form (`Expected table name but got <Token ...>`) — invalid Snowflake anyway, but the message doesn't say "quote the bucket".

**13.** The per-query log (`sql.py:447`) has user, query name and rules; refusals log only the `RlsError` text with no user, and successes have no row count. Add user + table keys + outcome to refusals and row count to successes — that is what makes the pilot evaluable.

**14.** `user` is still a tool argument (`sql.py:400-409, 443`); nothing on this head binds it to the transport principal. Still the boundary, acknowledged.

### Confirmed handled on this head (no action)

Every bucket-2 construct (`TABLE(...)`, `RESULT_SCAN`, `IDENTIFIER(...)`, `@stage`, `DIRECTORY`, `SEMANTIC_VIEW(...)`, `EXTERNAL_QUERY`, `ML.PREDICT`, `EXECUTE IMMEDIATE`, `:tbl`, `?`, 4-part names, `LATERAL FLATTEN`) · a no-rule table in every subquery position (SELECT list, WHERE, EXISTS, ORDER BY, HAVING, QUALIFY, LIMIT, LATERAL, VALUES, UNNEST, set operations) · all CTE shadowing shapes incl. `RECURSIVE`, quoted collisions, JOIN'd subquery, `EXISTS`, scalar subquery, UNION branch, nested WITH, later-sibling · alias equal to a protected name cannot fool `_check_output` (it walks `exp.Table` nodes only) · `in.`/`out.` prefixes are fine as long as the model uses the quoted FQN from `get_tables` (`"in.c-crm"."orders"` and `"DB"."in.c-crm"."orders"` both parse and match) · trailing `;`, `--`/`/* */` comments · `RESULT_SCAN` replay is moot since Query Service opens a fresh session per job · rules path unreachable from headers (`_HEADER_ELIGIBLE_FIELDS`, test present) · `query_data` absent from `tools/list` (`test_server.py:693`) · `t.*`, `EXCLUDE`/`EXCEPT`, `QUALIFY`, `GROUP BY ALL`, `MATCH_RECOGNIZE`, `CONNECT BY`, `UNNEST ... WITH OFFSET`, `TOP`, `FETCH FIRST` all round-trip with the wrapper.

### New cases for `TestRewriteQuery::test_rewrite_fails_closed`

All 32 pass against `500a1a4c` with the existing `rules` fixture; paste as-is.

```python
            # --- FROM sources with no plain identifier ---
            ('SELECT * FROM IDENTIFIER($tbl)', 'petr', 'snowflake', 'unsupported table reference'),
            ("SELECT * FROM IDENTIFIER('invoices')", 'petr', 'snowflake', 'unsupported table reference'),
            ('SELECT $1 FROM @my_stage/invoices.csv', 'petr', 'snowflake', 'unsupported table reference'),
            ('SELECT * FROM DIRECTORY(@stg)', 'petr', 'snowflake', 'unsupported table reference'),
            ('SELECT * FROM SEMANTIC_VIEW(sv METRICS invoices.total)', 'petr', 'snowflake', 'unsupported table reference'),
            ('SELECT * FROM a.b.c.d', 'petr', 'snowflake', 'unsupported table reference'),
            ('SELECT * FROM :tbl', 'petr', 'snowflake', 'unsupported table reference'),
            ("SELECT * FROM EXTERNAL_QUERY('conn', 'SELECT * FROM invoices')", 'petr', 'bigquery', 'unsupported table reference'),
            ('SELECT * FROM ML.PREDICT(MODEL `m`, TABLE invoices)', 'petr', 'bigquery', 'unsupported table reference'),
            ("EXECUTE IMMEDIATE 'SELECT * FROM invoices'", 'petr', 'bigquery', 'SELECT'),
            ('SELECT * FROM invoices, LATERAL FLATTEN(input => invoices.items) f', 'petr', 'snowflake', 'unsupported FROM source'),
            # --- a table without a rule, in every position a subquery can appear ---
            ('SELECT (SELECT MAX(x) FROM secret) FROM invoices', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM invoices WHERE EXISTS (SELECT 1 FROM secret)', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM invoices, LATERAL (SELECT * FROM secret) s', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM invoices ORDER BY (SELECT MAX(x) FROM secret)', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM invoices LIMIT (SELECT COUNT(*) FROM secret)', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM invoices QUALIFY id IN (SELECT id FROM secret)', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM (VALUES ((SELECT MAX(x) FROM secret))) v', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM UNNEST((SELECT ARRAY_AGG(x) FROM secret))', 'petr', 'bigquery', "table 'secret'"),
            ('SELECT * FROM (secret)', 'petr', 'snowflake', "table 'secret'"),
            # an alias equal to a protected table name must not stand in for a rule
            ('SELECT * FROM secret AS invoices', 'petr', 'snowflake', "table 'secret'"),
            ('SELECT * FROM (SELECT * FROM secret) AS invoices', 'petr', 'snowflake', "table 'secret'"),
            # --- CTE shadowing, remaining shapes ---
            (
                'WITH RECURSIVE invoices AS (SELECT 1 AS n UNION ALL SELECT n + 1 FROM invoices WHERE n < 3) '
                'SELECT * FROM invoices',
                'petr',
                'snowflake',
                'collide',
            ),
            ('WITH "Invoices" AS (SELECT 1) SELECT * FROM "Invoices"', 'petr', 'snowflake', 'collide'),
            ('WITH `Invoices` AS (SELECT 1) SELECT * FROM `Invoices`', 'petr', 'bigquery', 'collide'),
            ('SELECT * FROM secret JOIN (WITH secret AS (SELECT 1) SELECT * FROM secret) s ON TRUE', 'petr', 'snowflake', 'another scope'),
            ('SELECT * FROM secret WHERE EXISTS (WITH secret AS (SELECT 1) SELECT 1)', 'petr', 'snowflake', 'another scope'),
            ('SELECT (WITH secret AS (SELECT 1) SELECT 1) FROM secret', 'petr', 'snowflake', 'another scope'),
            ('SELECT * FROM secret UNION ALL (WITH secret AS (SELECT 1) SELECT * FROM secret)', 'petr', 'snowflake', 'another scope'),
            ('WITH a AS (WITH secret AS (SELECT 1) SELECT * FROM secret) SELECT * FROM a, secret', 'petr', 'snowflake', 'another scope'),
            ('WITH a AS (SELECT * FROM secret), secret AS (SELECT 1) SELECT * FROM a', 'petr', 'snowflake', 'another scope'),
            ('WITH secret AS (SELECT 1) SELECT * FROM public.secret', 'petr', 'snowflake', "table 'secret'"),
```

Worth pinning in `test_rewrite` too, so the fixes for #6/#7 don't regress them (all rewrite correctly today): `SELECT * FROM invoices;` · `SELECT * FROM invoices AS t, UNNEST(t.items) AS item WITH OFFSET AS off` (bigquery) · `SELECT id, ROW_NUMBER() OVER (PARTITION BY country ORDER BY id) rn FROM invoices QUALIFY rn = 1` · `WITH RECURSIVE r AS (SELECT 1 AS n UNION ALL SELECT n + 1 FROM r WHERE n < 3) SELECT * FROM r, invoices`.

padak:
Thanks — this was exactly the review the pilot needed. Everything below is on `feat/rls-query-tool` @ `e6b1cd11` (+ the principal commit noted at the end), one commit per finding, all 32 of your cases added first as the regression net (`9e3d6ffb`). Full suite 2401 green.

**High**
- **#1 write tools bypass the rewrite** — fixed, `4cd55443`: with `--rls-rules-path` the server is read-only as a whole — `ToolAuthorizationMiddleware` treats RLS mode as forced read-only (only `readOnlyHint=True` tools listed and callable, regardless of headers or token role); the raw preview route now gets the server state too, so it is not a back door. Tests in `test_authorization.py` / `test_server.py`.

**Medium**
- **#2 no dialect pin** — fixed, `82843679`: `dialect: snowflake|bigquery` is mandatory in the rules file; predicates are validated at load *in that dialect*; a workspace of a different dialect is refused at rewrite time.
- **#3 loose key matching** — fixed, `b5f34453`: keys must be `<bucket>.<table>`; bare-name fallback removed (an unqualified table reference is refused); a single quoted identifier containing dots does not match a qualified key; on BigQuery keys are normalised the same way the server names datasets (`in.c-crm` → `in_c_crm`), and dataset/table parts match case-sensitively there (`6ca43a5c`). Catalog is still ignored when present — the workspace can only reach its own project DB; documented.
- **#4 predicate disclosure** — fixed, `1c15142b`: `applied_rules` is table keys only; predicate-side failures are one generic message, details go to the server log.
- **#5 load-time validation** — fixed, `ff93db5f`: duplicates after case-fold, non-string/empty/`yes:`-style keys, non-boolean roots and any table/subquery inside a predicate are rejected at load, naming the key.
- **#6 FROM-less functions** — fixed, `9ca289a4`: a query without FROM may only call a small allowlist (`CURRENT_DATE`, casts, arithmetic…); `GET_DDL`, `SYSTEM$*`, `CORTEX.*`, `ML.*` are refused anywhere in the query. The `GET_DDL` happy-path test is inverted.
- **#7 CTE guard too broad** — fixed, `b497eb20`: the blanket "CTE named like a rule key" refusal is gone; `WITH orders AS (SELECT * FROM "in.c-crm"."orders" WHERE …) SELECT … FROM orders` now rewrites. Scope-chain resolution, quoting rules and the non-RECURSIVE self-reference refusal stay (your three `collide` cases were re-evaluated under the narrowed rule — see the commit message).

**Low**
- **#8** `_check_output` now requires the inserted WHERE on every wrapper and compares it to the predicate it inserted (`7b8e7d8d`).
- **#9** verified on a real BigQuery workspace: BQ folds CTE names case-insensitively and backticks are semantically null for CTE names; table/dataset names are case-sensitive. `_identifier_key` and rule matching are now dialect-aware (`6ca43a5c`), so the BQ false positives are gone.
- **#10** `sqlglot == 30.17.0` pinned with a comment naming the adversarial suite as the upgrade tripwire (`6862d094`).
- **#11** 20 000-char cap, rewrite in `asyncio.to_thread`, `RecursionError` mapped to a clean refusal (`c3249b68`).
- **#12** top-level parenthesised SELECT unwrapped; ANSI stripped from parse errors; unquoted `in.`/`out.` path gets a "quote the bucket" hint (`0d9f820f`).
- **#13** one-line `outcome=ok|refused` log with user, query name, table keys and row count (`ed9521ca`).
- **#14** identity — landing next (in progress on the branch): the principal is bound to the transport instead of a tool argument. New `X-RLS-Principal` request header + `KBC_RLS_PRINCIPAL_SOURCE=header|argument` (deployment-level, must be chosen explicitly when RLS is on). In `header` mode the tool has no identity argument at all, a missing header is refused, and passing the argument anyway is refused too. The intended topology: an application that authenticates its own users (AD/Google/Okta/app session) is the only caller of the MCP and asserts the principal per request; the MCP does not validate it, so it must not be reachable by the LLM/client directly. Rule keys are whatever the app asserts (emails, IdP subjects, group names). I will post the SHA here once it is pushed.

Not changed on purpose: catalog part of a reference is ignored (see #3). Happy to take another pass whenever you have time.

padak:
Follow-up to #14: pushed `e999d292` — `X-RLS-Principal` header + `KBC_RLS_PRINCIPAL_SOURCE=header|argument` (mandatory choice when RLS is on; the tool argument is now `principal` and is refused in header mode). `d5bbdcb3` turns `examples/rls-demo` into the wrapper: mock sign-in, MCP over streamable-HTTP with the principal asserted per request. Branch head `d5bbdcb3`, full suite 2428 green.

Matovidlo:
supperseeded by #709

PR #612, opened 2026-07-02
DMD-1701: RFC — merge request MCP tools (non-SOX)

martinsifra:
## Description

**Linear**: [DMD-1701](https://linear.app/keboola/issue/DMD-1701)

### Change Type

- [ ] Major (breaking changes, significant new features)
- [ ] Minor (new features, enhancements, backward compatible)
- [ ] Patch (bug fixes, small improvements, no new features)

> RFC only — no code yet, so no change-type/version bump applies. This PR is for design review per `CONTRIBUTING.md` (new tools require an agreed RFC before implementation).

### Summary

Adds `feature_spec/branches_merge_requests_mcp/RFC.md` — a design proposal for exposing the **non-SOX** Branches 2.0 merge-request flow (`branches-merge-requests`) through MCP tools, for the ["Branches 2.0 - MCP integration"](https://linear.app/keboola/project/finalize-dev-branches-327e7756d2fd/overview) milestone.

Highlights:
- Maps the nine Connection MR endpoints + state machine and the non-SOX specifics (0 default approvals, admin/share-only merge, async merge job).
- Proposes a tiered toolset: Tier A MVP (`get_merge_requests`, `get_merge_request`, `create_merge_request`, `request_merge_request_review`, `merge_merge_request`, a `publish_branch` orchestrator, and read-only `get_branches`) + Tier B parity (`approve`, `request_changes`, `update`).
- `publish_branch` is the "MCP simplifies the process" tool for less-technical AI-chat users: one call that find-or-creates the MR, checks conflicts, requests review, merges, and stops with a plain-language `next_step` at any gate needing a human.
- Feature + role gating via the existing `ToolsFilteringMiddleware`; ID→name resolution and human-readable changelog summaries.

Scope decisions baked in: MR flow + read-only branch listing only — **no** branch create/switch (deferred; needs session-state/workspace architecture work). SOX flow explicitly out of scope.

## Testing

- [ ] Tested with Cursor AI desktop (`Streamable-HTTP` transports)

> N/A — documentation/RFC only, no runtime changes.

## Checklist

- [x] Self-review completed
- [ ] Unit tests added/updated (if applicable)
- [ ] Integration tests added/updated (if applicable)
- [ ] Project version bumped according to the change type
- [ ] Documentation updated (if applicable)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #699, opened 2026-09-06
DMD-1701: merge-request MCP tools (Branches 2.0, non-SOX)

martinsifra:
## Description

**Linear**: [DMD-1701](https://linear.app/keboola/issue/DMD-1701)
**RFC**: [`feature_spec/branches_merge_requests_mcp/RFC.md`](https://github.com/keboola/mcp-server/pull/612) (PR #612, reviewed separately; this PR implements it)

### Change Type

### Summary

Implements the Branches 2.0 (non-SOX) merge-request flow as **9 MCP tools** in a new package `tools/merge_requests/`, so a user on a development branch can ship changes to production from the chat:

| Tool | Session | Role | Notes |
|---|---|---|---|
| `get_merge_requests` | any | any | list (optional `state` filter, raw or derived) or batched detail with the derived status |
| `create_merge_request` | dev branch | admin/share/OAuth | source = session branch, target = production; no branch params |
| `update_merge_request` | any | admin/share/OAuth | the only way to disarm auto-merge |
| `request_merge_request_review` | dev branch | admin/share/OAuth | rarely needed: merge skips review itself when approvals suffice |
| `approve_merge_request` | any | admin/share/OAuth | valid only in `in_review` |
| `request_merge_request_changes` | any | admin/share/OAuth | reviewer's "reject"; not terminal |
| `merge_merge_request` | dev branch | admin/share/OAuth | awaits the Storage job; 409 mapped by `code` (`conflicts` / `not_ready`) |
| `get_merge_request_conflicts` | dev branch | any | conflicts + three-way diff + per-path `changed_by` classification |
| `resolve_merge_request_conflict` | dev branch | admin/share/OAuth | `take=ours|theirs|delete` or a strict custom body, via rebase `{version, diff}` |

Every MR-returning tool carries the same **derived status** (`derived_state`, `merge_blockers`, `mergeable`, `allowed_actions`, `viewer`, live `conflicts`) and a deterministic **`next_step`** sentence (15-row decision table, `status.py`) so the agent branches on data, not prose. The derivations are a port of the kbagent CLI's Layer 2 and read server-first (DMD-1988) with the local tables as fallback.

Other changes:

- `clients/storage.py`: `merge_requests_list`, `merge_request_detail(include_activity_log)`, `merge_request_conflicts`, `merge_request_create/update/request_review/approve/request_changes/merge`, `configuration_diff`, `configuration_rebase(version, diff)`; `'branches-merge-requests'` added to `ProjectFeature`.
- `mcp.py`: three-axis gating in `authorize_tool_call` by name set (`MERGE_REQUEST_TOOL_NAMES`, `MERGE_REQUEST_BRANCH_ONLY_TOOLS` in `tools/constants.py`): feature (hide + deny), role with the existing OAuth carve-out (hide writes + deny), session branch (**deny only**, the branch-only tools stay visible on production so the model keeps their descriptions and can hand the user off). `on_call_tool` now stores `token_info` in session state (`TOKEN_INFO_STATE_KEY`) so tool bodies do not verify the token a second time.
- `links.py`: `get_merge_request_link` (the source branch's development overview, as Connection's MR emails link it).
- `TOOLS.md` regenerated; version `1.80.1 → 1.81.0`; `uv.lock` refreshed.

Deviations from the RFC worth knowing: `resolved` is typed `dict` in the tool schema (validated by `ResolvedConfiguration` with `StrictBool` before any network call) because the tool-schema test requires a `type` on every parameter; the write tools cost one more request than the RFC's count (the `branches_list` needed for branch names / links), i.e. 4 uncached / 3 with the token cache. `get_merge_requests` and `get_merge_request_conflicts` additionally take `project_id` and are dispatched single-target by `MultiProjectMiddleware` (merge-request ids are project-specific; the conflict tool is bound to the session branch), so they are not fanned out across a multi-project scope. `update_merge_request` also ships (RFC decision C2). A self-review pass (`/code-review`, 10 findings) is folded in as the last commit.

## Testing

### Optional testing

Unit tests (`tests/tools/merge_requests/`, `tests/test_mcp.py`, `tests/test_preview.py`, `tests/clients/test_client.py`, `tests/test_server.py`): derivation tables row by row incl. server-first overrides, all 15 `next_step` rows and their precedence rules, per-path classification (pseudo-paths, `agreed`, deleted/holed sides), the MR-id convention, merge 409 shapes / job success / job error / timeout, every resolve mode and refusal (no request sent on a validation failure), request counts, the gating table for `authorize_tool_call` and list-time filtering, the `token_info` cache.

Integration tests (`integtests/tools/test_merge_requests.py`): happy path `create → merge → published` and the conflict loop. **Not run yet** — they need a project with the `branches-merge-requests` feature (`INTEGTEST_STORAGE_TOKEN_MERGE_REQUESTS`, registered in `tox` `pass_env` and the integtests README) and are skipped when the variable is unset. Manual E2E on such a project is pending for the same reason.

## Checklist

- [x] Self-review completed
- [x] Integration tests added/updated (if applicable) — written, run pending a `branches-merge-requests` project
- [x] Documentation updated (if applicable) — `TOOLS.md`, `integtests/README.md`

### DIFF
--- a/integtests/tools/test_merge_requests.py
+++ b/integtests/tools/test_merge_requests.py
@@ -1,0 +1,281 @@
+"""Integration tests for the merge-request tools (Branches 2.0, non-SOX flow).
+
+They need a dedicated project WITH the `branches-merge-requests` feature, addressed by the
+`INTEGTEST_STORAGE_TOKEN_MERGE_REQUESTS` token (outside the pool). The whole module is skipped when the
+variable is not set, so the suite stays green until such a project exists.
+
+A merge deletes the source branch, so every test creates its own branch; production configurations
+created for the conflict scenario are deleted afterwards.
+"""
+
+import json
+import logging
+import os
+import time
+import uuid
+from collections.abc import Generator
+from dataclasses import dataclass
+from typing import Any
+
+import httpx
+import pytest
+import pytest_asyncio
+from fastmcp import Context
+from mcp.server.session import ServerSession
+from mcp.types import ClientCapabilities, InitializeRequestParams
+
+from keboola_mcp_server.clients.client import KeboolaClient
+from keboola_mcp_server.config import Config
+from keboola_mcp_server.mcp import ServerRuntimeInfo, ServerState
+from keboola_mcp_server.tools.merge_requests.models import MergeRequestsDetailOutput
+from keboola_mcp_server.tools.merge_requests.tools import (
+    create_merge_request,
+    get_merge_request_conflicts,
+    get_merge_requests,
+    merge_merge_request,
+    resolve_merge_request_conflict,
+)
+from keboola_mcp_server.workspace import WorkspaceManager
+
+LOG = logging.getLogger(__name__)
+
+MERGE_REQUESTS_TOKEN_ENV_VAR = 'INTEGTEST_STORAGE_TOKEN_MERGE_REQUESTS'
+COMPONENT_ID = 'keboola.python-transformation-v2'
+BRANCH_GONE_TIMEOUT_SEC = 120
+
+
+def _api_request(method: str, url: str, token: str, **kwargs: Any) -> Any:
+    headers = {'X-StorageApi-Token': token, 'Content-Type': 'application/json'}
+    resp = httpx.request(method, url, headers=headers, timeout=60, **kwargs)
+    resp.raise_for_status()
+    return resp.json() if resp.content else {}
+
+
+def _wait_for_storage_job(base_url: str, token: str, job_id: str, timeout: int = 120) -> dict:
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        job = _api_request('GET', f'{base_url}/v2/storage/jobs/{job_id}', token)
+        if job.get('status') == 'success':
+            return job
+        if job.get('status') in ('error', 'cancelled'):
+            raise RuntimeError(f'Storage job {job_id} failed: {job}')
+        time.sleep(2)
+    raise TimeoutError(f'Storage job {job_id} did not complete within {timeout}s')
+
+
+def _create_branch(base_url: str, token: str, name: str) -> str:
+    job = _api_request('POST', f'{base_url}/v2/storage/dev-branches', token, json={'name': name})
+    job = _wait_for_storage_job(base_url, token, str(job['id']))
+    return str(job['results']['id'])
+
+
+def _branch_exists(base_url: str, token: str, branch_id: str) -> bool:
+    try:
+        _api_request('GET', f'{base_url}/v2/storage/dev-branches/{branch_id}', token)
+        return True
+    except httpx.HTTPStatusError as e:
+        if e.response.status_code == 404:
+            return False
+        raise
+
+
+def _delete_branch(base_url: str, token: str, branch_id: str) -> None:
+    try:
+        if _branch_exists(base_url, token, branch_id):
+            job = _api_request('DELETE', f'{base_url}/v2/storage/dev-branches/{branch_id}', token)
+            _wait_for_storage_job(base_url, token, str(job['id']))
+    except Exception:
+        LOG.exception(f'Failed to delete branch {branch_id}')
+
+
+def _config_path(base_url: str, branch_id: str | None, config_id: str | None = None) -> str:
+    prefix = f'{base_url}/v2/storage' + (f'/branch/{branch_id}' if branch_id else '')
+    return f'{prefix}/components/{COMPONENT_ID}/configs' + (f'/{config_id}' if config_id else '')
+
+
+def _create_config(base_url: str, token: str, branch_id: str | None, name: str, parameters: dict) -> str:
+    result = _api_request(
+        'POST',
+        _config_path(base_url, branch_id),
+        token,
+        json={
+            'name': name,
+            'description': 'Integration test config',
+            'configuration': json.dumps({'parameters': parameters}),
+        },
+    )
+    return str(result['id'])
+
+
+def _update_config(
+    base_url: str, token: str, branch_id: str | None, config_id: str, parameters: dict, change: str
+) -> None:
+    _api_request(
+        'PUT',
+        _config_path(base_url, branch_id, config_id),
+        token,
+        json={'configuration': json.dumps({'parameters': parameters}), 'changeDescription': change},
+    )
+
+
+def _delete_config(base_url: str, token: str, config_id: str) -> None:
+    try:
+        _api_request('DELETE', _config_path(base_url, None, config_id), token)
+    except Exception:
+        LOG.exception(f'Failed to delete production config {config_id}')
+
+
+@dataclass
+class MergeRequestProject:
+    storage_api_url: str
+    storage_api_token: str
+
+
+@pytest.fixture(scope='session')
+def mr_project(storage_api_url: str, env_file_loaded: bool) -> MergeRequestProject:
+    """A dedicated project with the `branches-merge-requests` feature; skips the module when not configured."""
+    token = os.getenv(MERGE_REQUESTS_TOKEN_ENV_VAR, '').strip()
+    if not token:
+        pytest.skip(f'{MERGE_REQUESTS_TOKEN_ENV_VAR} not set; skipping merge-request integration tests.')
+    token_info = _api_request('GET', f'{storage_api_url}/v2/storage/tokens/verify', token)
+    features = token_info.get('owner', {}).get('features', [])
+    if 'branches-merge-requests' not in features:
+        pytest.fail(f'project {token_info["owner"]["name"]!r} must have the branches-merge-requests feature enabled')
+    return MergeRequestProject(storage_api_url=storage_api_url, storage_api_token=token)
+
+
+@pytest.fixture
+def dev_branch(mr_project: MergeRequestProject) -> Generator[str, Any, None]:
+    """A fresh development branch per test (a merge deletes it)."""
+    branch_id = _create_branch(
+        mr_project.storage_api_url, mr_project.storage_api_token, f'integtest-mr-{uuid.uuid4().hex[:8]}'
+    )
+    try:
+        yield branch_id
+    finally:
+        _delete_branch(mr_project.storage_api_url, mr_project.storage_api_token, branch_id)
+
+
+async def _build_context(mocker, project: MergeRequestProject, *, branch_id: str | None) -> Context:
+    keboola_client = KeboolaClient(
+        storage_api_url=project.storage_api_url,
+        legacy_storage_token=project.storage_api_token,
+        headers={'User-Agent': 'KeboolaMCPServer/integtest'},
+    )
+    if branch_id is not None:
+        keboola_client = await keboola_client.with_branch_id(branch_id)
+    workspace_manager = await WorkspaceManager.create(keboola_client)
+    ctx = mocker.MagicMock(Context)
+    ctx.session = mocker.MagicMock(ServerSession)
+    ctx.session.state = {KeboolaClient.STATE_KEY: keboola_client, WorkspaceManager.STATE_KEY: workspace_manager}
+    ctx.session.client_params = InitializeRequestParams(
+        protocolVersion='1', capabilities=ClientCapabilities(), clientInfo={'name': 'integtest-mr', 'version': '0.0.1'}
+    )
+    ctx.client_id = 'KeboolaMCPServer/integtest'
+    ctx.session_id = None
+    ctx.request_context = mocker.MagicMock()
+    ctx.request_context.lifespan_context = ServerState(
+        Config(storage_api_url=project.storage_api_url, storage_token=project.storage_api_token),
+        ServerRuntimeInfo(transport='stdio'),
+    )
+    return ctx
+
+
+@pytest_asyncio.fixture
+async def branch_context(mocker, mr_project: MergeRequestProject, dev_branch: str) -> Context:
+    return await _build_context(mocker, mr_project, branch_id=dev_branch)
+
+
+@pytest_asyncio.fixture
+async def production_context(mocker, mr_project: MergeRequestProject) -> Context:
+    return await _build_context(mocker, mr_project, branch_id=None)
+
+
+def _wait_until_branch_gone(project: MergeRequestProject, branch_id: str) -> None:
+    deadline = time.time() + BRANCH_GONE_TIMEOUT_SEC
+    while time.time() < deadline:
+        if not _branch_exists(project.storage_api_url, project.storage_api_token, branch_id):
+            return
+        time.sleep(3)
+    raise TimeoutError(f'Source branch {branch_id} still exists {BRANCH_GONE_TIMEOUT_SEC}s after the merge')
+
+
+@pytest.mark.asyncio
+async def test_happy_path_create_then_merge(
+    branch_context: Context, production_context: Context, mr_project: MergeRequestProject, dev_branch: str
+) -> None:
+    """Non-SOX default (0 approvals): create → merge, no review step; the MR is published and the branch is deleted."""
+    config_id = _create_config(
+        mr_project.storage_api_url,
+        mr_project.storage_api_token,
+        dev_branch,
+        f'integtest-new-{uuid.uuid4().hex[:6]}',
+        {'x': 1},
+    )
+    try:
+        created = await create_merge_request(branch_context, title='integtest happy path')
+        assert created.state == 'development'
+        assert created.status.mergeable is True
+        assert 'merge_merge_request' in created.status.next_step
+
+        merged = await merge_merge_request(branch_context)
+        assert merged.merged is True, merged
+        assert merged.state == 'published'
+        assert merged.source_branch_deleting is True
+        assert 'production' in merged.next_step
+
+        detail = await get_merge_requests(production_context, merge_request_ids=[created.id])
+        assert isinstance(detail, MergeRequestsDetailOutput)
+        published = detail.merge_requests[0]
+        assert published.derived_state == 'merged'
+        assert {c.configuration_id for c in published.changed_configurations} == {config_id}
+
+        _wait_until_branch_gone(mr_project, dev_branch)
+    finally:
+        _delete_config(mr_project.storage_api_url, mr_project.storage_api_token, config_id)
+
+
+@pytest.mark.asyncio
+async def test_conflict_is_reported_and_resolved(mocker, mr_project: MergeRequestProject) -> None:
+    """Change the same configuration on the branch and in production → conflict → take ours → merge succeeds."""
+    url, token = mr_project.storage_api_url, mr_project.storage_api_token
+    # A conflict needs the configuration on both sides, so it must exist in production BEFORE the branch is
+    # created (a development branch is a copy of production at creation time).
+    config_id = _create_config(url, token, None, f'integtest-conflict-{uuid.uuid4().hex[:6]}', {'query': 'SELECT 1'})
+    branch_id = _create_branch(url, token, f'integtest-mr-conflict-{uuid.uuid4().hex[:8]}')
+    try:
+        ctx = await _build_context(mocker, mr_project, branch_id=branch_id)
+        _update_config(url, token, branch_id, config_id, {'query': 'SELECT 2'}, 'branch change')
+        _update_config(url, token, None, config_id, {'query': 'SELECT 3'}, 'production change')
+
+        created = await create_merge_request(ctx, title='integtest conflict')
+        assert created.status.merge_blockers == ['conflicts']
+        assert created.status.mergeable is False
+
+        refused = await merge_merge_request(ctx)
+        assert refused.merged is False
+        assert refused.refusal == 'conflicts'
+        assert refused.conflicts is not None and {c.configuration_id for c in refused.conflicts} == {config_id}
+
+        conflicts = await get_merge_request_conflicts(ctx)
+        assert [c.configuration_id for c in conflicts.conflicts] == [config_id]
+        conflict = conflicts.conflicts[0]
+        assert conflict.conflicting_paths == ['/configuration/parameters/query']
+        assert {c.changed_by for c in conflict.changes} == {'both'}
+        assert conflict.suggested_take is None
+
+        resolved = await resolve_merge_request_conflict(
+            ctx, component_id=COMPONENT_ID, configuration_id=config_id, take='ours'
+        )
+        assert resolved.resolved is True and resolved.mode == 'ours'
+        assert resolved.remaining_conflicts == []
+        assert 'merge_merge_request' in resolved.next_step
+
+        merged = await merge_merge_request(ctx)
+        assert merged.merged is True, merged
+        production = _api_request('GET', _config_path(url, None, config_id), token)
+        assert production['configuration'] == {'parameters': {'query': 'SELECT 2'}}
+        _wait_until_branch_gone(mr_project, branch_id)
+    finally:
+        _delete_branch(url, token, branch_id)
+        _delete_config(url, token, config_id)
--- a/integtests/tools/test_merge_requests.py
+++ b/integtests/tools/test_merge_requests.py
@@ -59,7 +59,7 @@ def _wait_for_storage_job(base_url: str, token: str, job_id: str, timeout: int =
         job = _api_request('GET', f'{base_url}/v2/storage/jobs/{job_id}', token)
         if job.get('status') == 'success':
             return job
-        if job.get('status') in ('error', 'cancelled'):
+        if job.get('status') in ('error', 'warning', 'terminated', 'cancelled', 'canceled'):
             raise RuntimeError(f'Storage job {job_id} failed: {job}')
         time.sleep(2)
     raise TimeoutError(f'Storage job {job_id} did not complete within {timeout}s')
--- a/integtests/tools/test_merge_requests.py
+++ b/integtests/tools/test_merge_requests.py
@@ -33,7 +33,9 @@ from keboola_mcp_server.tools.merge_requests.tools import (
     get_merge_request_conflicts,
     get_merge_requests,
     merge_merge_request,
+    request_merge_request_review,
     resolve_merge_request_conflict,
+    update_merge_request,
 )
 from keboola_mcp_server.workspace import WorkspaceManager
 
@@ -218,6 +220,16 @@ async def test_happy_path_create_then_merge(
         assert created.status.mergeable is True
         assert 'merge_merge_request' in created.status.next_step
 
+        updated = await update_merge_request(
+            branch_context, merge_request_id=created.id, description='updated by integtest'
+        )
+        assert (updated.id, updated.description, updated.title) == (created.id, 'updated by integtest', created.title)
+
+        # 0 required approvals: request-review lands straight in `approved` and the changeLog gets populated
+        reviewed = await request_merge_request_review(branch_context)
+        assert reviewed.state == 'approved'
+        assert {c.configuration_id for c in reviewed.changed_configurations} == {config_id}
+
         merged = await merge_merge_request(branch_context)
         assert merged.merged is True, merged
         assert merged.state == 'published'
@@ -242,8 +254,9 @@ async def test_conflict_is_reported_and_resolved(mocker, mr_project: MergeReques
     # A conflict needs the configuration on both sides, so it must exist in production BEFORE the branch is
     # created (a development branch is a copy of production at creation time).
     config_id = _create_config(url, token, None, f'integtest-conflict-{uuid.uuid4().hex[:6]}', {'query': 'SELECT 1'})
-    branch_id = _create_branch(url, token, f'integtest-mr-conflict-{uuid.uuid4().hex[:8]}')
+    branch_id: str | None = None
     try:
+        branch_id = _create_branch(url, token, f'integtest-mr-conflict-{uuid.uuid4().hex[:8]}')
         ctx = await _build_context(mocker, mr_project, branch_id=branch_id)
         _update_config(url, token, branch_id, config_id, {'query': 'SELECT 2'}, 'branch change')
         _update_config(url, token, None, config_id, {'query': 'SELECT 3'}, 'production change')
@@ -277,5 +290,6 @@ async def test_conflict_is_reported_and_resolved(mocker, mr_project: MergeReques
         assert production['configuration'] == {'parameters': {'query': 'SELECT 2'}}
         _wait_until_branch_gone(mr_project, branch_id)
     finally:
-        _delete_branch(url, token, branch_id)
+        if branch_id is not None:
+            _delete_branch(url, token, branch_id)
         _delete_config(url, token, config_id)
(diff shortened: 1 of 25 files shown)


integtests/tools/test_merge_requests.py
```diff
@@ -0,0 +1,295 @@
...
+
+def _wait_for_storage_job(base_url: str, token: str, job_id: str, timeout: int = 120) -> dict:
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        job = _api_request('GET', f'{base_url}/v2/storage/jobs/{job_id}', token)
+        if job.get('status') == 'success':
+            return job
+        if job.get('status') in ('error', 'cancelled'):
+            raise RuntimeError(f'Storage job {job_id} failed: {job}')
+        time.sleep(2)
```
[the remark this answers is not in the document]
martinsifra: Fixed in 9afb5ea4: the poller now treats warning/terminated/canceled as terminal failures too.

src/keboola_mcp_server/links.py
```diff
@@ -99,6 +99,13 @@ def get_links(
         return []
 
     # --- Project ---
+    def get_merge_request_link(self, branch_from_id: str | int, title: str) -> Link:
+        """The Keboola UI page of a merge request: its source branch's development overview."""
+        return Link.detail(
+            title=f'Merge request: {title}' if title else 'Merge request',
+            url=self._url(f'branch/{branch_from_id}/development-overview'),
+        )
```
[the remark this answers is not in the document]
martinsifra: Fixed in 9afb5ea4: the link is built from base URL + project id without the manager's session-branch prefix.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,932 @@
...
+    conflict have the user choose: `take='ours'` / `'theirs'` / `'delete'`, or pass `resolved` with the
+    hand-merged content. Taking one side discards the other side's changes — say so. The configuration must
+    be in the merge request's live conflict set. `remaining_conflicts` tells you whether to continue the loop;
+    approvals survive the resolution. Follow `next_step`.
+    """
+    if (take is None) == (resolved is None):
+        raise ToolError("Pass exactly one of take='ours'|'theirs'|'delete' or a resolved configuration.")
+    if resolved is not None:
+        _validate_resolved(resolved)  # strict, before any network call
+
```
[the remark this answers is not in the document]
martinsifra: Fixed in 9afb5ea4: the body is validated once up front and the resulting model is reused.

src/keboola_mcp_server/tools/merge_requests/conflicts.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def _content(side: Mapping[str, Any] | None) -> dict[str, Any]:
+    envelope = (side or {}).get('diff') or {}
+    return {key: envelope[key] for key in DIFF_CONTENT_KEYS if key in envelope}
+
+
+def _ids(items: list[Any]) -> dict[str, Any] | None:
+    """`{id: item}` when every item is a dict with a unique `id` (configuration rows), else None."""
+    if not items or not all(isinstance(item, dict) and item.get('id') is not None for item in items):
```
renatmagadiev, line 66: **Blocking.** `if not items … return None` makes an empty `rows` list “not id-addressable”, so `_walk` (line 86) falls to the positional loop for that pair only, while the other pair still uses `/rows/<id>`. The two path sets never meet, so `conflicting_paths` (exact/prefix match) misses real conflicts and invents false ones.

Reproduced against this branch:

```
base=[r1{x:1}] ours=[r1{x:2}] theirs=[]
  /rows/0                  theirs
  /rows/r1/configuration/x ours
  conflicting_paths: []        <- modify-vs-delete-all reported as no conflict

base=[] ours=[r1] theirs=[r2]
  /rows/0 both agreed=False
  conflicting_paths: ['/rows/0']  <- two different insertions reported as a conflict
  (with base=[r9] the same insertions yield [] per test_rows_with_ids_align_by_id_not_index)
```

Fix: treat an empty list as a trivially id-keyed map — `if not items: return {}` — and add the two cases above to `test_conflicts.py` (the fixture default `rows: []` is currently only exercised with all three sides empty).

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+        raise ToolError("auto_merge='scheduled' requires auto_merge_at (ISO-8601 date-time).")
+    if require_pairing and at is not None and strategy != 'scheduled':
+        raise ToolError("auto_merge_at is only meaningful with auto_merge='scheduled'.")
+
+
+async def _await_storage_job(client: KeboolaClient, job_id: str) -> JsonDict | None:
+    """Polls a Storage job until `success`/`error`; returns None on timeout (the job keeps running)."""
+    deadline = time.monotonic() + MERGE_JOB_TIMEOUT_SEC
+    while True:
+        job = await client.storage_client.job_detail(job_id)
```
renatmagadiev, line 299: **Blocking.** Once `merge_request_merge` has returned a job the merge is irreversible, but this poll has no `try/except`. A single transient failure inside the up-to-600 s window (5xx, `ReadTimeout` — httpx default read timeout is 60 s, connection reset) propagates out of `merge_merge_request` as a generic error with no `job_id`, no `state='in_merge'` and none of the “do not edit the branch” guidance the timeout branch (line 678) gives. The model then believes the merge failed and may retry it (→ 409 not_ready) or keep editing a branch that is being deleted.

Suggest catching `httpx.HTTPError` around `job_detail` and returning the same `MergeResult` as the timeout branch with a warning like “lost contact with job {job_id}: {exc}; check with get_merge_requests”. (workspace.py's pollers have the same gap, but there a retry is harmless.)

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+)
+
+LOG = logging.getLogger(__name__)
+
+MERGE_CONFLICT_CODE = 'storage.mergeRequests.validation'
+MERGE_NOT_READY_CODE = 'storage.mergeRequests.notReadyToMerge'
+# Storage job statuses after which the job never changes again (same set as workspace.py's poller).
+STORAGE_JOB_TERMINAL_STATUSES = frozenset({'success', 'error', 'warning', 'terminated', 'cancelled', 'canceled'})
+MERGE_JOB_TIMEOUT_SEC = 600.0
+MERGE_JOB_POLL_INTERVAL_SEC = 1.0
```
renatmagadiev, line 74: nit: fixed 1 s polling for up to 600 s is ~600 `GET jobs/{id}`; workspace.py's `execute_query` poller steps 1→2→5→20 s for the same reason. Reusing that schedule (or lifting `_next_poll_interval` into a shared helper) would cut it to a few dozen requests. Also worth a sentence in the docstring that this call can block for up to 10 minutes — most MCP clients time out well before that, in which case the model gets a transport error and never sees the `in_merge` result.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+    for fn, annotations in (
+        (get_merge_requests, read_only),
+        (create_merge_request, destructive),
+        (update_merge_request, destructive),
+        (request_merge_request_review, write),
+        (approve_merge_request, write),
+        (request_merge_request_changes, write),
+        (merge_merge_request, destructive),
+        (get_merge_request_conflicts, read_only),
+        (resolve_merge_request_conflict, write),
```
renatmagadiev, line 100: `destructiveHint` looks swapped against the repo convention (`create_config`/`add_config_row`/`create_flow` → `False`; `update_config`/`update_config_row`/`update_flow` → `True`): `create_merge_request` only POSTs a new MR but is `destructive`, while `resolve_merge_request_conflict` calls `configuration_rebase`, which per its own docstring *replaces* the version and with `take='delete'` (or a deleted side) sends `diff={}` = deletes the configuration — yet it is `write` (`destructiveHint=False`). Clients that gate confirmations on the hint would prompt for the harmless one and let the destructive one through; TOOLS.md renders it the same way. I'd make `resolve_merge_request_conflict` destructive and `create_merge_request` non-destructive.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+        if exc.response.status_code != 409:
+            raise
+        body = _error_body(exc)
+        code = body.get('code')
+        message = str(body.get('error') or body.get('message') or exc)
+        conflicts: list[ConflictRef] | None
+        refusal: LastRefusal
+        if code == MERGE_NOT_READY_CODE:
+            refusal, conflicts = 'not_ready', None
+        elif code == MERGE_CONFLICT_CODE or code is None:
```
renatmagadiev, line 643: Two things here:

1. `code is None` → `refusal='conflicts'` also catches any 409 whose body isn't the Storage JSON envelope (`_error_body` → `{}` for a proxy/HTML body). With an empty live list that yields `refusal='conflicts'`, `conflicts=[]`, `mergeable=False` and a `next_step` pointing at `get_merge_request_conflicts`, which then returns nothing — the agent loops. `test_merge_conflict_409_without_errors_fetches_the_live_list` currently locks this in. When `code is None` **and** the live list is empty I'd rather fall to `not_ready` (or re-raise) so the tool has an “unknown refusal” outcome.
2. `await … merge_request_conflicts(mr_id)` on line 650 runs inside the `except` block; if it raises, that exception is what `tool_errors` shows and the 409 diagnosis survives only as `__context__` (never rendered). Capture the 409 data in the `except`, then do the follow-up fetch after it.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+            title=title,
+            description=description,
+            reviewer_ids=reviewer_ids,
+            auto_merge_strategy=auto_merge,
+            auto_merge_at=auto_merge_at,
+        )
+    except httpx.HTTPStatusError as exc:
+        if mapped := _map_write_error(exc):
+            raise mapped from exc
+        if 400 <= exc.response.status_code < 500:
```
renatmagadiev, line 557: This fires for *any* 400–499 except 403, without looking at the backend's error code. A validation 400 (bad `autoMergeAt`, unknown `reviewerIds`) on a branch that still has a listed MR — e.g. a `canceled` one, whose `branchFromId` is not nulled — is reported as “already has merge request N (state closed); a branch can have only one”, and the real message is lost. Same `await`-inside-`except` issue as in `merge_merge_request`: if `merge_requests_list()` or `_detail_with_conflicts()` fails, that error replaces the original one.

Suggest gating on the backend's duplicate-MR `code`/message (the unit test uses `code: 'x'`, so the real one isn't captured anywhere yet), or at least requiring `existing[0]` to be in an open state, and keeping the original `exc` if the lookup fails.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+        raise ToolError(MERGE_REQUEST_BRANCH_ONLY_MESSAGE)
+    if merge_request_id is None:
+        rows = await c.client.storage_client.merge_requests_list()
+        matches = [r for r in rows if same_id((r.get('branches') or {}).get('branchFromId'), c.client.branch_id)]
+        if not matches:
+            raise ToolError(
+                'The current development branch has no merge request yet. Create one with create_merge_request, '
+                'or pass merge_request_id if you meant another branch (open a session on that branch first).'
+            )
+        mr = matches[0]
```
renatmagadiev, line 272: `matches[0]` has no state filter. `merge_requests_list()` returns every MR of the project and only branch *deletion* nulls `branchFromId`, so a `canceled` MR stays attached to a live branch. If a branch has a canceled MR plus a newer open one, the id-less tools (`merge_merge_request()`, `request_merge_request_review()`, `get_merge_request_conflicts()`) act on the dead one — the backend then 4xx's (not 403, so `_map_write_error` passes it through raw) or the conflicts/status shown belong to the wrong MR. Same for `existing[0]` at line 559.

If the backend really allows at most one MR per branch *ever* this is moot — but nothing in the repo says so (the docstring's “A branch can have only one merge request” is unqualified). Filtering `matches` to non-terminal states (or preferring the newest) is cheap insurance either way.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+) -> JsonDict:
+    """
+    The MR-id convention for branch-only tools: with the id omitted, the session branch's single MR; with the
+    id given, the MR is validated to belong to the session branch — never act on another branch's MR.
+    """
+    if c.client.branch_id is None:
+        raise ToolError(MERGE_REQUEST_BRANCH_ONLY_MESSAGE)
+    if merge_request_id is None:
+        rows = await c.client.storage_client.merge_requests_list()
+        matches = [r for r in rows if same_id((r.get('branches') or {}).get('branchFromId'), c.client.branch_id)]
```
renatmagadiev, line 266: Right after a successful `merge_merge_request` the MR's `branchFromId` is nulled and the branch is being deleted asynchronously. The next id-less branch-only call from the same (now dead) session finds `matches == []` and tells the model “The current development branch has no merge request yet. Create one with create_merge_request” — and `create_merge_request` then fails at line 154 with “Branch … not found” (or, inside the deletion window, actually POSTs an MR on a branch being torn down). `_resolve_branch_mr` never looks at `c.session_branch` although `_load` already fetched the branch list; checking it here (or spotting a published MR whose branch is gone) would let you emit the production-handoff message instead.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+            )
+        mr = matches[0]
+        if include_activity_log:
+            mr = await c.client.storage_client.merge_request_detail(mr['id'], include_activity_log=True)
+        return mr
+    mr = await c.client.storage_client.merge_request_detail(merge_request_id, include_activity_log=include_activity_log)
+    branch_from_id = (mr.get('branches') or {}).get('branchFromId')
+    if not same_id(branch_from_id, c.client.branch_id):
+        name = c.branch_from_name(mr)
+        where = f"branch '{name}' (id {branch_from_id})" if name else f'branch id {branch_from_id}'
```
renatmagadiev, line 280: For an already-`published` MR passed by id, `branch_from_id` is `None`, `branch_from_name()` returns `None`, and this renders: “Merge request 43 belongs to **branch id None**, not to the current session branch. Tell the user to open a session on that branch and ask again there.” Short-circuit on `branch_from_id is None` with “this merge request is already merged/closed (derived state …)” before the mismatch check.

src/keboola_mcp_server/tools/merge_requests/tools.py
```diff
@@ -0,0 +1,930 @@
...
+    branches: list[JsonDict]
+    links: ProjectLinksManager
+
+    @property
+    def admin_id(self) -> Any:
+        admin = self.token_info.get('admin')
+        return admin.get('id') if isinstance(admin, Mapping) else None
+
+    @property
+    def can_write(self) -> bool:
```
renatmagadiev, line 128: `can_write` models the `ToolsFilteringMiddleware` rules but cannot see the header gates in `ToolAuthorizationMiddleware` (`X-Read-Only-Mode`, `X-Allowed-Tools`, `X-Disallowed-Tools`), which run outside it. With `X-Read-Only-Mode: true` and an admin token, `get_merge_requests` on an `approved` MR returns `next_step = "Ready: merge it with merge_merge_request …"` for a tool the server neither lists nor allows. `test_next_step_never_recommends_a_write_the_session_cannot_do` doesn't cover that axis. Either feed the authorization config into `SessionContext.can_write`, or soften the guarantee in the docstring. Related: `client.readonly` is checked here but not by the middleware, so an MR write on a read-only scope dies in `RawKeboolaClient.put` with a raw `RuntimeError` instead of a `ToolError` (pre-existing for all write tools, just noting it since this PR introduces `ROLE_DENIED_MESSAGE`).

src/keboola_mcp_server/tools/merge_requests/status.py
```diff
@@ -0,0 +1,309 @@
...
+
+def derive_merge_blockers(mr: Mapping[str, Any], conflicts: Sequence[Any] | None) -> list[MergeBlocker]:
+    """
+    What currently blocks `merge`, as a list so concurrent blockers do not mask each other. Never `None`.
+
+    - `conflicts`: the live conflicts list is non-empty; `conflicts=None` (not fetched) skips only this check
+    - `approvals`: `state == in_review` (the state collapses the requirement; no count until DMD-1969)
+    - `state`: `in_merge` / `published` / `canceled`
+    """
+    server = mr.get('mergeBlockers')
```
renatmagadiev, line 100: Server-first `mergeBlockers` returns the backend list (filtered to known values) and ignores the `conflicts` argument the caller just fetched from `/conflicts`; unknown vocabulary is filtered to nothing. Latent until DMD-1988 ships the field, but then: `mergeBlockers=[]` or `['some_new_kind']` with a non-empty live list → `blockers=[]`, `mergeable=True`, `next_step="Ready: merge it …"` while `get_merge_request_conflicts` returns N conflicts (reproduced). And with `mergeBlockers=['conflicts']` but `conflicts=None` (the merge 409 `not_ready` branch) row 6 renders “**0 conflicts** block the merge; call get_merge_request_conflicts…”, pre-empting the not-ready text (line 220: `n = … else 0`).

Suggest: always union local evidence (`if conflicts and 'conflicts' not in result: append`), don't silently drop unknown blockers (keep them or at least don't set `mergeable=True`), and word row 6 for the unknown-count case.

src/keboola_mcp_server/tools/merge_requests/models.py
```diff
@@ -0,0 +1,355 @@
...
+        branches = raw.get('branches') or {}
+        branch_from_id = branches.get('branchFromId')
+        branch_into_id = branches.get('branchIntoId')
+        merge = raw.get('merge') or {}
+        creator = raw.get('creator') or {}
+        return cls(
+            id=int(raw['id']),
+            title=str(raw.get('title') or ''),
+            description=_opt_str(raw.get('description')),
+            state=raw.get('state'),
```
renatmagadiev, line 202: `state` (also `Reviewer.status` line 58, `auto_merge` line 211, `DerivedStatus.state`) is a strict `Literal` fed the raw API value. Any value outside the set — or a missing `state` → `None` — raises a `ValidationError`, and in list mode (tools.py:379) that fails the **whole** `get_merge_requests()` for one bad row; the write tools hit it in `_detail_with_conflicts` *after* the PUT/POST already succeeded. Meanwhile `derive_state` / `derive_allowed_actions` in status.py deliberately tolerate unknown states, so that defensiveness is unreachable. Since the docstring calls these “raw” API values, either widen to `str` (as `clients/data_science.py` / `scheduler.py` do) or coerce unknowns before validation. Low probability today, high blast radius when it happens.

src/keboola_mcp_server/mcp.py
```diff
@@ -1112,6 +1143,7 @@ async def on_call_tool(
             return await call_next(context)
 
         token_info = await self.get_token_info(context.fastmcp_context)
+        context.fastmcp_context.session.state[TOKEN_INFO_STATE_KEY] = token_info
```
renatmagadiev, line 1146: This writes `token_info` for **every** non-bootstrap tool call into `session.state`, which persists across requests on stdio / `--no-stateless-http` (`config.py` `session_state_persists`), and `_load` accepts whatever Mapping it finds there; MultiProject's `finally` restores the client/workspace keys but not this one. With a multi-project scope and two in-flight calls (A: `get_merge_requests(project_id=X, ids=[…])`, B: anything with `project_id=Y`), B's write can land between A's write here and A's read in `_load` (there are awaits in between), so A builds links with `project_id=Y` and `viewer`/`can_write` from Y's admin. It widens the client-swap race already noted in `tests/test_multiproject.py:344` to every tool pair. Cheap fix: store it on the `KeboolaClient` instance (or a contextvar) instead of session state, or `pop` it in a `finally`.

src/keboola_mcp_server/multiproject.py
```diff
@@ -33,7 +33,9 @@
 # project_id argument to say which -- same single-target resolution/swap as a write tool, just
 # without the write semantics. get_project_info resolves through the active project's
 # WorkspaceManager (workspace id / sql dialect), so it can only ever report one project at a time.
-_SINGLE_TARGET_READ_TOOLS = {'get_project_info'}
+# Merge requests are project-specific ids and the conflict tool is bound to the session branch: neither can be
+# fanned out across projects, so both target one project like a write tool does.
+_SINGLE_TARGET_READ_TOOLS = {'get_project_info', 'get_merge_requests', 'get_merge_request_conflicts'}
```
renatmagadiev, line 38: Heads-up on the branch-only tools under multi-project scope: `client_for_project` builds the swapped client with `branch_id=None`, so any of the five `MERGE_REQUEST_BRANCH_ONLY_TOOLS` called with a `project_id` ≠ active project is always denied by `authorize_tool_call` with “This tool runs only from a development-branch session…” — even though the user *is* on a dev branch. The real cause (a cross-project target can never carry a branch) is never stated, yet all five advertise `project_id: ProjectIdArg`. Either exclude them from cross-project targeting with a specific message, or drop `project_id` from them. No test covers a non-active `project_id` on these tools.

renatmagadiev (changes requested):
Thanks, this is a big and carefully structured PR — the status derivation / `next_step` table and the strict `resolved` validation are nice. I pulled the branch, ran the unit tests (569 passed), `ruff` and `check-tools-docs` (both OK).

Requesting changes for two correctness issues that I consider blocking (details inline):

1. **`conflicts.py` `_ids([])` returns `None`** → when exactly one side has `rows == []`, that pair is walked positionally (`/rows/0`) while the other is walked by id (`/rows/<id>`), so a modify-vs-delete row conflict is reported as `conflicting_paths=[]` (and two different insertions on an empty base are reported as a conflict). Reproduced by running `classify_three_way`. One-line fix.
2. **`_await_storage_job` has no error handling** after the irreversible merge job has started — a single transient 5xx/timeout on `job_detail` surfaces as a generic error with no `job_id`/`in_merge`, so the model may retry the merge or keep editing a branch that is being deleted.

Non-blocking but worth fixing in this PR (inline): swapped `destructiveHint` on `create_merge_request`/`resolve_merge_request_conflict`; code-less 409 → `conflicts` + awaits inside `except` blocks that hide the original error; `_resolve_branch_mr` picking `matches[0]` without a state filter and its two confusing error messages (`branch id None`, “create one” right after a merge); server-first `mergeBlockers` overriding a live non-empty conflict list.

Process notes (not blocking on their own, but please align before merge):
- Branch name `ms/dmd-1701-implementation` doesn't follow the `DMD-1701-…` rule in `CLAUDE.md` / `CONTRIBUTING.md`.
- `CONTRIBUTING.md` requires the RFC to be agreed **before** the implementation PR and to live in `feature_spec/<name>/RFC.md`; #612 is still open and the directory is absent at this PR's head, while `mcp.py` / `status.py` comments already point to it. Merging #612 first (or folding the RFC into this PR) would fix that.

Cleanup nits I didn't comment inline: `resolve_branch_pair` is dead in `src/` (only its own test calls it) and builds `_MrContext(links=None)` behind a `type: ignore`; the `try/except httpx.HTTPStatusError → _map_write_error` block is copy-pasted 7×, a small `@asynccontextmanager` would do; `_MrContext.branch_names` is a plain property rebuilt per MR row in the list comprehension (`cached_property`); `MergeRequestConflictsOutput.next_step` / `ResolveConflictResult.next_step` always equal `status.next_step`; `_resolve_branch_mr(include_activity_log=…)` is never passed `True`; the `same_id(session_branch, default_branch)` check in `create_merge_request` is unreachable because `KeboolaClient.with_branch_id` normalizes the default branch to `None`.

Matovidlo:
@martinsifra this needs to be properly tested both from MCP only perspective and from Kai agent, can we release it to testing stack using dev tag?

src/keboola_mcp_server/tools/merge_requests/conflicts.py
martinsifra: Fixed in f9229223: `_ids([])` now returns `{}` so both pairs are aligned by id. Added both repro cases to `test_conflicts.py` (modify-vs-delete-all → conflict; two insertions on an empty base → no conflict).

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Fixed in f9229223: `_await_storage_job` catches `httpx.HTTPError`; a poll failure returns the same `MergeResult` as the timeout branch (`state='in_merge'`, `job_id`, warning "Lost contact with the merge job …; the merge itself was not interrupted", next_step "do not merge again / do not edit the branch"). Test simulates a `ReadTimeout` mid-poll.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Done in 1e22fff4: polling steps 1→2→5→20 s (same schedule as `workspace.py`), and the tool docstring now says the call can block for up to 10 minutes and how to read an `in_merge` result.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Swapped in 1e22fff4: `create_merge_request` → `destructiveHint=False`, `resolve_merge_request_conflict` → `True` (rebase replaces / deletes). `test_server.py` rows updated; TOOLS.md regenerated.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Fixed in 1e22fff4: (1) a code-less 409 consults the live `/conflicts` list; if it is empty the refusal becomes `not_ready` (unknown refusal) instead of an empty conflict list — the old test was replaced by two (`…_without_live_conflicts_is_not_ready`, `…_with_live_conflicts`). (2) The 409 is diagnosed inside `except` and every follow-up request runs after it.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Fixed in 1e22fff4: the fallback keys on the backend's message `There is already a merge request (N) …` and reads the existing id from it (no list scan); any other 4xx passes through with the backend's original error. The existing-MR lookup runs outside the `except` and falls back to the original error if it fails. Tests: `…_names_the_existing_one` (real message) and `…_passes_other_400s_through`.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Fixed in 1e22fff4: `_open_branch_mrs` keeps only non-terminal MRs of the branch, newest first (`published`/`canceled` skipped). Two new `test_mr_id_convention` rows cover a canceled + open pair and a canceled-only branch.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Fixed in 1e22fff4 + 12beb8d5: with no open MR the resolver first checks the session branch — gone from `branches_list` → "…was probably merged and is being deleted; open a session on the production branch"; still listed but with a `published` MR attached (the deletion window) → the same handoff. Only then "no open merge request; create one". Tests for both windows.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Fixed in 1e22fff4: `branchFromId is None` short-circuits to "Merge request N is already merged (state published) and its source branch is gone; there is nothing to do on it" before the mismatch check. Test row `given_published_says_merged_not_branch_none`.

src/keboola_mcp_server/tools/merge_requests/tools.py
martinsifra: Partly modelled, rest documented (1e22fff4): `can_write` now honours `X-Read-Only-Mode` (read via `ToolAuthorizationMiddleware._get_authorization_config()`) and the read-only client; the docstring states that `X-Allowed-Tools` / `X-Disallowed-Tools` are not modelled. Test axis `x_read_only_mode_header` added. The raw `RuntimeError` from a read-only client is pre-existing for all write tools and left out of this PR.

src/keboola_mcp_server/tools/merge_requests/status.py
martinsifra: Fixed in 1e22fff4: server-first `mergeBlockers` is kept verbatim (unknown kinds included, so they never read as mergeable) and unioned with the live conflicts the caller fetched; `MergeBlocker` is typed `str`. Row 6 has an unknown-count wording when the blocker came from the backend and the list was not fetched. Tests: `server_first_unions_live_conflicts`, `server_first_keeps_unknown_kind`, `6c_conflicts_blocker_without_a_fetched_list`.

src/keboola_mcp_server/tools/merge_requests/models.py
martinsifra: Fixed in 1e22fff4: `MergeRequestState`, `Reviewer.status`, `auto_merge` (and therefore `DerivedStatus.state` / `MergeResult.state`) are `str` with the known vocabulary documented; `from_api` coerces via `str(...)`. Input vocabularies (`AutoMergeStrategy` on create/update, `TakeMode`) stay `Literal`.

src/keboola_mcp_server/mcp.py
martinsifra: Fixed in 1e22fff4: `token_info` no longer touches `session.state`. `on_call_tool` sets a `ContextVar` (`TOKEN_INFO_VAR`) around `call_next` and resets it in `finally`; `_load` reads it. Task-scoped, so concurrent multi-project calls cannot see each other's token info. Test asserts the value inside `call_next` and the reset afterwards.

src/keboola_mcp_server/multiproject.py
martinsifra: Fixed in 1e22fff4: the five branch-only tools no longer take `project_id` (a cross-project target can never carry a branch), so `MultiProjectMiddleware` always dispatches them to the active project. `test_server.py` asserts no `project_id` on `MERGE_REQUEST_BRANCH_ONLY_TOOLS`.

martinsifra:
@renatmagadiev thanks for the thorough review. Both blocking findings are fixed (f9229223), the non-blocking ones and the cleanup nits in 1e22fff4 / 12beb8d5 (replies inline). The branch is rebased on `main` (version 1.83.0) and `tox` is green.

On the process notes:
- The branch name stays as is — renaming would mean a new PR.
- The RFC PR #612 is kept separate and is not a prerequisite for this PR.

Not addressed here: the raw `RuntimeError` from a read-only client on write tools (pre-existing for all write tools).

PR #703, opened 2026-09-09
fix(data-apps): note that direct-grant writes fail on BigQuery

cjayyy:
## Description

**Linear**: [AI-3903](https://linear.app/keboola/issue/AI-3903/support-17607-zapisovatelny-workspace-pro-data-app-storage-access-rw) (platform bug: [DMD-1259](https://linear.app/keboola/issue/DMD-1259))

### Change Type

### Summary

The `storage` parameter description on `modify_python_js_data_app` / `create_python_js_data_app` states that declaring `unload_strategy: "direct-grant"` means "the workspace is granted direct SELECT/INSERT/UPDATE/DELETE/TRUNCATE on the destination Storage table".

**On BigQuery the write half of that is not true.** Reads succeed; `INSERT`/`UPDATE`/`DELETE`/`TRUNCATE` fail with `Permission bigquery.tables.updateData denied`. The direct-grant table IAM is granted to the workspace's **own service account**, while Query Service executes as a **separate user it mints via its `/credentials` endpoint**, which receives only project-level read. Snowflake is unaffected. Tracked as DMD-1259, open since 2026-04.

Why this matters here rather than only in the docs: this description is what the agent reads at runtime. Acting on it, Kai builds an app that writes to Storage, hits the denial, and then loops on the remedy documented in `ai-kit` ("add the table with `direct-grant`; redeploy so the ephemeral workspace is re-provisioned") — which does not grant the permission. That is exactly what happened in AI-3903: five redeploy attempts, then a fallback to writing through a transformation's output mapping.

The new wording keeps direct-grant declarations valid (they are correct and harmless on BigQuery), but tells the agent not to build an app that depends on app-side writes there, and to say so to the user instead of discovering it at deploy time.

Companion docs fix in the skill the agent loads alongside this: keboola/ai-kit#109.

## Testing

Not exercised through a client — this changes only a parameter description string, with no code path or schema shape affected. What I did verify:

- `uvx ruff format --check` and `uvx ruff check` on the touched file — clean
- `pytest tests/tools/test_data_apps.py` — **140 passed**
- `python -m keboola_mcp_server.generate_tool_docs` — `TOOLS.md` regenerated, single-line diff matching the new description (`tox -e check-tools-docs` should be green)

The underlying claim was verified against a live BigQuery project (5839, stack `us-east4.gcp`): added a table that had **never** appeared in an earlier config version, redeployed so a genuinely new workspace was provisioned (`14390382` → `14390470`), then from inside the app `SELECT COUNT(*)` returned 9 rows while `DELETE` on the same table from the same workspace was denied. Reproduced on two tables in two different buckets.

### Optional testing

## Checklist

- [x] Self-review completed
- [ ] Unit tests added/updated (if applicable) — n/a, description-only change; no test asserts this string
- [ ] Integration tests added/updated (if applicable) — n/a
- [x] Project version bumped according to the change type (1.80.1 → 1.80.2)
- [x] Documentation updated (if applicable) — `TOOLS.md` regenerated

Draft: the wording is a judgement call about how strongly to steer the agent away from BigQuery write-back apps, and it should stay consistent with ai-kit#109. Happy to soften or sharpen it — and to drop the version bump if you would rather batch it.


MiroCillik:
@cjayyy does this mean, the Data apps on BigQuery can't use direct write? How come, we don't know about this 😱 
I'm gonna try it, coz I think this has worked.

cjayyy:
@MiroCillik you're right to doubt it — I got this wrong, and direct write on BigQuery is not broken. **Project 5839 simply never had the `direct-grant-output-mapping` feature enabled.** Verified just now on `GET /v2/storage`: 13 features, and that one is not among them.

So the `updateData denied` errors were the backend behaving correctly — the app declared `unload_strategy: "direct-grant"` output tables, but with the feature off no table grant is ever issued. Reads kept working through the project-level `dataViewer` binding, which is exactly the read-works/write-denied split I spent the day chasing and misread as a bug.

Where I went wrong: the trace I was investigating (SUPPORT-17607) recorded Kai's earlier attempt as "Storage Access feature enabled", I treated that as established, and never checked the project's feature list myself. I then found [DMD-1259](https://linear.app/keboola/issue/DMD-1259) — a real, still-open issue with the identical error string — and stopped looking, because a matching known bug felt like confirmation. Both reproductions I ran were on this same feature-less project, so re-running the test twice only made me more confident in the wrong answer. `kbc-stacks/features.yaml:118-124` documents it as supported on both backends, which I also mis-cited earlier from a stale Slack thread.

So please don't take my claim as evidence against your recollection — if it worked for you, it worked.

### What that means for these PRs

Both this one and [ai-kit#109](https://github.com/keboola/ai-kit/pull/109) are built on the wrong premise and should not merge as written. I'll rework them today into what the evidence actually supports:

- **The real gap is diagnosability, not capability.** With the feature off you get a raw GCP `bigquery.tables.updateData denied` and nothing pointing at the missing feature — no mention in the error, and the config saves happily. That is what burned a day here and what sent Kai into a redeploy loop.
- The genuinely useful guidance is: on an `updateData denied`, **check `direct-grant-output-mapping` on the project first**, before touching config or redeploying. Not "writes are broken on BigQuery".
- The Storage API write path I documented does work (verified end-to-end), but it belongs as an alternative pattern, not as a workaround for a bug that isn't there.

I'll also correct my comments on DMD-1259, which currently assert this reproduces with the feature enabled — it doesn't, and that claim shouldn't sit there misleading whoever picks it up.

Worth asking separately, since you own this: should enabling Storage Access in the app UI surface a warning when the project feature is off? The UI let me add writable tables and save with no indication anything was missing.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

cjayyy:
@MiroCillik Checked every precondition on project 5839 (BigQuery, `us-east4.gcp`):

- `data-apps-storage-workspace` ✅ (on from the start)
- `direct-grant-output-mapping` ✅ (Vašek enabled it today)
- 8 output tables with `unload_strategy: "direct-grant"` ✅
- fresh workspace `14392046`, provisioned **after** both features ✅

Query Service `DELETE` on the target table → `Permission bigquery.tables.updateData denied`. `SELECT` on the same table, same workspace, same token: 164 rows, fine. A **Storage API** import to that table works in ~10s. Enabling `direct-grant-output-mapping` changed nothing.

That points back at [DMD-1259](https://linear.app/keboola/issue/DMD-1259) (Erik, April): the table IAM goes to the workspace SA, but Query Service runs as the user it mints via `/credentials`, which gets read only.

**Your recollection is the lead I'd rather chase than argue with** — two questions:

1. **Snowflake or BigQuery?** Snowflake uses a different grant path and is expected to work; if that was your case, both stories hold.
2. If BigQuery — which stack, and was the write from app code via Query Service, or from a transformation / workspace / SQL editor? Those use the workspace SA directly and do work; it's specifically the app→Query Service hop that fails here.

(I posted and deleted an earlier reply claiming the feature was missing — that was me misreading `/v2/storage` component features for project ones. Apologies if you saw it.)

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

MiroCillik:
Guys from DMD are on it too: https://keboolaglobal.slack.com/archives/C0553EM2RGX/p1789029545494349

cjayyy:
Thanks — that thread closes it. Zajca confirmed the mechanism in code on DMD-1259: a BigQuery DGOM workspace has **two** service accounts, and table IAM is granted only to the primary one (`CreateWorkspaceHandler.php:130-200`), while `/credentials` mints a second (`<stack>-ws-<id>-qs-<rand>`) that never receives it. Snowflake escapes it because `GRANT ROLE <ws_role> TO USER` lets the credentials user inherit the grants; BigQuery has no such indirection. He also flagged that DGIM has the same cause and fails harder, plus a second gap where mapping changes would leave the credentials SA stale.

So the platform bug is real and now owned, and Vojta's "spíš zamýšlené chování" is fair — nobody designed `/credentials` with DGOM in mind.

I'll rework this PR's wording once their fix lands or its shape is settled, so the description matches reality rather than the current failure. The Storage API path stays worth documenting as an alternative either way.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

MiroCillik:
Yes, let's wait for the DMD-1259 resolution.

cjayyy:
Btw. IF DGOM is required for this to work, that's another issue, because DGOM is not enabled by default.

cjayyy:
Closing — obsolete. The underlying platform bug [DMD-1259](https://linear.app/keboola/issue/DMD-1259) is fixed: [connection#8444](https://github.com/keboola/connection/pull/8444) ("BigQuery workspace credentials inherit direct grants") merged 2026-09-16 and is rolled out.

Retested on a BigQuery project today (2026-09-17) through the Query Service, against the `-qs-` credentials service account that used to be denied: `INSERT`, `UPDATE` and `DELETE` all complete with `rowsAffected: 1`. The `storage` parameter description this PR adds a caveat to would now be misleading — `unload_strategy: "direct-grant"` works on BigQuery as documented, so no change is needed.

PR #711, opened 2026-09-17
AI-3515: bump fastmcp to 4.0.0 and mcp to 2.2.0

Matovidlo:
## Description

**Linear**: [AI-3515](https://linear.app/keboola/issue/AI-3515/review-mcp-spec-release-candidate-implications)

### Change Type

- [x] Major (breaking changes, significant new features)

### Summary

**CI is fully green** (all Python versions, ruff, check-tools-docs, and Integration Tests against
the live backend). One item remains a recommended follow-up rather than a blocker — see "Still
open" below.

AI-3515 asks us to review the MCP spec release candidate (2026-07-28) and its implications for
this server. The Python side of that is: `mcp` (the official SDK) shipped a major v2 line built
against that spec, and `fastmcp` only gained v2 support at `fastmcp==4.0.0`. The two are coupled —
`fastmcp` 3.x hard-caps `mcp<2.0`, `fastmcp>=4.0.0` requires `mcp>=2.0,<3.0` — so it's one atomic
bump, not two independent ones.

This PR bumps `fastmcp` 3.4.7 → `4.0.0` and `mcp` 1.28.1 → `2.2.0` and fixes every break the bump
caused, including two design-level changes (`Tool.serializer` removal, and a real, previously
latent bug the new client behavior exposed).

**Mechanical fixes:**
- `fastmcp.tools.tool.ToolResult` moved to `fastmcp.tools.ToolResult`.
- `mcp.shared.context.RequestContext` was removed; test fixtures now use
  `fastmcp.server.context.FastMCPRequestContext` as their mock spec.
- The `mcp.server.fastmcp` compat shim (embedded FastMCP 1.0) is gone in `mcp` 2.x; fixed 9 test
  files that still imported `Context` from there.
- `mcp.types.ToolAnnotations` fields are `camelCase` in v1, `snake_case` in v2. Construction still
  works via a pydantic alias; only the places that **read** the attribute back needed updating
  (`mcp.py::is_read_only_tool`, `generate_tool_docs.py`, and matching test fixtures/assertions).
- `ServerNotification` is a plain union type in mcp v2, not constructible; `send_notification()`
  now takes the raw notification directly (`tools/sql.py`).
- `InitializeRequestParams.clientInfo` → `client_info` (same rename pattern), in `errors.py`'s
  user-agent construction.
- `FastMCPRequestContext.meta` is the raw wire `_meta` mapping now, not an object with a
  `.progressToken` attribute — `getattr()` on it silently returned `None`, which would have
  silently disabled out-of-band query-cancellation notifications in production
  (`tools/sql.py::_client_progress_token`). Unit tests alone could not have caught this since they
  mock `Context` wholesale.
- `ctx.session.state` (the per-connection dict used ~80 places for `KeboolaClient`/
  `WorkspaceManager`/OAuth session/scope) is confirmed **unchanged** — no redesign needed.

**Design-level fix — `Tool.serializer` removal:**
`Tool`/`FunctionTool` dropped the `serializer` field entirely in fastmcp 4. This server used it for
a blanket default (`KeboolaMcpServer.add_tool()`) plus ~19 explicit
`serializer=toon_serializer_compact` / `toon_serializer` call sites, for TOON-format compact tool
output. fastmcp's guidance: *"Return `ToolResult` from your tools instead."* `Tool.convert_result()`
is the actual extension point, and `FunctionTool.from_function()` respects `cls`, so this PR adds
three `FunctionTool` subclasses (`PlainFunctionTool`, `ToonFunctionTool`, `ToonCompactFunctionTool`)
that delegate to `super().convert_result()` (so `structured_content` is still correctly derived
from each tool's `output_schema`) and only swap in the custom-encoded text `content`. Verified with
a real `fastmcp.Client` round-trip that both `structured_content` and the compact/exclude-none text
content come out correctly.

**Design-level fix — a real, previously latent bug in `flow/model.py`, exposed by mcp v2's client:**
mcp v2's client now reconstructs a tool's result by dumping the return value with `by_alias=False`
(Python field names) and **re-validating that dump against the same model**
(`mcp/client/session.py::validate_tool_result`). `BaseExtraModel` (base for every conditional-flow
node: phases, tasks, transitions) declares fields like `JobTaskConfiguration.component_id` with a
bare `alias='componentId'` and no `populate_by_name=True` — a shape that has never survived a
dump-then-revalidate round trip. That validation failure happened inside
`Flow.configuration`'s undiscriminated union (`FlowConfiguration | ConditionalFlowConfiguration`),
so pydantic silently fell back to the legacy `FlowConfiguration` branch (built from lenient,
`extra='ignore'` models with no `next` field) instead of raising — silently dropping real phase
transitions from `get_flows`'s response instead of erroring. This bug predates this PR entirely;
it just had nothing to trigger it until mcp v2's client added the round-trip. Found via a full
local repro (`create_conditional_flow` → `modify_flow` → `get_flows` through a real
`fastmcp.Client`, against a fake in-memory Storage backend to rule out live-backend flakiness) and
fixed with one line: `BaseExtraModel` now sets `populate_by_name=True`, matching the pattern
`FlowPhase`/`FlowTask` already use via `AliasChoices`.

**Test evidence:**
- `tox` (pytest + ruff + check-tools-docs): all green, 2186 passed / 24 skipped / 0 failed.
- CI: green on Python 3.10/3.11/3.12, Analyze, CodeQL, **and Integration Tests (13m11s) against the
  live backend**.

### Still open (recommended follow-up, not a blocker)

- `src/keboola_mcp_server/oauth.py` — `SimpleOAuthProvider` subclasses
  `fastmcp.server.auth.auth.OAuthProvider` and uses `mcp.server.auth.provider` /
  `mcp.server.auth.settings` / `mcp.shared.auth` types directly. FastMCP 4's release notes describe
  a full OAuth rewrite (hardened consent flows, issuer validation, JWT verification), and mcp v2's
  migration guide lists several OAuth-shape changes. Unit tests for it pass, but given the
  `BaseExtraModel` bug above turned out to be exactly the kind of thing unit tests can't catch on
  their own (needs a real dump → revalidate round trip to surface), I'd want a dedicated pass on
  the OAuth surface — ideally exercised against a real OAuth flow — before fully trusting it. This
  is the server's auth boundary; not something to wave through on "unit tests still pass."
- A repo-wide grep found ~17 other files with the same bare-`alias=`-without-`populate_by_name`
  shape that caused the `flow/model.py` bug. Whether each is actually reachable through a tool's
  `structured_content` path needs individual verification — worth a follow-up audit, not bundled
  here since it's outside this PR's actual trigger.

## Testing

- [ ] Tested with Cursor AI desktop (`Streamable-HTTP` transports) — not done with a live desktop
      client in this session; verified instead via `tox`, CI's full build matrix, and CI's
      Integration Tests suite against the real backend (which exercises the exact same
      streamable-HTTP tool-call path).

### Optional testing

## Checklist

- [x] Self-review completed
- [x] Unit tests added/updated (if applicable) — updated existing tests to match the mcp v2 API
      shape; no new test scenarios beyond what the migration and the `flow/model.py` fix required
- [x] Integration tests added/updated (if applicable) — no new integtests added; the existing
      `test_update_flow` now passes against the live backend, confirming the fix
- [x] Project version bumped according to the change type — `1.82.1` (patch); given the
      `flow/model.py` fix corrects real (if latent) production behavior, worth reconsidering
      whether this should be a patch or something else before merge
- [ ] Documentation updated (if applicable) — n/a, no tool-facing behavior changed; `TOOLS.md`
      regenerated with no diff

---


Matovidlo:
Tested sample flow of logging in, exeucting couple of tools it seems to be working

cjayyy (approved):

